PR #7, opened 2015-09-16
fix fboss agent and dependency build error

superchild:
fix build error for fboss agent and its dependency library.


jaymzh:
This can be closed - this made it into this repo a long time ago via some other merge...

PR #8, opened 2015-09-28
README.md for Python Tools

capveg:
@simpkins @sonoble

FYI, we were able to resolve the python tool dependencies using the steps in the attached README.  Please take a look and (1) comment on what we did wrong and (2) consider if this might be of use for others.


simpkins:
Imported as 6f3382ea5d690f510a6abb3ddd266ee14ca825de

PR #12, opened 2015-10-30
Fixed build script to include wangle

ziozzang:
Fixed script building facebook/thrift with facebook/wangle.
I tested to build fboss, but It failed and show error "need wangle to build thrift"
so, I fixed build script to include wangle to build.

but, there's still error exist on build thrift itself. (maybe linking?)
anyway, I fixed thrift with wangle.

this is pull-request, but also requesting of fixing broken environment. :)

I search google this issue, but It's generic FB's environment issue..
but many of issue say as solution "sudo make install".
but It's not good solution as you know..

Thanks.

### DIFF
--- a/getdeps.sh
+++ b/getdeps.sh
@@ -22,9 +22,13 @@ function build() {
     (
         echo "building $1..."
         cd $1
-        if [ ! -x ./configure ]; then
-            autoreconf --install
-            ./configure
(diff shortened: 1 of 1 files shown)


jaymzh:
This can be closed - this was fixed in a different way.

PR #13, opened 2015-11-11
Fixing Build on Ubuntu 14.04

neuhausler:


oridb:
@facebook-github-bot import

PR #19, opened 2016-01-09
Add NDP support to cli

jaymzh:


jaymzh:
This was pushed via an internal->external sync.

PR #35, opened 2016-01-22
Fix build issue resulting from an old/incompatible folly or fbthrift previously installed on host

rizard:
Do not look in /usr/local/lib for fbthrift, folly, wangle, or any other 'external' package. The host might have a different version installed that is incompatible w/fboss...as I did. Cmake will prioritize /usr/local/lib and other typical install locations over the user-defined paths in the find_library() function. Specifying NO_DEFAULT_PATH will tell cmake to only use the PATH(s) explicitly specified in find_library().


oridb:
Hm, the bot doesn't seem to have posted with the request yet -- Do you think you could sign the CLA? 

https://code.facebook.com/cla

rizard:
@oridb Just signed the CLA

oridb:
Just as a heads up, this will probably take a little bit of time to land, since there's some cleanup I need to do before it can be imported to our internal repository, but I'm working on it.

capveg:
Hi @rizard !  I'm cleaning this up now and I don't think this patch is needed any more.  Can you confirm?

rizard:
@capveg this looks like my netlink listener code. Maybe I pushed to the wrong branch at some point. The title of this commit seems to indicate it was related to the build

capveg:
This is no longer needed - thanks for the contribution!

PR #36, opened 2016-02-01
Fix to QSFP port mapping

neuhausler:
and fix to readme.

### DIFF
--- a/fboss/agent/tools/fboss_route.py
+++ b/fboss/agent/tools/fboss_route.py
@@ -124,14 +124,14 @@ def enable_port(args):
     port = args.en_port
     with get_client(args) as client:
         portnum = int(port)
-        client.setPortState(5, True)
+        client.setPortState(portnum, True)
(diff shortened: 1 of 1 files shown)


neuhausler:
Thanks @simpkins  for the explanation, and yes I see and now understand the problem you are in!

Just some background info from my side. The "issue" got raised when I dived in to the log-output of FBOSS. I used one QSDF/SFP splitter cable and plugged it in to different ports, and the logs I got mentioned for example links 5,6,7,8 where up and QSFP port 0 .. or link 1,2,3,4 and port ! .. which obviously added additional confusion while debugging your real issue.

Btw, the wedge spec on OCP side mentions that the first QSFP port, port 0 is the first/lower plug. So that will change?

capveg:
Hi Marcel,

I don't think this is an issue anymore.  Do you mind if I close this PR?

See the new qsfp_server code if you're interested in how this got re-written.

PR #40, opened 2016-03-29
Netlink manager

rizard:
Add tap interfaces to switch host OS. There is a 1:1 mapping between a tap interface and an FBOSS Interface. Use netlink to monitor these tap interfaces for state changes, such as add/del/mod IPv4, IPv6, MAC addresses, MTU, ARP and NDP entries, and routes. Upon an observed change, notify FBOSS of the event. For example, if an IPv4 address is added to a tap interface, say tap01, FBOSS's corresponding Interface will be updated with the IP address. This and other updates received by the netlink manager are applied as a state change through the SwSwitch.

Notes:
-- The netlink manger and the existing TunManager are mutually exclusive
-- Limitation of one tap interface per VLAN at present
-- Work in progress (95% complete as of 03/29/16)

Changes to FBOSS:
-- Addition of libnl3 libraries
-- New GFLAGS for toggling b/t netlink and FBOSS managed modes
-- Addition of new NetlinkListener (soon to be renamed NetlinkManager)
-- Incorporation of netlink code into SwSwitch
-- Addition of L2 packet handling/forwarding to/from tap interfaces in SwSwitch
-- Minor changes to IPv4Handler, IPv6Handler, and ArpHandler for punting packets to CPU if in netlink managed mode
-- Modification of switch state types (e.g. state/Interface) for ease of performing state updates
-- Modification of BCM OpenNSL code to allow for interface updates outside of a warm boot scenario

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,7 @@ include_directories(${CMAKE_BINARY_DIR}/gen)
 include_directories(${CMAKE_BINARY_DIR}/gen/fboss/agent/if/gen-cpp2)
 include_directories(${CMAKE_BINARY_DIR}/gen/common/fb303/if/gen-cpp2)
 include_directories(${CMAKE_BINARY_DIR}/gen/common/network/if/gen-cpp2)
+include_directories(/usr/include/libnl3)
 
 # System libraries
 find_library(GLOG glog)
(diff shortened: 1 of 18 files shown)


capveg:
I'm closing this out as all of this logic has been ported into the netlink_manager code as a external daemon.  Thanks for the contribution!  I definition accelerated us in our development!

PR #42, opened 2016-05-19
Fix fboss build on ubuntu 14.04

neuhausler:


oridb:
Heh, I had just sent out a diff to do this internally; waiting on review, but it should be out in a bit.

PR #47, opened 2016-11-17
Add missed-out commit directly to github

allwync:
Summary:
Push to github has been failing since Oct 3, 2016.  On investigation
it appears that a commit from January 2016 was not pushed out, which
is resulting in merge issues because of a modification to one of
the files that the Oct commit is changing.  We will resolve this issue
by doing this commit from github.

Reviewed By: has


oridb:
Note that you can do this directly from shipit, which will preserve the commit message and internal correlations correctly.

saifhhasan:
Oh. Will look into `shitip` next time. Thanks for the pointer @oridb

PR #53, opened 2017-09-07
cmake test with system libraries

linhbui:
2nd attempt, with system libraries

### DIFF
--- a/getdeps.sh
+++ b/getdeps.sh
@@ -62,6 +62,25 @@ function build_cmake() {
     )
 }
 
+function build_gtest_and_gmock() {
+    (
+        mkdir build.test && cd build.test
+        for LIB in gmock gtest
+        do
+          mkdir $LIB && cd $LIB
+          echo "building $LIB..."
+          test -d "/usr/src/$LIB" || die "Missing dir /usr/src/$LIB"
+          cmfile=/usr/src/$LIB/CMakeLists.txt
+          test -e $cmfile || die "Missing cmake file for $LIB"
+          echo ${PWD}
+          echo cmake $LIB $CMAKEFLAGS
+          cmake $cmfile  $CMAKEFLAGS -B./ || die "Cmake failed for $LIB"
+          make -j "$NPROC" || die "Failed to build $LIB"
+          cd ..
+        done
+    )
+}
+
 function get_packages() {
     echo "installing packages"
     sudo apt-get install -yq autoconf automake libdouble-conversion-dev \
@@ -69,7 +88,7 @@ function get_packages() {
         libevent-dev flex bison libgoogle-glog-dev scons libkrb5-dev \
         libsnappy-dev libsasl2-dev libnuma-dev libi2c-dev libcurl4-nss-dev \
         libusb-1.0-0-dev libpcap-dev libdb5.3-dev cmake libnl-3-dev \
-        libnl-route-3-dev iptables-dev
+        libnl-route-3-dev iptables-dev libgtest-dev google-mock
 
    sudo apt-get install -yq shtool pkg-config
 }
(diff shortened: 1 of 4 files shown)


getdeps.sh
```diff
@@ -62,14 +62,31 @@ function build_cmake() {
...
+function build_gtest_and_gmock() {
+    (
+        for LIB in gmock gtest
+        do
+          echo "building $LIB..."
+          cd "/usr/src/$LIB" || die "Missing dir /usr/src/$LIB"
+          if [ -e ./CMakeLists.txt ]; then
+              echo cmake .. $CMAKEFLAGS
+              sudo cmake CMakeLists.txt $CMAKEFLAGS || die "Cmake failed for $LIB"
+              sudo make -j "$NPROC" || die "Failed to build $LIB"
```
capveg: This looks like you're creating the libraries and .o files in /usr/src, which is probably not the right thing.  What about something like this instead:

```
mkdir build.test && cd build.test
for LIB in gmock gtest ; do
          mkdir $LIB && cd $LIB
           echo "building $LIB..."
           test -d /usr/src/$LIB || die "Missing dir /usr/src/$LIB"
           cmfile=/usr/src/$LIB/CMakeLists.txt
           test -e $cmfile || dir "Missing $cmfile"
           cmake $cmfile $CMAKEFLAGS || die "Cmake failed for $LIB"
           make -j "$NPROC" || die "Failed to build $LIB"
           cd ..
       end
```

getdeps.sh
```diff
@@ -62,14 +62,31 @@ function build_cmake() {
...
+
 function get_packages() {
     echo "installing packages"
     sudo apt-get install -yq autoconf automake libdouble-conversion-dev \
         libssl-dev make zip git autoconf libtool g++ libboost-all-dev \
         libevent-dev flex bison libgoogle-glog-dev scons libkrb5-dev \
         libsnappy-dev libsasl2-dev libnuma-dev libi2c-dev libcurl4-nss-dev \
         libusb-1.0-0-dev libpcap-dev libdb5.3-dev cmake libnl-3-dev \
-        libnl-route-3-dev iptables-dev
+        libnl-route-3-dev iptables-dev libgtest-dev googlemock
```
capveg: On my system this is 'google-mock' (with a hyphen) - is this a typo?

linhbui: fixed

getdeps.sh
```diff
@@ -62,14 +62,33 @@ function build_cmake() {
...
+        for LIB in gmock gtest
+        do
+          mkdir $LIB && cd $LIB
+          echo "building $LIB..."
+          test -d "/usr/src/$LIB" || die "Missing dir /usr/src/$LIB"
+          cmfile=/usr/src/$LIB/CMakeLists.txt
+          test -e $cmfile || die "Missing cmake file for $LIB"
+          echo ${PWD}
+          echo cmake $LIB $CMAKEFLAGS
+          sudo cmake $cmfile  $CMAKEFLAGS -B./ || die "Cmake failed for $LIB"
```
linhbui: @capveg I have to add sudo & -B./ here because it wouldn't work otherwise on my vm.

capveg: In general, files in /usr/... are not meant for people to write to unless their' updating the system.  The way that I've (tried to) describe involves building everything in a new directory as part of the fboss code checkout, e.g., in "build.test".  If you do it that way, you won't need 'sudo'.  Does that make sense?

linhbui: I see. I should have checked to see if I can remove the sudo after I figured out the -B./ part of it. Thanks! I will update the pr now.

linhbui:
I don't want to deal with merge conflict so created a new pr. This one became obsolete so closing.

PR #63, opened 2018-02-14
Propose fix two typos

jeis2497052:
are these updates OK?


vaibhav4516 (commented):
I think, changes are ok.

PR #64, opened 2018-03-05
Added hack to enable Open/R for OSS FBOSS

capveg:
Open/R uses IPv6 multicast for route discovery.  Unfortunately,
the old version of OpenNSL that we're compiling against
does not have ACL support (though newer versions do) so apply
this band-aid to make Open/R work until we can update OpenNSL
to a more modern version.

The specifics are we add a static L2 entry for mac 33:33:00:00:01
for each VLAN to send to CPU.  This is the well-known MAC address
that ff00::1 (the route discover address) translates to.  We do
this because neither ACL tables nor L2 Cache are exposed in
this version of OpenNSL.


capveg:
NOTE: this code causes a packet loop as the outgoing packets with dst addr=33:33:00:00:01 are punted *back* to the source switch's CPU.  Closing and trying a different tack.

PR #69, opened 2018-06-17
Allow zero interface multiple vlan configuration

ventris:
Summary:
Currently when applying a configuration with multiple vlans and no L3 interfaces the message "Vlan 552 refers to 0 interfaces" is displayed and configuration is rejected.
Most likely this was an accident in the original code as having no L3 interfaces is quite rare.


capveg:
Hi @bluecmd - finally digging into this.

We actually have an internal operational requirement that there's a 1:1 relationship between vlan and interfaces (and it's a config mistake if it's not) but I admit this is a totally reason thing in general.  

What do you think about wrapping your change in an "#ifdef IS_OSS" to make this change for our open source FBOSS only?

PR #80, opened 2019-01-15
[fboss] refactor python thrift client code

shrikhare:
We have quite a few python utilities which create FBOSS thrift clients.  This
adds a new fboss.thrift_clients module, to consolidate the logic for creating
python thrift clients.

Differential Revision: D1979276

fbshipit-source-id: a43d9ac4b79bdd62288e27b5eaece0b6121a3b44


shrikhare:
The issue is fixed in github already

PR #156, opened 2023-09-21
weutil : Changes to support Meru800bia

arajeev-ARISTA:
### Summary
**Use FIELD_UINT instead of FIELD_HEX to parse CRC16 to be consistent with V4 spec:**
`parseHex()` function is currently used to parse FIELD_HEX values. `parseHex()` function internally uses `std::stoi()` to convert the bytes to a hex value. Since the CRC16 bytes are uint values (and not their ASCII equivalent) `parseHex()` function won’t work. Changing to FIELD_UINT would make sure the right parsing function `parseUint()` is used.

**Introduce a new local variable `lastByte` in `loadEeprom()` function to handle large offset values:**
This helps to avoid limiting `fileSize` to a maximum value equal to the buffer size in the line `fileSize = fileSize > max ? max : fileSize;`. As a result, the offset value can be larger than the buffer size. Previously, the following check caused exceptions whenever the offset exceeded the buffer size
```
if (offset > fileSize) {
    throw std::runtime_error("Offset exceeds EEPROM size.");
 }
```
Additionally, `lastByte` also helps to track the number of bytes to be read when `offset` value is larger than the buffer size

**Store lowercase string in translatedFrom:**
In the `getEepromPath()` function, set `translatedFrom` to the lowercase string `lEeprom`. `translatedFrom` is later used to look up `config_.configEntry` which stores keys in lowercase.

### Testing
Read EEPROM with the command `weutil --eeprom SCM --config_file weutil.json`

### DIFF
--- a/fboss/platform/weutil/WeutilImpl.cpp
+++ b/fboss/platform/weutil/WeutilImpl.cpp
@@ -82,7 +82,7 @@ void WeutilImpl::initializeFieldDictionaryV4() {
       {18, "Extended MAC Address Size", FIELD_UINT, 2, VARIABLE});
   fieldDictionaryV4_.push_back(
       {19, "EEPROM location on Fabric", FIELD_STRING, VARIABLE, VARIABLE});
-  fieldDictionaryV4_.push_back({250, "CRC16", FIELD_HEX, 2, VARIABLE});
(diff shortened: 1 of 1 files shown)


arajeev-ARISTA:
Closing as the issues are resolved by other changes.

PR #158, opened 2023-09-27
Meru800bia: update BIOS commands in fw_util config file

arajeev-ARISTA:
### Summary
`-i aboot_conf` option in the BIOS commands would result in updating the configuration section in the BIOS layout. However, the configurations are programmed by BMC and they should remain persistent during BIOS updates. Removed the `-i aboot_conf` option from the commands to fix this issue.

### Testing
Updated BIOS using `fw_util` and made sure the configurations remain unchanged
Read the BIOS using `fw_util` with two different sets of BIOS configurations and made sure the output of read operations are the same

### DIFF
--- a/fboss/platform/configs/meru800bia/fw_util.json
+++ b/fboss/platform/configs/meru800bia/fw_util.json
@@ -1,12 +1,12 @@
 {
   "fwConfigs" : {
     "bios" : {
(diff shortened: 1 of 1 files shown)


arajeev-ARISTA:
Closing the pull request as the change is already incorporated.

PR #185, opened 2024-02-06
Do not read bytes ACL stat for Yuba Asic

NStetskovych:
Read only pakcet stats in  HwAclCounterTest for Yuba Asic


fboss/agent/hw/test/dataplane_tests/HwAclCounterTests.cpp
```diff
@@ -167,17 +167,23 @@ class HwAclCounterTest : public HwLinkStateDependentTest {
           EXPECT_EVENTUALLY_GE(aclPktCountAfter, aclPktCountBefore + 1);
           // At most we should get a pkt bump of 2
           EXPECT_EVENTUALLY_LE(aclPktCountAfter, aclPktCountBefore + 2);
-          EXPECT_EVENTUALLY_GE(
-              aclBytesCountAfter, aclBytesCountBefore + sizeOfPacketSent);
+          if (getAsic()->getAsicType() != cfg::AsicType::ASIC_TYPE_YUBA) {
```
srikrishnagopu, line 170: Instead, we should add a hw asic (HwAsic.h) feature and use that to guard this.

I will have Vasant fix this internally for you.

PR #198, opened 2024-05-29
meru800: Update fw_util configs to remove spidev binding

aalamsi22:
# Description

Update fw_util configs to remove spidev binding since it is no longer necessary as the spi driver already binds to spidev as specified by platform_manager.

# Test
Successfully tested upgrade/downgrade with fw_util


Scott8440:
Thanks, I will import this to FB once we have release a new bsp version that includes the related diff from fboss.bsp.arista which will enable this.

aalamsi22:
@Scott8440 , I will close this since the changes are included in the following PR #200

PR #200, opened 2024-05-31
meru fw_util: config file changes

arajeev-ARISTA:
### Summary of Changes
1. Use –noverify-all option with flashrom write.
This would skip verification of regions not included in the layout file to improve performance.

2. Leading zeros in the FW version numbers read through sysfs can cause them to be interpreted as octal numbers by shell. This change removes leading 0’s from the version numbers except when there is a ‘0x’ prefix. The sed commands to search and replace are written in a way that the leading 0’s are replaced only if the version number starts with a 0 followed by another digit.

3. Remove spidev binding since it is no longer necessary as the spi driver already binds to spidev as specified by platform_manager

### Testing
Verified fw upgrade/downgrade with fw_util

```
Change in programming time:
meru800bia smb_fpga:
real    4m55.654s => 0m55.890s
user    0m9.705s  => 0m9.748s
sys     2m55.965s => 0m26.562s

meru800bfa smb_fpga0:
real    4m57.611s => 0m58.643s
user    0m13.195s => 0m13.162s
sys     2m54.678s => 0m26.260s
```

### DIFF
--- a/fboss/platform/configs/meru800bfa/fw_util.json
+++ b/fboss/platform/configs/meru800bfa/fw_util.json
@@ -11,63 +11,63 @@
     },
     "scm_cpld": {
       "preUpgradeCmd": "",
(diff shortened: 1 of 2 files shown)


arajeev-ARISTA:
Closing since https://github.com/facebook/fboss/pull/236 has the same changes

PR #237, opened 2024-09-24
platformmanager: fix publish firmware version issue

QiuyunXie:
### Description
On Montblanc & Tahan & Janga device, running platform_manager will report firmware version read failed log as below 
image

When one CPLD has hwmon sensor path, but the cpld_ver/cpld_sub_ver file isn't at hwnon folder path. It will fail to find the cpld_ver/cpld_sub_ver file. Fix this issue, it's also suitable for "/run/devmap/cplds/FAN0_CPLD/hwmon/hwmon20/" case.

### Test log
image


mikechoifb (changes requested):
Thanks @QiuyunXie  Actually a Meta engineer is almost done with the fix for this bug (along with other issues he found) - you will have the fix for this soon. Can you please cancel this PR?

PR #243, opened 2024-09-25
fan_service: Adapt to fan LED is controlled by hardware

QiuyunXie:
### Description
For Minerva project, the FAN LEDs are controlled by FAN CPLD and compute blades do not participate in the control of the FAN LEDs. So "ledSysfsPath" entry can be empty. When "ledSysfsPath" is empty, no need to control the FAN LED.

### Test log
image
Attach the detailed log:
[Tahan_remove_fanLED_log.txt](https://github.com/user-attachments/files/17126727/Tahan_remove_fanLED_log.txt)


mikechoifb (changes requested):
.

PR #271, opened 2024-10-15
DARWIN: update fw_util.json file

adamcalabrigo:
# Description

Updates `fw_util.json` for DARWIN platform with the following changes:

1. Lowercase firmware targets.
2. Use relative paths for programming binaries.
3. Fix jam target for satellite CPLDs.

# Test Plan

Tested on Darwin with Linux kernel 6.4 and CentOS 9.

Verified programming of all firmware targets.

### DIFF
--- a/fboss/platform/configs/darwin/fw_util.json
+++ b/fboss/platform/configs/darwin/fw_util.json
@@ -1,6 +1,6 @@
 {
   "fwConfigs" : {
-    "BIOS" : {
+    "bios" : {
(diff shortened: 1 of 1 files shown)


joancaneus:
@adamcalabrigo  @aalamsi22 I actually need the name of the FPD in this diff to be lower cased. Upper case will cause issue.

joancaneus:
@aalamsi22  by FPD, I mean field programmable device. Things like BIOS, SMC_CPLD, SMB_FPGA.... etc.

PR #307, opened 2024-12-16
Put sensors for fans in the FCB slot

scottpaulsen:
Verified that platform manager loaded without error.

Verified that sensor_service_client shows the sensors.

### DIFF
--- a/fboss/platform/configs/morgan800cc/platform_manager.json
+++ b/fboss/platform/configs/morgan800cc/platform_manager.json
@@ -374,18 +374,6 @@
       ],
       "i2cDeviceConfigs": [
         {
-          "busName": "MCB_IOB_I2C_MASTER_14",
-          "address": "0x26",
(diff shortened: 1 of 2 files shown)


fboss/platform/configs/morgan800cc/platform_manager.json
```diff
@@ -459,7 +447,9 @@
         "FCB_SLOT@0": {
           "slotType": "FCB_SLOT",
           "outgoingI2cBusNames": [
-            "MCB_MUX_A@0"
+            "MCB_MUX_A@0",
+            "MCB_IOB_I2C_MASTER_14",
```
truongd3, line 451: Why do we need to add these 2 I2C buses? Does your PR relate to any issue?

scottpaulsen: Have to put the sensors in the right slot for the sensor service to have it setup per slot correctly.

PR #316, opened 2025-01-08
Darwin48V: Configure Aboot configuration on upgrade

geoffhanson:
When upgrading to the official Aboot 7.5.3 version, existing fw_util scripts configure the aboot_conf section from the Aboot image. Since boot methods are built into the image, this is not an issue for Darwin. But for Darwin48V, the DMI_BOARD_NAME needs to be configured in order for dmidecode to show the right thing.

In addition to the Darwin48V change, there is a change to the Darwin config to fix an issue where fw_util was not working with certain SPI chips. The issue was the and/or operators in the command did not contain parenthesis, so the chip type N25Q128..3E was being used unconditionally.

Test Plan

1. Tested the upgrade on a Darwin48V system. Verified that after the upgrade, dmidecode -t 2 showed DARWIN48V.

2. Tested the upgrade on a Darwin system. Verified that after the upgrade, dmidecode -t 2 showed DARWIN.

3. Tested the upgrade on a Darwin system with a MX25L12805D SPI chip. Verified the upgrade succeeded.

### DIFF
--- a/fboss/platform/configs/darwin/fw_util.json
+++ b/fboss/platform/configs/darwin/fw_util.json
@@ -1,7 +1,7 @@
 {
   "fwConfigs" : {
     "BIOS" : {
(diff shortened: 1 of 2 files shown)


geoffhanson:
Closing this as there are other PRs that supersede this change.

PR #315, opened 2025-01-08
Correcting temp sensor mapping to match HW for INT/EXT

sthenaracisco:

PR #317, opened 2025-01-10
Correcting temp sensor mapping to match HW for INT/EXT

sthenaracisco:
Correcting temperature sensors as per the specification

Test plan:

1) run paltform manager binary
2) run sensor_service 
3) sensor_service_test

observe the values of the mentioned sensors

Out put from Testbed
----


sthenaracisco:
@somasun This pR is not different from #315 315

somasun:
Closing since #315 has been merged

PR #309, opened 2024-12-18
Mp3:platform_manager:Validating and Waiting for Watchdog Device Creation.

karthik-gengan:
# Description:
  Implementing Watchdog device creation validation in platform_managerv  before PCI Explorer completion.
    1. Waiting for the device readiness.
    2. Granularly check every x ms until 5 seconds

# Motivation:
 In rare scenarios, the initialization process for a device can extend beyond the standard timeframe. This can cause the platform_manager to fail to locate the watchdog device in the specified location, resulting in an error message.

     I1216 08:09:08.409639  2801 ExplorationSummary.cpp:50] Explored montblanc with 2 unexpected errors and 0 expected errors...
     I1216 08:09:08.409647  2801 ExplorationSummary.cpp:58] Unexpected Failures in Device MCB_FAN_CPLD for PmUnit 
     MINIPACK3_MCB at SlotPath /
      I1216 08:09:08.409651  2801 ExplorationSummary.cpp:66] 1. Failed to explore I2C device MCB_FAN_CPLD at /. Failed to 
      resolve Watchdog CharDevPath. Reason: Couldn't find watchdog directory under /sys/bus/i2c/devices/15-0033
      I1216 08:09:08.409656  2801 ExplorationSummary.cpp:66] 2. Failed to create a symlink 
      /run/devmap/watchdogs/FAN_WATCHDOG for DevicePath /[MCB_FAN_CPLD]. Reason: Could not find CharDevPath for 
      /[MCB_FAN_CPLD] 
 

# Testcase:
  A) Delayed Device Registration
       **Simulate Delay:** Introduce a simulated delay in the driver code to mimic a scenario where device registration takes 
                                       longer than usual. 
        **Validation:** Implement a validation process that actively waits for the watchdog device to be successfully created. 
        **Assertion:** Verify that the platform_manager can locate and utilize the watchdog device within a reasonable timeframe, even with the simulated delay.

   B) The above procedure was executed with a reboot and power cycle.
   
[platform_manager_Before_fix.txt](https://github.com/user-attachments/files/18183519/platform_manager_Before_fix.txt)
[platform_manager_After_fix.txt](https://github.com/user-attachments/files/18183521/platform_manager_After_fix.txt)

PR #323, opened 2025-01-28
platform_manager:  Validating and Waiting for Watchdog Character Device Creation

karthik-gengan:
# Description:

 After creating the sysfs watchdog device, the corresponding character device entry in /dev/watchdog should be automatically generated. However, in some cases, the character device entry may take time to appear, which could cause the platform manager to fail.

 To address this, a validation check for the presence of the character device (/dev/watchdog) has been implemented.
 
>   [root@localhost ~]# ls -l /sys/bus/i2c/devices/15-0033/watchdog/ --full-time
> drwxr-xr-x 2 root root 0 2025-01-21 09:42:56.008935280 +0800 watchdog1 
> [root@localhost ~]# ls -l /dev/watchdog1 --full-time 
> crw------- 1 root root 246, 1 2025-01-21 09:42:56.030935280 +0800 /dev/watchdog1

 # Motivation:
 In rare scenarios, the initialization process for a device can extend beyond the standard timeframe. This can cause the platform_manager to fail to locate the watchdog device in the specified location, resulting in an error message.

`
I1207 16:37:55.981444 1201 ExplorationSummary.cpp:66] Failed to explore I2C device MCB_FAN_CPLD at /. Failed to resolve Watchdog CharDevPath. Reason: /dev/watchdog2 does not exist in the system `

# Testcase:
A) Delayed Device Registration
**Simulate Delay**: Introduce a simulated delay in the driver code to mimic a scenario where device registration takes longer than usual.
**Validation**: Implement a validation process that actively waits for the watchdog device to be successfully created.
**Assertion**: Verify that the platform_manager can locate and utilize the watchdog device within a reasonable timeframe, even with the simulated delay.
B) The above procedure was executed with a reboot and power cycle.

### DIFF
--- a/fboss/platform/platform_manager/Utils.cpp
+++ b/fboss/platform/platform_manager/Utils.cpp
@@ -108,7 +108,7 @@ std::string Utils::resolveWatchdogCharDevPath(const std::string& sysfsPath) {
               "Watchdog CharDevPath is not created. Waited for at most {}s",
               kWatchdogDevCreationWaitSecs.count()),
           kWatchdogDevCreationWaitSecs)) {
-  	throw std::runtime_error(fmt::format(
+    throw std::runtime_error(fmt::format(
         "{}. Reason: {} does not exist in the system", failMsg, charDevPath));
   }
   return charDevPath;
--- a/fboss/platform/platform_manager/Utils.cpp
+++ b/fboss/platform/platform_manager/Utils.cpp
@@ -102,8 +102,13 @@ std::string Utils::resolveWatchdogCharDevPath(const std::string& sysfsPath) {
         "{}. Reason: Couldn't find watchdog under {}", failMsg, sysfsPath));
   }
   auto charDevPath = fmt::format("/dev/watchdog{}", *watchdogNum);
-  if (!fs::exists(charDevPath)) {
-    throw std::runtime_error(fmt::format(
+  if (!Utils().checkDeviceReadiness(
+          [&]() -> bool { return fs::exists(charDevPath); },
+          fmt::format(
+              "Watchdog CharDevPath is not created. Waited for at most {}s",
+              kWatchdogDevCreationWaitSecs.count()),
+          kWatchdogDevCreationWaitSecs)) {
+  	throw std::runtime_error(fmt::format(
         "{}. Reason: {} does not exist in the system", failMsg, charDevPath));
   }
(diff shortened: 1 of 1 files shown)


fboss/platform/platform_manager/Utils.cpp
```diff
@@ -102,8 +102,13 @@ std::string Utils::resolveWatchdogCharDevPath(const std::string& sysfsPath) {
         "{}. Reason: Couldn't find watchdog under {}", failMsg, sysfsPath));
   }
   auto charDevPath = fmt::format("/dev/watchdog{}", *watchdogNum);
-  if (!fs::exists(charDevPath)) {
-    throw std::runtime_error(fmt::format(
+  if (!Utils().checkDeviceReadiness(
```
Scott8440, line 105: There is another `checkDeviceReadiness` on line 80. Do we need both of them? If so, can you explain the different situations in which each of these are required?

karthik-gengan: The resolveWatchdogCharDevPath function creates two entries for the watchdog timer:
1.  sysfs entry (e.g., /sys/bus/i2c/devices/15-0033/watchdog/)  
2. character device entry (e.g., /dev/watchdog1).

A previous issue (addressed in #309 ) involved the sysfs entry sometimes taking longer than expected to be created. 
This could cause problems during platform manager validation.

After extended stress testing and reboots, a similar timing issue was discovered with the character device creation. 
In rare instances, the character device would also take an unexpected time to be created again, impacting platform manager validation.

To resolve this, a `checkDeviceReadiness` function has been implemented to verify the character device's readiness, similar to how PR309 addressed the sysfs entry timing issue. This ensures the platform manager is validated successfully on all possible timing scenarios.

**Logs from our observation:**
[root@localhost ~]# ls -l /sys/bus/i2c/devices/15-0033/watchdog/ --full-time
drwxr-xr-x 2 root root 0 2025-01-21 09:42:56.008935280 +0800 watchdog1
[root@localhost ~]# ls -l /dev/watchdog1 --full-time
crw------- 1 root root 246, 1 2025-01-21 09:42:56.030935280 +0800 /dev/watchdog1

Scott8440:
Thanks for the explanation! Please fix the clang format errors and then I will merge

karthik-gengan:
Addressed clang format error...

PR #325, opened 2025-01-30
[Darwin]: sensor_service POS_3V3_ALW missing compute attribute

aalamsi22:
# Description

This fixes an issue where the sensor POS_3V3_ALW was throwing critical alarms due to a missing configuration with how the sensor value is computed.

```
POS_3V3_ALW 3273.000000 2025-01-28 14:29:21.0 PST
3.795000 2.800000
```

The new added `"compute": "@/1000.0"` will correctly calculate the sensor value.

# Testing
Value reported is within thresholds.


aalamsi22:
Changes included in #365

PR #331, opened 2025-01-31
EepromV6: Validating the ProductionState as per new fboss commit 

karthik-gengan:
# Description:
  A recent update to the EEPROM format (V6) in fboss (commit [18b03f1e74d78a302adbaa1bb754bc08569ebf67](https://github.com/facebook/fboss/commit/18b03f1e74d78a302adbaa1bb754bc08569ebf67)) changes the `ProductionState` field to a human-readable string.  This change doesn't affect the V5 EEPROM format, which uses a numerical representation. The existing code doesn't handle the new human-readable format introduced in V6.  Therefore, this change modifies the code to support both the numerical (V5) and human-readable (V6) formats for the `ProductionState` field.
image

# Motivation:
 In V6, the platform manager reports an error because it cannot handle the human-readable `ProductionState` field.
 This unsupported format causes the platform manager to fail and return an error.

> 1554 PlatformExplorer.cpp:349] Found ProductProductionState `<ABSENT>` ProductVersion `1` ProductSubVersion `2` in IDPROM /sys/bus/i2c/devices/16-0056/eeprom at /COME_SLOT@0
> 1554 PlatformExplorer.cpp:371] Found PmUnit name `NETLAKE` in IDPROM /sys/bus/i2c/devices/16-0056/eeprom at /COME_SLOT@0
> 1554 PlatformExplorer.cpp:380] SlotType PmUnitName: NETLAKE
> 1554 PlatformExplorer.cpp:390] Going with PmUnit name `NETLAKE` defined in config for /COME_SLOT@0
> 1554 DataStore.cpp:107] At SlotPath /COME_SLOT@0, unexpected partial versions: ProductProductionState `<ABSENT>` ProductVersion `1` ProductSubVersion `2`. Skipping updating PmUnit NETLAKE
> 1554 DataStore.cpp:117] At SlotPath /COME_SLOT@0, updating to PmUnit NETLAKE with
> 1554 DataStore.cpp:140] Resolved /COME_SLOT@0 to default PmUnitConfig of NETLAKE. No ProductSubversion was read from IDPROM at the slotPath.
**Refer to the complete log :**  
[platform_manager_GITHUB_ACTION.txt.log](https://github.com/user-attachments/files/18619160/platform_manager_GITHUB_ACTION.txt.log)

# Test Case:

1. The platform manager executes successfully with both V5 and V6 EEPROM formats.
2. The above test is performed after a power cycle.
**[Unit test Log:]**- 
[platform_manager.txt](https://github.com/user-attachments/files/18618230/platform_manager.txt)
[Weutil.txt](https://github.com/user-attachments/files/18618231/Weutil.txt)

### DIFF
--- a/fboss/platform/weutil/CachedFbossEepromParser.h
+++ b/fboss/platform/weutil/CachedFbossEepromParser.h
@@ -40,27 +40,22 @@ class CachedFbossEepromParser {
     for (const auto& [key, value] : getContents(eepromFilePath, offset)) {
       // < V6 - "Product Production State"
       // = V6 - "Production State"
-      if (key == "Product Production State" || key == "Production State") {
-        try {
-            // Attempt conversion to integer first.
-            return std::stoi(value);
-        } catch (const std::invalid_argument& e) {
-            // If integer conversion fails, check for string values.
-            if (value == "0") {
-                 throw std::runtime_error("Production State is not set!");
-            } else if (value == "EVT") {
-                 return 1;
-            } else if (value == "DVT") {
-                 return 2;
-            } else if (value == "PVT") {
-                 return 3;
-            } else if (value == "MP") {
-                 return 4;
-            } else {
-                 throw std::runtime_error("Unknown Production State: " + value); // Include the value in the error message
-             }
-        } catch (const std::out_of_range& e) {
-             throw std::runtime_error("Production State value out of range: " + value); // Handle potential out-of-range errors
+      if (key == "Product Production State"){
+        return std::stoi(value);
+      } else if (key == "Production State") {
+        // If integer conversion fails, check for string values.
+        if (value == "0") {
+          throw std::runtime_error("Production State is not set!");
+        } else if (value == "EVT") {
+           return 1;
+        } else if (value == "DVT") {
+           return 2;
+        } else if (value == "PVT") {
+           return 3;
+        } else if (value == "MP") {
+           return 4;
+        } else {
+           throw std::runtime_error("Unknown Production State!"); 
         }
       }
     }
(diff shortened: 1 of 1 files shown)


fboss/platform/weutil/CachedFbossEepromParser.h
```diff
@@ -40,8 +40,28 @@ class CachedFbossEepromParser {
...
+            } else if (value == "PVT") {
+                 return 3;
+            } else if (value == "MP") {
+                 return 4;
+            } else {
+                 throw std::runtime_error("Unknown Production State: " + value); // Include the value in the error message
+             }
+        } catch (const std::out_of_range& e) {
+             throw std::runtime_error("Production State value out of range: " + value); // Handle potential out-of-range errors
+        }
```
kimdo8736, line 64: just simply, `return std::stoi(value);`

We don't need try/catch here. Value being shown as 'EVT'/'DVT' is because the Parser converts the int to the human readable string.

That being said. Removing trailing whitespace in `"Production State"` is enough.

kimdo8736 (changes requested):

fboss/platform/weutil/CachedFbossEepromParser.h
```diff
@@ -40,27 +40,22 @@ class CachedFbossEepromParser {
...
-            } else {
-                 throw std::runtime_error("Unknown Production State: " + value); // Include the value in the error message
-             }
-        } catch (const std::out_of_range& e) {
-             throw std::runtime_error("Production State value out of range: " + value); // Handle potential out-of-range errors
+      if (key == "Product Production State"){
+        return std::stoi(value);
+      } else if (key == "Production State") {
+        // If integer conversion fails, check for string values.
+        if (value == "0") {
```
jdgkim, line 47: We can have conditions to check for `EVT | DVT | PVT | MP`. Otherwise, throw an exception ("unexpected production state")

karthik-gengan:
@somasun : Move the ProductionState conversion to the presentation layer. Hence, this change is not required. 
https://github.com/facebook/fboss/commit/fda71c2a8c2d7a03895960091520d5838793ec2a

PR #337, opened 2025-02-18
janga: config: added patch files for rack fan service

zhongedward:
Description
This PR is for janga & tahan rack fan service path files.

Motivation
1.Based on quanta's input fan configuration file, I created a fan_service patch file and a platform_manager patch file based on the current files in the fboss repo.
2.In the patch files, I mainly added the gpio symbolic link.

image

image

3.Note:
The fan_service patch file is a transitional version, quanta will change some parameters in the future, I will keep a close eye on it.

Test Plan
1.Tested both in janga & tahan rack machines.

image

2.Test log as follows:

[tahan_rack_test_log_2_17.txt](https://github.com/user-attachments/files/18840554/tahan_rack_test_log_2_17.txt)
[janga_rack_test_log_2_17.txt](https://github.com/user-attachments/files/18840559/janga_rack_test_log_2_17.txt)


kimdo8736:
I was expecting fan_service.json and platform_manager.json from our discussion in workplace.

The PR contains .patch files which has content diffs. Can you please update with new json files?

zhongedward:
@kimdo8736  Hi kim, I have updated the new json files, please review, thanks

kimdo8736:
@zhongedward,

I see two changes in this PR:

1. PWMs values changes per temperatures / sensor name change
2. GPIO presence introduction.

Can you separate the PRs? I can import #1 changes to internal Meta. We may need to wait #2 as the Janga/Tahan PVT units in our lab is still in test fixture.

zhongedward:
@kimdo8736 Hi kim, yes, fan_service.json is provided by quanta and it is only used for rack macine, while platform_manager.json is Matching, and their 2 files should be used together on the rack machine, not separately.
Sorry, I'm a bit confused by your suggestion, are you saying I need to separate the GPIO part (both in fan_service.json and platform_manager.json)? If so, can it be used in the test fixture? Could you please give me some more clear instructions? thanks!

zhongedward:
I'm thinking again that it either doesn't import, or if it does, it needs to be imported together and then applied to the rack machine. Because the test fixure's fan config file and the rack's fan config file are separate. Please correct me if I'm wrong, thanks!

zhongedward:
Based on the replies in the workspace, I've turned off PR for now. i'll turn it back on when Quanta is done.
image

PR #351, opened 2025-03-19
fw_util config file for Darwin48V

arajeev-ARISTA:
### Summary
fw_util config file for Darwin48V compliant with the recent fw_util refactoring
### Testing
Verified fw upgrade/downgrade with fw_util

### DIFF
--- a/fboss/platform/configs/darwin48v/fw_util.json
+++ b/fboss/platform/configs/darwin48v/fw_util.json
@@ -1,57 +1,304 @@
 {
   "newFwConfigs": {
-    "BIOS": {
-      "version": {
-        "versionType": "full_command",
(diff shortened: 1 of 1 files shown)


arajeev-ARISTA:
> Please double check and verify the file syntax. This fails to pass our internal tests because there is some parsing error with this file. Please add a full test plan with actual pastes/screenshots/etc.

The error is from the chip name `GD25Q16(B)` as defined in the flashrom source code.

Scott8440:
The failures from ConfigLib were due to a delimiter just being "()" which caused the string to be closed when this chip name was read. Will merge this PR once my change to ConfigLib is merged.

Scott8440:
@arajeev-ARISTA I imagine you ran into this problem when testing. In the future can you please include this information in the pull request itself?

arajeev-ARISTA:
> @arajeev-ARISTA I imagine you ran into this problem when testing. In the future can you please include this information in the pull request itself?

Will do, thanks!

PR #352, opened 2025-03-19
enable fw_util for Darwin platforms

arajeev-ARISTA:
### Summary
Removed the check which restricted the use of fw_util in Darwin and Darwin48V systems.

Darwin firmware versions will still be printed using the special function `printDarwinVersion()`. This ensures that fw_util remains compliant with Darwin systems that have older BSP versions, which use a different type of firmware version sysfs entry.
### Testing
Verified fw upgrade/downgrade with fw_util in a Darwin48V system. Also verified reading firmware versions in a Darwin system.

### DIFF
--- a/fboss/platform/fw_util/FwUtilImpl.cpp
+++ b/fboss/platform/fw_util/FwUtilImpl.cpp
@@ -73,19 +73,6 @@ void FwUtilImpl::doFirmwareAction(
   auto fwConfig = iter->second;
 
   if (action == "program") {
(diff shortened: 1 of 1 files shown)


ezeob002:
@arajeev-ARISTA, could you please make the necessary changes in the FwUtilVersionHandler class?. We restructured the code base.

arajeev-ARISTA:
> @arajeev-ARISTA, could you please make the necessary changes in the FwUtilVersionHandler class?. We restructured the code base.

Done!

PR #353, opened 2025-03-19
added fw_util support to program custom contents in a flash device

arajeev-ARISTA:
### Background
During BIOS upgrade in Darwin48V systems, `DMI_BOARD_NAME` needs to be updated through Aboot aconf programming. Previously we achieved this by adding the aconf programming logic to the `fw_util` config file.
However, a recent refactoring of the `fw_util` prevents adding arbitrary shell commands in the config file. This change addresses the issue by adding the support to program custom contents in a flash device which is updated through `flashrom`.
### Summary of changes

- Added two optional arguments `custom_content` and `custom_content_offset` in the Thrift description for `FlashromConfig`
- If a `flashrom` command description with both `custom_content` and `custom_content_offset` are found in the `fw_util` config file, a temporary binary file will be created. The binary file will have `custom_content` written at `custom_content_offset`. A `flashrom` command would then be constructed using this temporary binary file instead of the firmware image. The regions in the flash memory which will be updated are restricted by specifying the layout
- The command to program the custom content will be in addition to the existing command to program the firmware binary. We are able to specify multiple commands because it’s allowed by the thrift description for upgrade operation

### Testing
Upgraded BIOS in a Darwin48V system with fw_util and verified that DMI_BOARD_NAME was programmed.

### DIFF
--- a/fboss/platform/fw_util/FwUtilFlashrom.cpp
+++ b/fboss/platform/fw_util/FwUtilFlashrom.cpp
@@ -64,9 +64,10 @@ void FwUtilImpl::addCommonFlashromArgs(
 
   if (flashromConfig.custom_content().has_value() &&
       flashromConfig.custom_content_offset().has_value()) {
-     /* Create a file with custom_content written at custom_content_offset.
-        The flashrom command would use this file instead of the firmware image */
-    const std::string customContentFileName = "/tmp/" + fpd + "_custom_content.bin";
+    /* Create a file with custom_content written at custom_content_offset.
+       The flashrom command would use this file instead of the firmware image */
+    const std::string customContentFileName =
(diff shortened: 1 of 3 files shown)


joancaneus:
@arajeev-ARISTA -- We use the linter below to format code at meta

Lint code: CLANGFORMAT
Lint name: format

There are some lints failing on this diff which I will format, but for future reference (when you done with those PR) can you please look into how you can possibly use that linter as well?

arajeev-ARISTA:
> @arajeev-ARISTA -- We use the linter below to format code at meta
> 
> Lint code: CLANGFORMAT Lint name: format
> 
> There are some lints failing on this diff which I will format, but for future reference (when you done with those PR) can you please look into how you can possibly use that linter as well?

Will do, thanks!

PR #416, opened 2025-04-25
Migrate wedge_qsfp_util to use utils::createQsfpServiceClient

paulcruz74:
Summary:
This change is in order to test a hypothesis that some difference in client
creation between QsfpClient.cpp and ThriftServiceClient.cpp is responsible for
additional latency when executing thrift requests in an OSS context.

Differential Revision: D73670141


srikrishnagopu:

PR #423, opened 2025-05-05
[meru800bia] Update Max6581 temperature thresholds based on hardware limits

dzarista:
Broadcom guidance is that max excursion temp should be under 115 for core/serdes and under 105 for HBM. The thresholds are updated based on these values and the accuracy of Max6581.

No sensor_service or platform_manager issues.


somasun:
@dzarista  - can you please further elaborate in the PR on why these thresholds are getting updated now?

dzarista:
> @dzarista - can you please further elaborate in the PR on why these thresholds are getting updated now?

Updated the description. Please let me know if more info is needed. Thanks.

PR #424, opened 2025-05-05
[meru800biab] Remove unnecessary references

dzarista:
Platform logic has been changed so that meru800biab uses meru800bia configs - https://github.com/facebook/fboss/pull/381 

This PR simply updates the codebase to remove unnecessary meru800biab references, which happens to only be in one place, as the other references are needed for the agent/qsfp_service to distinguish between products.

Tested that there are no issues on platform_manager and qsfp_service.


somasun:
thanks for this cleanup

PR #445, opened 2025-05-21
scripts: Don't generate fboss_bins.tar.zst.tar.gz

timn-nexthop:
`package-fboss.py --compress` generated 2 tar files:
1. fboss_bins.tar.zst.tar.gz, which is really just a .tar.gz of all the binaries. Because it was using single-threaded compression this took 10 minutes.
2. fboss_bins.tar.zst. Generating this takes about 16 seconds.

10 minutes is way too long to wait for a build to finish, and having the same data in 2 formats isn't helpful.

Step 2 was added in #251, but it left step 1 in place. Remove step 1. You do now need zstd support to extract the archive, but zstd has been around for a long time now.


timn-nexthop:
https://github.com/facebook/fboss/commit/2528b4ebbdf210fb08a2700c59273afb9a46577d already made this change just before I made my PR, but I didn't notice because I hadn't updated to latest main yet.

PR #450, opened 2025-05-28
Platform_manager: fix platform name in config does not match the inferred platform name from bios

joywu-coder:
### Description
The platform name in the config does not match the inferred platform name from the BIOS during platform manager startup.

image
### Motivation
The platform alias transformation for a new project(icecube or icetray) is not supported in the current fboss code. The code fix is to add the transformation logic(ICECUBE-->ICECUBE800BC, ICETRAY-->ICETRAY800BC).
### Test Plan
Run platform_manager and check if the platform alias transformation is working.

image
image
image

[pm_verification_log.txt](https://github.com/user-attachments/files/20472157/pm_verification_log.txt)

### DIFF
--- a/fboss/platform/helpers/PlatformNameLib.cpp
+++ b/fboss/platform/helpers/PlatformNameLib.cpp
@@ -39,6 +39,14 @@ std::string sanitizePlatformName(const std::string& platformNameFromBios) {
     return "MERU800BIA";
   }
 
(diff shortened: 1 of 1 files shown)


joywu-coder:
Meta suggests that BIOS side will implement the transformation of platform name, so close the PR.

PR #441, opened 2025-05-16
[meru] Update platform_manager configs to load adc rail driver

dzarista:
Loads the adc rail driver for meru800bia scm and smb and meru800bfa scm.

Tested that platform manager successfully loads the driver for all applicable devices and that all voltage readings are correctly showing up in `sensors`.

PR #474, opened 2025-06-14
MERU800BIA/BFA: Initialize PSUs with write-protect

aalamsi22:
# Description

Some registers when read, cause the psu to raise an error through the status_cml register at 0x7e. This is normal for some PSUs with unsupported registers. However, problems arise when the write-protect is disabled, the linux pmbus driver sees the status register reporting an error and decides to ignore the valid values it reads.

# Motivation

While this issue shouldn't hit because the WP register is enabled by default when the PSU turns on, this change ensures that WP is enabled in the rare cases where a WP may have been disabled manually or by another process prior to setting up fboss on the system.

# Testing
Manually disabled WP and confirmed platform manager enables it again. All expected psu sensors are read successfully.

### DIFF
--- a/fboss/platform/configs/meru800bfa/platform_manager.json
+++ b/fboss/platform/configs/meru800bfa/platform_manager.json
@@ -4104,7 +4104,9 @@
           "initRegSettings": [
             {
               "regOffset": 16,
-              "ioBuf": [-128]
+              "ioBuf": [
+                -128
+              ]
             }
           ]
         }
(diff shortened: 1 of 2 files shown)


joancaneus:
@aalamsi22 Does this PR change require BSP version 0.7.10 ?

I am just trying to understand if the same comments in https://github.com/facebook/fboss/pull/441 also apply to it.

aalamsi22:
> @aalamsi22 Does this PR change require BSP version 0.7.10 ?
> 
> I am just trying to understand if the same comments in #441 also apply to it.

No, this is independent of the bsp.

srikrishnagopu:

PR #477, opened 2025-06-16
Build: add support package for caching in S3.

timn-nexthop:
The FBOSS build builds several 3rd party applications before it starts its own build. There is an existing mechanism to cache that build, but no default implementation. It saves 10-20 minutes per build.

This change adds various features that combine to let you pass a cache configuration file into the container where the build takes place. While this implementation supports S3, it would be easy to extend the code to support other cache backends.

## Details

Modified cache.py to create an S3 cache if the appropriate configuration is present. The configuration consists of a ConfigParser file with the following keys in an [s3] section:
* region, region the S3 bucket is in
* bucket, name of S3 bucket that contains the cache
* prefix, prefix for every object name stored in the bucket (optional)
* aws_access_key_id
* aws_secret_access_key

Modified docker-build.py to take 3 new arguments:
* --schedule-type, passed on to getdeps.py. It's needed because getdeps.py won't use a cache unless the schedule type is "continuous".
* --extras-dir, pointing to an extra local directory that will be mounted into the container at /var/extras
* --cache-config, pointing at a configuration file in the container that contains details on how to access the S3 cache

Modified getdeps.py to take one new argument:
* --cache-config, which expects a ConfigParser file describing the cache.

The intent is to use those options together to enable caching e.g. as follows:
```
docker-build.py --schedule-type continuous --extras-dir ~/extras --cache-config /var/extras/cache.cfg
```

Then ~/extras/cache.cfg might contain:
```
[s3]
bucket = fboss-cache
region = us-east-2
aws_access_key_id = 12345
aws_secret_access_key = 67890
```

## Backwards Compatibility

Note that this changes getdeps.cache.create_cache() to take an optional argument. The caller is modified so that if that argument would be None, then no argument is passed at all to keep things backwards compatible for getdeps.facebook monkey patching create_cache().


Scott8440:
@KevinYakar Can you take a look at this? I know you were looking into speeding up OSS builds. Not sure if this is the direction you want to go in.

KevinYakar:
> @KevinYakar Can you take a look at this? I know you were looking into speeding up OSS builds. Not sure if this is the direction you want to go in.

Hi @Scott8440 yes I did take a look at this. It is useful but we may not head this direction as it requires modifying getdeps which is maintained by a different team. That said we are still considering it.

There has been some progress in reducing build time by caching dependencies in a different way though and I hope to get it implemented sometime soon.

Scott8440:
We might consider this later, but are currently looking into different caching methods.

PR #465, opened 2025-06-13
tahansb: config: created platform_manager FBOSS config for EVT1

CliffyWong:
**Description**
This PR is for tahansb platform manager config file

**Motivation**
Based on the TAHANSB EVT1 hardware specification, there are four pmUnits:

- TAHANSB_MCB
- BMC
- NETLAKE
- SMB

The corresponding configurations for these pmUnits should be added to the platform manager configuration file.
image

**Test Plan**
1.The correctness of the format has been verified on this website https://jsonlint.com/ 
2. Used jq command to pretty the format
3. Compilation and config validation have passed
image

Building Log:
[build.txt](https://github.com/user-attachments/files/20874978/build_20250624.txt)

PR #480, opened 2025-06-19
tahansb: config: created fan_service FBOSS config for EVT1

CliffyWong:
**Description**

This PR is for tahansb fan_service config file

**Motivation**

Based on the TAHANSB EVT1 hardware specification, the configurations for fan service control and the sensors determine the fan speed should be configured in the FBOSS fan service config file.

image

**Test Plan**

1.The correctness of the format has been verified on this website https://jsonlint.com/ 
2. Used jq command to pretty the format
3. Compilation and config validation have passed
image

Building log:
[build.txt](https://github.com/user-attachments/files/20811797/fan.log)

### DIFF
--- a/fboss/platform/configs/tahansb/fan_service.json
+++ b/fboss/platform/configs/tahansb/fan_service.json
@@ -1,0 +1,1325 @@
+{
+  "pwmBoostOnNumDeadFan": 2,
+  "pwmBoostOnNumDeadSensor": 0,
+  "pwmBoostOnNoQsfpAfterInSec": 0,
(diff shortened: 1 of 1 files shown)


Scott8440:
@CliffyWong  Can you please add logs from fan_service showing that the service runs as expected with this config?

CliffyWong:
> @CliffyWong Can you please add logs from fan_service showing that the service runs as expected with this config?

Hi, pls see my comments in https://github.com/facebook/fboss/pull/465

PR #490, opened 2025-07-10
fw_util: Don’t change Darwin firmware target names to lowercase

arajeev-ARISTA:
### Summary
fw_util configuration file for Darwin has firmware target names in uppercase to function correctly with the AnR tool. Hence, don’t change Darwin firmware target names to lower case in fw_util.
### Testing
Couldn’t directly test this change because the `--config_file` option is not working in FBOSS OSS right now. Tested by modifying an older version of fw_util.

### DIFF
--- a/fboss/platform/fw_util/fw_util.cpp
+++ b/fboss/platform/fw_util/fw_util.cpp
@@ -11,6 +11,7 @@
 #include "fboss/platform/fw_util/FwUtilImpl.h"
 #include "fboss/platform/fw_util/fw_util_helpers.h"
 #include "fboss/platform/helpers/InitCli.h"
+#include "fboss/platform/helpers/PlatformNameLib.h"
 
 using namespace facebook::fboss::platform::fw_util;
 using namespace facebook::fboss::platform;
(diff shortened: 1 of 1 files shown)


Scott8440:
Hi @arajeev-ARISTA, can you give me some more context on the ANR failure? I just want to make sure this is the best way to fix the problem

arajeev-ARISTA:
> Hi @arajeev-ARISTA, can you give me some more context on the ANR failure? I just want to make sure this is the best way to fix the problem

Hello Scott, I don't have access to the ANR tool. The target names in the config file were originally in lower case. They were changed to upper case citing compatibility issue with ANR - but it broke fw_util functionality in FBOSS OSS.

Scott8440:
@arajeev-ARISTA Ok that makes sense. I was confused I thought the failure was in AnR. I'll take a look at this.

Scott8440:
@arajeev-ARISTA Thanks for the submission. I'm going to implement a more general fix to this problem so that we don't have to add a special case for certain platforms.

PR #500, opened 2025-07-30
Meru800biac: define new platform, add platform and port mapping

huyc-arista:
# Description

> NOTE: This is part of a series of PRs to add support for Meru800biac. The dependency chain of the PRs is as follows:
> 
> Meru800biac: define new platform, add platform and port mapping #500
> Meru800biac: add bsp mapping, platform mapping, qsfp_service, led_service #501 - Depends on #500
> Meru800biac: add PM, fan_service, sensor_service and wedge_qsfp_util support #502 - Depends on #501

- Add PlatformType `PLATFORM_MERU800BIAC = 40` to fboss_common.thrift
- Define the string definition for the platform type
- Updated PlatformProductInfo.cpp to identify meru800biac
- Reference meru800bia platform_mapping and platform_port_mapping


fboss/lib/if/fboss_common.thrift
```diff
@@ -45,6 +45,7 @@ enum PlatformType {
   PLATFORM_DARWIN48V = 37,
   PLATFORM_MINIPACK3N = 38,
   PLATFORM_ICECUBE800BC = 39,
+  PLATFORM_MERU800BIAC = 41,
```
harshitgulati18: Could you change this to 40? You can add PLATFORM_GLATH05A_64O=41 later when you are ready to make those changes

huyc-arista: Updated

harshitgulati18:
Running into the following error while building. Please fix

fboss/agent/platforms/common/PlatformMappingUtils.cpp:69:11: error: enumeration value 'PLATFORM_MERU800BIAC' not handled in switch [-Werror,-Wswitch]
   69 |   switch (type) {

PR #508, opened 2025-08-11
[showtech] Ignore scanning certain i2c buses for meru800bia and meru800bfa

huyc-arista:
## Background
Currently showtech service scans every available i2c bus, this can take very long on some unused buses. Adding a way to skip specific based on showtech_configs. This PR also remove showtech verbose mode.

## Changes
- removed verbose mode
- added showtech I2cHelper class
- added showtech_configs for Meru800bia, Meru800bfa, Darwin and Darwin48v
- added `i2cBusIgnore` to showtech_configs, skip scanning these bus based on configs

## Testing
- Manually ran showtech and checked output on Meru800bia, Meru800bfa devices


Scott8440:
@huyc-arista How much of a difference in run time does this make?

If this change is worth it (based on how much time it saves) I wonder if a more general solution would be better. One thing I'm considering something like using the platform_manager.json config to determine which i2c busses have no devices attached and then just skipping all of those. Might be more work than it's worth, but worth considering @somasun

huyc-arista:
@Scott8440 I observed just the i2c scans for meru800bia went from 2m35s to 35s and meru800bfa from 7m to 30s. Skipping i2c-0 saves 2m. On meru800bfa we also also skip busses with no device or just a xcvr.

@dzarista because he originally implemented this for our internal showtech.

Side question: Is there an easy way of using platform_manager.json config to find i2c busses for specific devices e.g. busses for psus?

Scott8440:
> @Scott8440 I observed just the i2c scans for meru800bia went from 2m35s to 35s and meru800bfa from 7m to 30s. Skipping i2c-0 saves 2m. On meru800bfa we also also skip busses with no device or just a xcvr.

Ok yeah that definitely seems worth it.

> Side question: Is there an easy way of using platform_manager.json config to find i2c busses for specific devices e.g. busses for psus?

Unfortunately I don't think it's easy to do this as of now. The main problem is that you can have `INCOMING` i2c adapters which means that to know which adapter a device is connected to you might have to "traverse" the whole platform manager config. 

I think this comes down to what information you're interested in getting from i2cdetect. Potentially you could use `ls /sys/bus/i2c/devices` to find which busses have devices attached to them and then only scan those. But that seems like you would miss out on information if there was a problem that caused certain devices to not be created.

My main concern is that it would be really easy to forget to update this config if we modify the platform_manager config and move a device from one channel to another. Then you could end up skipping valuable information in showtech.

huyc-arista:
For now are we ok with ignoring just i2c-0, `"SMBus iSMT adapter at <addr>"`? This would save about 2 minutes on both systems.

I can revisit the other buses on meru800bfa when there is a better way of determining empty busses.

Scott8440:
> For now are we ok with ignoring just i2c-0, `"SMBus iSMT adapter at <addr>"`? This would save about 2 minutes on both systems.

Yes that should be fine for now.

> I can revisit the other buses on meru800bfa when there is a better way of determining empty busses.

Please file an issue to follow up with this request

PR #509, opened 2025-08-12
fboss:fan_service: fix coredump issue

QiuyunXie:
Description
After debugging, I found that starting from this commit 19e64dd2a71fc5608bc26479b5b2020c8ee5bdfd, a coredump issue will occur after running the fan service for a period (30 mins+). This issue arises after the introduction of the getAsicTempData() function. It's likely caused by a conflict with the existing getSensorData() function during multi-threaded asynchronous I/O operations.

Motivation
To avoid coredump issue caused by conflicts in multi-threaded asynchronous I/O operations, the following code modifications were made:

1. Remove thread_

1. Use synchronous Thrift calls instead

I also modified the agentTempThriftPort_ to 5959. The original "5972" is the fanService Thrift Port number. The function of get ASIC temperature through agent thrift port still need to verified later.

Test Plan
Running fan_service around 24 hours, the coredump issue not reproduced.
Attach partial test log: 
[TH5_DVT_fanservice_testlog.txt](https://github.com/user-attachments/files/21736685/TH5_DVT_fanservice_testlog.txt)


mikechoifb:
@QiuyunXie , thanks for the fix! If I understand correctly, your change in datafetcher.cpp is the actual fix, and other changes in BSP.h and BSP.cpp is code cleanup.  

If fact, Meta is using another set of the code that is not visible to Celestica, and the change in bsp.h and bsp.cpp are causing some problem. (we still want to use asynchronous thrift call in Meta's code base, as it's not causing any problem with Meta's library, and Meta internally use asynchronous calls.) Also, we have reserved the thrift port number, and do not want to change.

Can you do me a favor?  Can you try running your 24 hour tests again only with the change in "datafetcher.cpp", and see whether the service can continue running for 24 hours?   If the change in "datafetcher.cpp" is enough to prevent crash, can you revise the PR to ONLY include the change in datafetcher.cpp?

Thank you!

QiuyunXie:
@mikechoifb When BSP.h and BSP.cpp remain unchanged and only datafetcher.cpp is modified to use a synchronous Thrift call, the fan_service cannot run properly. Running the fan_service will immediately cause a coredump. 
image

somasun:
@QiuyunXie - using multiple eventbases to drive the same thread is unusual.  Can you please explain the details of the problem you were facing?

QiuyunXie:
> @QiuyunXie - using multiple eventbases to drive the same thread is unusual. Can you please explain the details of the problem you were facing?

@somasun Actually I didn't use multiple eventbases to drive the same thread. I give up using the "thread_" which is created in Bsp.h and Bsp.cpp. I create local EventBase variables "evbSensorInternal" and "evbAsicInternal" in 2 functions separately and did not use the local EventBase variables to drive thread.

The problem I am facing is that the fan service will coredump after running about 30mins. Below is the coredump info: 
image
Here is the detailed information reported in workplace: https://fb.workplace.com/groups/264616536023347/permalink/781414707676858/

PR #522, opened 2025-08-23
Bind OSFP ISL68226 devices to the bp4a_isl68226 driver

arajeev-ARISTA:
This PR updates meru800bfa platform manager config to bind the OSFP power ISL68226 devices to the `bp4a_isl68226` driver instead of the `isl68226` driver.

`bp4a_isl68226` driver now has a bug fix(BSP kmods version v0.7.14) which updates incorrect inductor config values on OSFP D ISL68226. This change is needed for the bug fix to take effect.


Scott8440:
We added this change in an internal diff to make rollout easier

PR #526, opened 2025-08-28
Download autoconf from github instead of gnu.org

KevinYakar:
Summary: We currently download the autoconf tarball from gnu.org, but it has had outages of variying timeframes. It interrupts our github workflows for too long, so we should download from the github repo instead.

Differential Revision: D81247169


joseph5wu:
I don't think we need such change from this old PR

PR #538, opened 2025-09-09
Mount fboss in Docker container instead of copying

KevinYakar:
Summary:
When building the fboss image, we used to copy the contents of the fboss repo on the host machine into the container. The user would have to re-build the image each time they make changes on their host machine in order to test them. We should instead mount the repository for ease of use and so that changes can persist on the host machine.

We still need to copy the fboss repo during image creation so we can use scripts for fetching dependencies, but this will be deleted afterwards so that the user will be using the mounted repository when entering the container.

Reviewed By: joseph5wu

Differential Revision: D81816484

### DIFF
--- a/fboss/oss/scripts/github_actions/docker-unittest.py
+++ b/fboss/oss/scripts/github_actions/docker-unittest.py
@@ -18,6 +18,7 @@ FBOSS_CONTAINER_NAME = "FBOSS_BUILD_CONTAINER"
 
 TEST_PATH_REGEX = re.compile(".*tests?$")
 HW_TEST_PATH_REGEX = re.compile(".*hw_tests?$")
+CONTAINER_WORKDIR = "/var/FBOSS/fboss"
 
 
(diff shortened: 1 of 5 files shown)


fboss/oss/scripts/github_actions/docker-unittest.py
```diff
@@ -29,6 +30,28 @@ def get_docker_path():
...
+
+
+def use_stable_hashes():
+    cwd = os.getcwd()
+    os.chdir(get_repo_path())
+    cmd = [
+        "rm",
+        "-rf",
+        "build/deps/github_hashes/",
+        "&&",
```
benoit-nexthop, line 46: There is a bug here. The `&&` is not evaluated because the command is not run with `shell=True`. This means that the latest stable hashes are not used when `use_stable_hashes()` is called, yikes.

Either this needs to be broken down in two `subprocess.run()` calls, or `shell=True` needs to be used.

Fix proposal in #595

PR #559, opened 2025-09-23
[MERU800BFA]: Increased fan_service ttable granularity

aalamsi22:
# Description

Increased granularity to the temperature tables by adding more points to help reduce temperature oscillations.

- More points were added to the lower half of the optics table to prevent large pwm changes across small temperature changes
- Same idea was applied to the asic zone

# Testing

The peak-to-peak temperature measurements observed to be relatively lower with the added granularity to the table as show in the tables below.  

## BEFORE
Screenshot 2025-09-25 at 7 38 21 PM

## AFTER
Screenshot 2025-09-25 at 7 37 34 PM


joancaneus:
@aalamsi22  how was this change tested to ensure it achieve the desired goal?

aalamsi22:
> @aalamsi22 how was this change tested to ensure it achieve the desired goal?

We ran multiple tests on two different systems both with the old and new profile. Observed improved stability via a lower peak-to-peak temperature swings. The increased granularity removed the large jumps in fan PWM that occured in small temperature changes.

PR #562, opened 2025-09-23
Icetea: led_service: fix the led config mode not supported in led_service_hw_test

joywu-coder:
# Description

1. Resolve the issue with 'defaultCommandLineArgs mode' not being supported for the LED service hardware test.
2. Change the code to support led_service_hw_test for the "Not Subscribed to FSDB Scenario" when the option `ledControlledThroughService` is set to false in `led.conf`.

# Motivation

1. We know there are two methods for identifying the platform type: one is through the field "Product Name" in the file located at /var/facebook/fboss/fruid.json, and the other is by checking the "defaultCommandLineArgs mode" in the led.conf file. We found that the current code for the led hw test does not support parsing the option "mode".
2. Need to run led_service_hw_test to verify that the LEDs on the switch are working correctly.

# Test Plan
1. Compilation has passed for led_service, led_service_hw_test.
4. Configure the led.config. 
[led_conf.txt](https://github.com/user-attachments/files/22488120/led_conf.txt)
3. Run led_service_hw_test for LedServiceTest.checkLedBlinking case, then check the logs.
image
image

[led_service_hw_test_log.txt](https://github.com/user-attachments/files/22488255/led_service_hw_test_log.txt)


fboss/led_service/hw_test/LedServiceTest.cpp
```diff
@@ -30,7 +48,7 @@ void LedServiceTest::SetUp() {
   // and the LED is managed by service now
   ledManager_ = getLedEnsemble()->getLedManager();
   CHECK_NE(ledManager_, nullptr);
-  CHECK(ledManager_->isLedControlledThroughService());
+  // CHECK(ledManager_->isLedControlledThroughService());
```
harshitgulati18: We don't need to comment out this line. FSDB service is not required to be run during led_service_hw_test. We mock the data we receive from FSDB during these tests.

Do you still see a crash when you set ledControlledThroughService to true in the led config? Config should look like this -

{
  "defaultCommandLineArgs": {
    "mode": "icetea800bc"
  },
  "ledControlledThroughService": true
}

shiva-menta: @harshitgulati18 Yes it fails when ledControlledThroughService is true.

joywu-coder: > @harshitgulati18 Yes it fails when ledControlledThroughService is true.

Yes, the crash is still there if setting ledControlledThroughService to true. This issue has been aligned with @shiva-menta in the workplace.  @shiva-menta Thanks a lot for the reply.

shiva-menta: @joywu-coder After some discussion with Harshit, we realized that there is some crash happening in FSDB mode, but it's due to some unexpected cleanup of FSDB stubs. Can you remove the commenting out of this line for now?

joywu-coder: > Contributor

@shiva-menta Yes, I confirmed with Harshit yesterday. The new changes have been submitted. Please check. Thank you.

PR #569, opened 2025-09-25
[MERU800BFA]: Change voltage readings to mili in sensor_service

aalamsi22:
# Description

Decimal points are rounded down in ODS, which limit debugging certain sensors where the decimals hold significance. This has been an issue when debugging vrm issues were certain voltages hold a value betwen 0-1V. 

The proposed solution is to display voltages with thresholds that can fall below 1V as millivolts with "_MILLIVOLT" in the sensor name.

# Testing
All sensor readings are within thresholds.

### DIFF
--- a/fboss/platform/configs/meru800bfa/sensor_service.json
+++ b/fboss/platform/configs/meru800bfa/sensor_service.json
@@ -226,14 +226,13 @@
           "compute": "@/1000.0"
         },
         {
-          "name": "SMB_VRM_R3R0_CORE_VOUT_0V75",
+          "name": "SMB_VRM_R3R0_CORE_VOUT_0V75_MILLIVOLT",
           "sysfsPath": "/run/devmap/sensors/SMB_RAA228926_R3R0_CORE/in3_input",
           "type": 1,
           "thresholds": {
-            "upperCriticalVal": 1.008,
-            "lowerCriticalVal": 0.504
-          },
-          "compute": "@/1000.0"
+            "upperCriticalVal": 1008.0,
+            "lowerCriticalVal": 504.0
+          }
         },
         {
           "name": "SMB_VRM_R3R0_CORE_TEMP",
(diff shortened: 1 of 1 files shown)


somasun:
hmmm, we rely on this being Watts as the standard - https://github.com/facebook/fboss/blob/main/fboss/platform/sensor_service/if/sensor_config.thrift?#L11

> The proposed solution is to display all voltage readings in Milivolts.

Can you please shed more context on this?  cc: @joancaneus

aalamsi22:
> hmmm, we rely on this being Watts as the standard - https://github.com/facebook/fboss/blob/main/fboss/platform/sensor_service/if/sensor_config.thrift?#L11
> 
> > The proposed solution is to display all voltage readings in Milivolts.
> 
> Can you please shed more context on this? cc: @joancaneus

I can revert the power units to be in watts as rounding down to integers by ods here will have negligible effects. Same can be done for current readings too.

However, for certain voltage sensors where the range of the reading is 0 to 1 volts, what we found when trying to reference ODS logs is that all reads were rounded down to 0.

E.g. The following sensor would always show up with a reading of 0 on ODS
```
          "name": "SMB_VRM_R3R0_ANLG0_VOUT_0V75",
          "sysfsPath": "/run/devmap/sensors/SMB_ISL68226_R3R0_ANLG0/in4_input",
          "type": 1,
          "thresholds": {
            "upperCriticalVal": 0.9,
            "lowerCriticalVal": 0.6
          },
          "compute": "@/1000.0"
        },
```

The proposed solution here is to get around the ODS limitation by reporting readings in milivolts.

somasun:
> for certain voltage sensors where the range of the reading is 0 to 1 volts,

How many such sensors are there?  If there are only a handful I propose we change the unit to millivolt only for those.  And reflect this in the sensor name with something like `_MILLIVOLT`

somasun (changes requested):
Putting it back in your queue to address the question.

aalamsi22:
> > for certain voltage sensors where the range of the reading is 0 to 1 volts,
> 
> How many such sensors are there? If there are only a handful I propose we change the unit to millivolt only for those. And reflect this in the sensor name with something like `_MILLIVOLT`

There aren't too many. The '_MILLIVOLT' idea should be ok. I will test this and update this PR with the changes

PR #720, opened 2025-12-04
Update/add SCM sensors and thresholds

pawagh1:
This update refreshes the SCM section of sensor_service.json so it matches the latest threshold guidance and exposes the additional telemetry that the platform now reports.

Key highlights of this update:

Threshold Definition Updates:
- Refreshed upperCriticalVal and lowerCriticalVal settings for all SCM voltages and temperatures.
- Introduced maxAlarmVal and minAlarmVal where the new guidance specifies warning bands, improving early alarm coverage.

Additional Sensors:
- Added SCM_Vp1P0_PCIE4, SCM_I_VP1P05_CPU, and SCM_I_VP1P8_VCCIN so the service can monitor the new power rails and currents delivered by the SCM regulators.

No other fields (slotPath, sysfsPath, pmUnitName, type) were modified; only the thresholds and the new sensor entries were updated.

Test Plan:
- Confirm the sensor_service loads the updated JSON without validation errors.
- Verify each SCM sensor reports values inside the refreshed threshold ranges on hardware.
- Spot-check that the newly added sensors emit readings and appear in downstream monitoring dashboards.


mikechoifb (changes requested):
Sensor name such as SCM_Vp1P0_PCIE4 should not contain lower-case letter. Please use uppercase.
After replacing SCM_Vp1P0_PCIE4 to SCM_VP1P0_PCIE4, now we see that this name is used twice (duplicate) - please rename the relevant entry correctly.

srikrishnagopu:

PR #834, opened 2026-01-24
[Nexthop] Adjust cmake targets to match BUCK build

benoit-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

The BUCK build has a single `fboss-error` target containing:
- `FbossError.h`
- `FbossHwUpdateError.h`
- `SysError.h`

The cmake build previously had two separate targets:
- `error` (only `FbossError.h`)
- `fboss_error` (`FbossError.h` and `SysError.h`)

This caused divergence between the two build systems.

Changes made:
1. Remove the `error` target from `cmake/Agent.cmake`
2. Update `fboss_error` to include all three headers to match BUCK:
   - `FbossError.h`
   - `FbossHwUpdateError.h` (was missing)
   - `SysError.h`
3. Update `fboss_error` dependencies to match BUCK exactly:
   - `fboss_cpp2` (matches `//fboss/agent/if:fboss-cpp2-types`)
   - `Folly::folly` (matches `//folly:conv` and `//folly/logging:logging`)
   - Remove `fboss_types` (not in BUCK dependencies)
   - Plot twist: add`switch_config_cpp2` dependency to `logging_util` target.
     This was previously provided transitively via `fboss_error` -> `fboss_types` -> `switch_config_cpp2`, but removing `fboss_types` from `fboss_error` broke that chain. The `logging_util` target directly includes `switch_config_types.h` so it needs to declare this dependency explicitly.
4. Replace all 23 instances of `error` dependency with `fboss_error` across 16 cmake files
5. Add missing `fboss_error` dependency to `external_phy` target (required in the BUCK file but was missing in cmake)

Additionally, two more fixes had to be fixed to address all the remaining issue with `system_scale_test_utils` found by the script, and those were resolved once again by matching BUCK files:
   - Remove bundled source files (`PortFlapHelper.cpp`, `MacLearningFloodHelper.cpp`)
   - Add dependencies on existing `port_flap_helper` and `mac_learning_flood_helper` targets (defined in `AgentTestUtils.cmake`)

Oh and a few more errors that sneaked into the code more recently and that I missed in #833 because I was working off a fork a few days old:
1. In `cmake/AgentState.cmake`, the `state` library was missing a dependency on `agent_features` (from Buck dep: `//fboss/agent:agent_features`)
2. In `cmake/fsdb/FsdbIf.cmake`, the `thriftpath_lib` library was missing a dependency on `fsdb_utils` (from Buck dep: `//fboss/fsdb/common:utils`)
3. In `cmake/fsdb/FsdbTests.cmake`, the `thriftpath_test_cpp2` library was missing a dependency on `common_cpp2` (from Buck dep: `//fboss/agent/if:common`)

Bonus: add duplicate cmake target detection to `buck_cmake_dep_checker.py`, because while fixing that last error I inadvertently made the mistake of creating a duplicate rule in a cmake file that was defined in another.

Also fix the tool to handle thrift sub-targets. Buck thrift_library generates sub-targets like `//path:name-cpp2-types`, `//path:name-cpp2-services`, etc. The tool now maps these to the corresponding cmake thrift target, which uncovered 10 additional pre-existing missing dependencies that are now fixed:
- Add `switch_config_cpp2` to `logging_util` (includes `switch_config_types.h`)
- Add `ctrl_cpp2` to `label_forwarding_action` (includes `ctrl_types.h`)
- Add `agent_hw_test_ctrl_cpp2` to `aqm_test_utils`
- Add `data_corral_service_cpp2`, `fan_service_cpp2`, `rackmon_cpp2` to `fboss2_lib`
- Add `fboss_common_cpp2` to `product_info`
- Add `fan_service_cpp2` to `cross_config_validator`
- Add `fan_service_cpp2` to `fan_service_config_validator`
- Add `transceiver_cpp2` to `qsfp_stats`

Now the script doesn't report any errors anymore and could potentially be used in CI to ensure no further differences between the BUCK build and the cmake build are introduced.

# Test Plan

FBOSS still builds successfully.

```
$ ./fboss/oss/scripts/buck_cmake_dep_checker.py
Parsing BUCK files...
  Found 730 Buck targets
Parsing cmake files...
  Found 652 cmake targets
Matching targets by source files...
  Found 510 matches
Building Buck-to-cmake dependency mapping...
Analyzing missing dependencies...

================================================================================
MISSING DEPENDENCIES
================================================================================

No missing dependencies found!
```

### DIFF
--- a/cmake/Agent.cmake
+++ b/cmake/Agent.cmake
@@ -148,7 +148,7 @@ add_library(utils
 target_link_libraries(utils
   agent_dir_util
   asic_utils
-  error
+  fboss_error
   ctrl_cpp2
   hw_switch_fb303_stats
   load_agent_config
@@ -260,7 +260,7 @@ add_library(hw_switch_thrift_client_table
 )
 
 target_link_libraries(hw_switch_thrift_client_table
-  error
+  fboss_error
   fboss_types
   hw_ctrl_cpp2
   Folly::folly
@@ -432,15 +432,6 @@ set(core_libs
 
 target_link_libraries(core ${core_libs})
 
-add_library(error
-  fboss/agent/FbossError.h
-)
-
-target_link_libraries(error
-  ctrl_cpp2
-  Folly::folly
-)
-
 add_library(thrifthandler_utils
   fboss/agent/ThriftHandlerUtils.cpp
 )
@@ -491,12 +482,12 @@ target_link_libraries(fboss_event_base
 
 add_library(fboss_error
   fboss/agent/FbossError.h
+  fboss/agent/FbossHwUpdateError.h
   fboss/agent/SysError.h
 )
 
 target_link_libraries(fboss_error
   fboss_cpp2
-  fboss_types
   Folly::folly
 )
 
@@ -514,7 +505,7 @@ target_link_libraries(platform_base
   agent_config_cpp2
   agent_dir_util
   ctrl_cpp2
-  error
+  fboss_error
   fboss_event_base
   fboss_types
   Folly::folly
--- a/cmake/AgentHw.cmake
+++ b/cmake/AgentHw.cmake
@@ -84,7 +84,7 @@ target_link_libraries(hw_switch_fb303_stats
 
 target_link_libraries(hw_fb303_stats
   counter_utils
-  error
+  fboss_error
   fb303::fb303
   Folly::folly
   switch_config_cpp2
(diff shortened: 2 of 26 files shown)


benoit-nexthop:
I found an issue with the removal of `fboss_types` from the dependencies of `fboss_error`: while this dependency does not exist in the BUCK build, some other cmake targets had become implicitly reliant on it, and they need to be fixed as a result of aligning the cmake build with the BUCK build.

benoit-nexthop:
I found another set of issues related to Thrift sub-targets that the Python script didn't handle. Fixing this now.

PR #893, opened 2026-02-03
[Nexthop][fboss2-dev] Rewrite fboss2-dev CLI end-to-end tests in C++ with gtest

benoit-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

This change enables running multiple CLI commands within a single process, which is required for C++ end-to-end tests. The first two end-to-end tests that were originally written in Python, along with helper functions, have been rewritten in C++, and can now run as a self-contained gtest binary, without depending on the fboss2-dev binary.

Some fixes were necessary to make this work, so the CLI could be re-used as a library instead of being invoked as a subprocess:

1. Remove the `hasRun` static variable from `CmdHandler`
   - Previously, a static `hasRun` flag prevented commands from executing more than once per process
   - Now we wrap callbacks in `CmdSubcommands` to check `get_subcommands().empty()` to detect leaf commands, eliminating the need for static state

2. Throw exceptions instead of calling `exit(1)` in `CmdHandler`
   - `CmdHandler::run()` now rethrows exceptions instead of calling `exit(1)`
   - This allows tests to catch errors and verify exit codes
   - `Main.cpp` catches exceptions and returns exit code 1

3. Factor out `CLI::App` initialization into `CliAppInit.h`
   - Header-only `initApp()` function shared between `Main.cpp` and tests
   - Eliminates code duplication for CLI setup
   - This couldn't be packaged into a library of its own without going into circular dependency hell, hence why it's a header-only helper

4. Reset state in `ConfigSession::initializeSession()`
   - Clear `ConfigSession` attributes when creating a new session (when session file doesn't exist)
   - Allows re-using the same `ConfigSession` singleton after `commit()`

5. Update state in `ConfigSession::commit()`
   - Set `base_` to the new commit SHA after successful commit
   - Set `configLoaded_ = false` to force config reload on next access
   - Allows re-using the same `ConfigSession` singleton after `commit()`

6. Automatically initialize session in `ConfigSession::loadConfig()`
   - If session file doesn't exist (e.g., after commit deleted it), call `initializeSession()` to create a fresh session

These changes allow the ConfigSession singleton to be properly reused across multiple CLI command invocations within the same process.

Note: this change is part of a series, the previous one is #881, the next one is #896.

# Test Plan

Added new end-to-end tests in C++ that replace `test_config_interface_mtu.py` and `test_config_interface_description.py` (the Python scripts will be removed in a future PR soon).

Sample output (simplified, without timestamps and other noise):

```
[==========] Running 2 tests from 2 test suites.
[----------] 1 test from ConfigInterfaceDescriptionTest
[ RUN      ] ConfigInterfaceDescriptionTest.SetAndVerifyDescription
[Step 1] Finding an interface to test...
  Using interface: eth1/1/1 (VLAN: 2001)
[Step 2] Getting current description...
  Current description: 'CLI_E2E_TEST_DESCRIPTION_ALT'
[Step 3] Setting description to 'CLI_E2E_TEST_DESCRIPTION'...
  Description set to 'CLI_E2E_TEST_DESCRIPTION'
[Step 4] Verifying description via 'show interface'...
  Verified: Description is 'CLI_E2E_TEST_DESCRIPTION'
[Step 5] Restoring original description ('CLI_E2E_TEST_DESCRIPTION_ALT')...
  Restored description to 'CLI_E2E_TEST_DESCRIPTION_ALT'
TEST PASSED
[       OK ] ConfigInterfaceDescriptionTest.SetAndVerifyDescription (412 ms)

[----------] 1 test from ConfigInterfaceMtuTest
[ RUN      ] ConfigInterfaceMtuTest.SetAndVerifyMtu
[Step 1] Finding an interface to test...
  Using interface: eth1/1/1 (VLAN: 2001)
[Step 2] Getting current MTU...
  Current MTU: 9000
[Step 3] Setting MTU to 1500...
  MTU set to 1500
[Step 4] Verifying MTU via 'show interface'...
  Verified: MTU is 1500
[Step 5] Verifying kernel interface MTU...
  Verified: Kernel interface fboss2001 has MTU 1500
[Step 6] Restoring original MTU (9000)...
  Restored MTU to 9000
TEST PASSED
[       OK ] ConfigInterfaceMtuTest.SetAndVerifyMtu (384 ms)

[==========] 2 tests from 2 test suites ran. (797 ms total)
[  PASSED  ] 2 tests.
```

### DIFF
--- a/fboss/cli/fboss2/CmdHandler.cpp
+++ b/fboss/cli/fboss2/CmdHandler.cpp
@@ -11,17 +11,37 @@
 #include "fboss/cli/fboss2/CmdHandler.h"
 #include <thrift/lib/cpp2/visitation/for_each.h>
 #include "fboss/cli/fboss2/CmdGlobalOptions.h"
+#include "fboss/cli/fboss2/CmdList.h"
+#include "fboss/cli/fboss2/gen-cpp2/cli_types.h"
+#include "fboss/cli/fboss2/utils/AggregateOp.h"
+#include "fboss/cli/fboss2/utils/AggregateUtils.h"
 #include "fboss/cli/fboss2/utils/CmdUtilsCommon.h"
 #include "thrift/lib/cpp/util/EnumUtils.h"
 #include "thrift/lib/cpp2/protocol/Serializer.h"
 
+#include <fmt/format.h>
+#include <folly/Conv.h>
+#include <folly/Demangle.h>
+#include <folly/ScopeGuard.h>
+#include <folly/Traits.h>
 #include <folly/executors/CPUThreadPoolExecutor.h>
 #include <folly/logging/xlog.h>
+#include <folly/stop_watch.h>
+#include <cstddef>
+#include <cstdint>
 #include <cstring>
+#include <exception>
 #include <future>
 #include <iostream>
+#include <map>
+#include <memory>
+#include <optional>
 #include <queue>
 #include <stdexcept>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
 
 using ThriftField = apache::thrift::metadata::ThriftField;
 using ThriftType = apache::thrift::metadata::ThriftType;
@@ -159,20 +179,10 @@ void printAggregate(
 
 namespace facebook::fboss {
 
-static bool hasRun = false;
-
 template <typename CmdTypeT, typename CmdTypeTraits>
 void CmdHandler<CmdTypeT, CmdTypeTraits>::runHelper() {
-  // Parsing library invokes every chained command handler, but we only need
-  // the 'leaf' command handler to be invoked. Thus, after the first (leaf)
-  // command handler is invoked, simply return.
-  // TODO: explore if the parsing library provides a better way to implement
-  // this.
-  if (hasRun) {
-    return;
-  }
-
-  hasRun = true;
+  // Note: The callback wrapper in CmdSubcommands.cpp ensures that only the
+  // leaf command handler is invoked. It checks if get_subcommands() is empty.
   auto extraOptionsEC =
       CmdGlobalOptions::getInstance()->validateNonFilterOptions();
   if (extraOptionsEC != cli::CliOptionResult::EOK) {
@@ -294,11 +304,19 @@ void CmdHandler<CmdTypeT, CmdTypeTraits>::runHelper() {
     }
   }
 
-  // Collect errors and display at end of execution
+  // Collect errors and display at end of execution, then throw if any failures
+  std::string combinedErrors;
   while (!executionFailures.empty()) {
     auto [host, errStr] = executionFailures.front();
     executionFailures.pop();
     XLOG(ERR) << host << " - Error in command execution: " << errStr;
+    if (!combinedErrors.empty()) {
+      combinedErrors += "; ";
+    }
+    combinedErrors += fmt::format("{}: {}", host, errStr);
+  }
+  if (!combinedErrors.empty()) {
+    throw std::runtime_error(combinedErrors);
   }
 }
 
@@ -333,7 +351,9 @@ void CmdHandler<CmdTypeT, CmdTypeTraits>::run() {
     runHelper();
   } catch (const std::exception& e) {
     std::cerr << e.what() << std::endl;
-    exit(1);
+    // Rethrow so the caller can handle appropriately (e.g., tests can catch and
+    // check exit codes, CLI main() can exit with non-zero status)
+    throw;
   }
 }
 
--- a/fboss/cli/fboss2/test/integration_test/CliTest.cpp
+++ b/fboss/cli/fboss2/test/integration_test/CliTest.cpp
@@ -1,0 +1,270 @@
+// (c) Meta Platforms, Inc. and affiliates. Confidential and proprietary.
+
+#include "fboss/cli/fboss2/test/integration_test/CliTest.h"
+
+#include <CLI/App.hpp>
+#include <CLI/Error.hpp>
+#include <fmt/format.h>
+#include <folly/String.h>
+#include <folly/Subprocess.h>
+#include <folly/init/Init.h>
+#include <folly/json/dynamic.h>
+#include <folly/json/json.h>
+#include <folly/logging/Init.h>
+#include <folly/logging/xlog.h>
+#include <gtest/gtest.h>
+#include <cstdlib>
+#include <exception>
+#include <filesystem>
+#include <iostream>
+#include <map>
+#include <sstream>
+#include <stdexcept>
+#include <streambuf>
+#include <string>
+#include <system_error>
+#include <vector>
+
+#include "fboss/cli/fboss2/CliAppInit.h"
+
+namespace fs = std::filesystem;
+
+namespace facebook::fboss {
+
+void CliTest::SetUp() {
+  XLOG(INFO) << "CliTest::SetUp - starting CLI test";
+  // Discard any stale session from previous runs to ensure we start fresh
+  discardSession();
+}
+
+void CliTest::TearDown() {
+  XLOG(INFO) << "CliTest::TearDown - cleaning up CLI test";
+}
+
+void CliTest::discardSession() const {
+  // Delete the session files to ensure we start with a fresh session
+  // based on the current HEAD. ConfigSession::initializeSession() will
+  // reset internal state when it detects no session file exists.
+  // NOLINTNEXTLINE(concurrency-mt-unsafe): HOME is read-only in practice
+  const char* home = std::getenv("HOME");
+  if (home == nullptr) {
+    XLOG(WARN) << "HOME environment variable not set, cannot discard session";
+    return;
+  }
+  fs::path sessionDir = fs::path(home) / ".fboss2";
+  fs::path sessionConfig = sessionDir / "agent.conf";
+  fs::path sessionMetadata = sessionDir / "cli_metadata.json";
+
+  std::error_code ec;
+  if (fs::exists(sessionConfig, ec)) {
+    XLOG(INFO) << "Discarding session config: " << sessionConfig.string();
+    fs::remove(sessionConfig, ec);
+    if (ec) {
+      XLOG(WARN) << "Failed to remove session config: " << ec.message();
+    }
+  }
+  if (fs::exists(sessionMetadata, ec)) {
+    XLOG(INFO) << "Discarding session metadata: " << sessionMetadata.string();
+    fs::remove(sessionMetadata, ec);
+    if (ec) {
+      XLOG(WARN) << "Failed to remove session metadata: " << ec.message();
+    }
+  }
+}
+
+CliTest::Result CliTest::executeCliCommand(
+    const std::vector<std::string>& args) const {
+  // Create a new CLI::App for this command
+  CLI::App app{"FBOSS CLI Test"};
+  utils::initApp(app);
+
+  // Build argv-style argument list
+  // Prepend program name and --fmt json
+  std::vector<std::string> fullArgs = {"cli_test", "--fmt", "json"};
+  fullArgs.insert(fullArgs.end(), args.begin(), args.end());
+
+  // Convert to argc/argv format
+  std::vector<char*> argv;
+  argv.reserve(fullArgs.size());
+  for (auto& arg : fullArgs) {
+    argv.push_back(arg.data());
+  }
+
+  // Redirect stdout and stderr to capture output
+  std::stringstream capturedStdout;
+  std::stringstream capturedStderr;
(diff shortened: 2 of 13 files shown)


fboss/cli/fboss2/session/ConfigSession.cpp
```diff
@@ -817,8 +831,12 @@ ConfigSession::CommitResult ConfigSession::commit(const HostInfo& hostInfo) {
         ec.message());
   }
 
-  // Reset action level after successful commit
+  // Reset internal state after successful commit so subsequent commands
+  // start with a fresh session based on the new commit.
   resetRequiredAction(cli::AgentType::WEDGE_AGENT);
+  base_ = commitSha;
+  // Force config reload from system config on next access
+  configLoaded_ = false;
```
joseph5wu, line 935: Do we also need to clear commands_ and requiredActions_ like you did for line 639-642 above?
If so, can you make a private resetSession() function to make sure all the necessary variables get reset

fboss/cli/fboss2/CmdHandler.cpp
```diff
@@ -9,17 +9,38 @@
...
 #include <iostream>
+#include <map>
+#include <memory>
+#include <optional>
 #include <queue>
 #include <stdexcept>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
```
joseph5wu, line 44: Why suddenly includes so many libraries? Your change in the implementation below doesn't seem to rely on all these libraries

fboss/cli/test/CliTest.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+#ifdef IS_OSS
+FOLLY_INIT_LOGGING_CONFIG("DBG2; default:async=true");
+#else
+FOLLY_INIT_LOGGING_CONFIG("fboss=DBG2; default:async=true");
+#endif
+
+int main(int argc, char* argv[]) {
+  return facebook::fboss::cliTestMain(argc, argv);
+}
```
joseph5wu: Why not just put the implementation of cliTestMain() in the main()?
I didn't see `cliTestMain()` is used anywhere else?

joseph5wu (commented):
Overall it matches what I would expect an end-to-end CLI test binary should look like.
Although this is the first diff, I can see there're some future improvements on how to write these tests in a more concise way. As most of the tests should follow the below steps:
- pick the subject for your tests. could be an interface, a vlan or a qos
- prepare the exact cli you wants to run and then execute
- call a validation function
- maybe we should also properly check the config version updated or config session infra logic in the test too
- restore the config

Hopefully with more and more tests you're about to add, you can consider what's a best way to make CliTest.cpp more structured and all the commands can reuse more of the code.

What's more, we should also introduce "batch cli" in your end-to-end tests; or cli with different action levels.

fboss/cli/test/CliTest.h
```diff
@@ -0,0 +1,133 @@
...
+struct CliResult {
+  int exitCode;
+  std::string stdout;
+  std::string stderr;
+};
+
+/**
+ * Represents a network interface from the output of 'show interface'.
+ */
+struct Interface {
```
joseph5wu: Let's put this inside CliTest class so it won't mess up with the real facebook::fboss::Interface
(https://github.com/facebook/fboss/blob/main/fboss/agent/state/Interface.h#L56)

fboss/cli/fboss2/test/integration_test/CliTest.cpp
```diff
@@ -0,0 +1,274 @@
...
+  // reset internal state when it detects no session file exists.
+  // NOLINTNEXTLINE(concurrency-mt-unsafe): HOME is read-only in practice
+  const char* home = std::getenv("HOME");
+  if (home == nullptr) {
+    XLOG(WARN) << "HOME environment variable not set, cannot discard session";
+    return;
+  }
+  fs::path sessionDir = fs::path(home) / ".fboss2";
+  fs::path sessionConfig = sessionDir / "agent.conf";
+  fs::path sessionMetadata = sessionDir / "cli_metadata.json";
```
joseph5wu, line 56: Since we're touching files/folders for the fboss2 config cli, I think we should consider to introduce 
something similar to 
https://github.com/facebook/fboss/blob/main/fboss/agent/AgentDirectoryUtil.h
for our config related directory to store these static settings

Feel free to use future PR to address this improvement

fboss/cli/fboss2/test/integration_test/CliTest.h
```diff
@@ -0,0 +1,133 @@
...
+   */
+  void commitConfig() const;
+
+  /**
+   * Get the MTU of a kernel interface using 'ip -json link'.
+   * @param vlanId The VLAN ID (interface name will be fboss<vlanId>)
+   * @return MTU value, or -1 if interface not found
+   */
+  int getKernelInterfaceMtu(int vlanId) const;
+
```
joseph5wu, line 110: Since this function is only needed for the ConfigInterfaceMtuTest.cpp.
we don't have to define this method in the general CliTest.cpp.
We should set the standard that this CliTest.cpp only handles the common utility functions,
any other functions related to specific tests should be handled in their own test classes.

fboss/cli/test/CliTest.cpp
joseph5wu: Let's move the cliTestMain() implementation to the main() directly.
I don't see there's other place needed to call cliTestMain()

fboss/cli/fboss2/test/integration_test/CliTest.cpp
```diff
@@ -0,0 +1,274 @@
...
+        fmt::format(
+            "CLI command failed with exit code {}: {}",
+            result.exitCode,
+            result.stderr));
+  }
+  if (result.stdout.empty()) {
+    return folly::dynamic::object();
+  }
+  return folly::parseJson(result.stdout);
+}
```
joseph5wu, line 130: I am not a big fan of this function, because all the other run cli or command functions are return `CliResult` while this returns folly::dynamic and throw exception if there's an error. 

I'd recommend to remove such function, and always let users call `runCli()`. And then you just need to provide a function like `folly::dynamic toFollyDynamic(const std::string& stdout)`. This will force your test to do EXPECT_EQ(result.exitCode, 0) first before calling this function to convert the string

fboss/cli/fboss2/test/integration_test/CliTest.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+int CliTest::getKernelInterfaceMtu(int vlanId) const {
+  auto kernelIntf = fmt::format("fboss{}", vlanId);
+  // Use full path to ip command since folly::Subprocess doesn't search PATH
+  auto result = runCmd({"/usr/sbin/ip", "-json", "link", "show", kernelIntf});
+
+  if (result.exitCode != 0) {
+    XLOG(WARN) << "Kernel interface " << kernelIntf << " not found";
+    return -1;
+  }
```
joseph5wu, line 243: All FBOSS Interface should be shown in kernel interface.
Let's treat it as an error if such interface doesn't exist

fboss/oss/scripts/run_scripts/run_test.py
```diff
@@ -1291,116 +1291,115 @@ def _filter_tests(self, tests: List[str]) -> List[str]:
...
-        test_scripts = []
-        for filename in sorted(os.listdir(test_dir)):
-            if filename.startswith("test_") and filename.endswith(".py"):
-                test_scripts.append(os.path.join(test_dir, filename))
+        CLI tests are simpler than hardware tests - they don't need config file
+        manipulation, warmboot/coldboot setup, or SAI-specific logging. They just
+        run against the already-running agent via Thrift.
+        """
+        tests_to_run = self._get_tests_to_run()
+        tests_to_run = self._filter_tests(tests_to_run)
```
joseph5wu: I'm okay with this diff doesn't properly orchestra all the necessary services to run cli_test.
But ideally, I would like to see run_test.py to properly setup agent, qsfp, fsdb and etc services just like link tests.

fboss/oss/scripts/run_scripts/run_test.py
```diff
@@ -1291,116 +1291,115 @@ def _filter_tests(self, tests: List[str]) -> List[str]:
...
+            print(f"Running all tests took {delta_time}")
+
+            if result.returncode != 0:
+                sys.exit(result.returncode)
+
+        except subprocess.TimeoutExpired:
+            print(f"[  TIMEOUT ] CLI tests timed out after {args.test_run_timeout}s")
+            sys.exit(1)
+        except Exception as e:
+            print(f"[  ERROR   ] Failed to run CLI tests: {e}")
```
joseph5wu: Now as cli_test as a cpp test binary just like all the other test binaries, can we keep the run logic identical

joseph5wu (changes requested):

fboss/cli/fboss2/CmdHandler.cpp
benoit-nexthop: This is adding already required headers. I generate the includes with `clang-include-cleaner`.

fboss/cli/fboss2/session/ConfigSession.cpp
benoit-nexthop: This change wasn't part of this PR. It was part of the PR adding rebase: #832

fboss/cli/fboss2/test/integration_test/CliTest.h
benoit-nexthop: I put it here because I felt like it was likely to be needed for other tests in the future.

fboss/cli/test/CliTest.cpp
benoit-nexthop: Done.

fboss/cli/fboss2/test/integration_test/CliTest.cpp
benoit-nexthop: Will do.

fboss/cli/test/CliTest.h
benoit-nexthop: Done. Also move the `CliResult` class that was right before it, so that it became `CliTest::Result`.

fboss/cli/fboss2/test/integration_test/CliTest.cpp
benoit-nexthop: Done.

fboss/cli/fboss2/test/integration_test/CliTest.cpp
benoit-nexthop: I don't want to have to repeat an `EXPECT_EQ()` followed by a call to `toFollyDynamic()` everywhere we run a CLI command. If your concern is with the exception throwing here, can we instead do this:
```c++
  EXPECT_EQ(result.exitCode, 0) << "CLI command failed: " << result.stderr;

  if (result.exitCode != 0 || result.stdout.empty()) {
    return folly::dynamic::object();
  }
```

fboss/oss/scripts/run_scripts/run_test.py
benoit-nexthop: Done. I had to override the `fruid_path` and `coldboot` args for `CliTestRunner` in order to disable passing `--fruid_path` and `--setup-for-warmboot` as those flags are not supported by the `cli_test` binary.

fboss/oss/scripts/run_scripts/run_test.py
benoit-nexthop: OK, I can do this in a follow up change. Should `CliTestRunner` mimic what `LinkTestRunner` does in `_setup_coldboot_test()` and `_setup_warmboot_test()` then? If yes maybe it would be better to refactor the code so as to avoid duplication between `LinkTestRunner` and `CliTestRunner`. I can send a PR for this if you agree.

fboss/cli/fboss2/test/integration_test/ConfigInterfaceMtuTest.cpp
```diff
@@ -0,0 +1,83 @@
...
+  int kernelMtu = getKernelInterfaceMtu(*interface.vlan);
+  if (kernelMtu > 0) {
+    EXPECT_EQ(kernelMtu, newMtu)
+        << "Kernel MTU is " << kernelMtu << ", expected " << newMtu;
+    XLOG(INFO) << "  Verified: Kernel interface fboss" << *interface.vlan
+               << " has MTU " << kernelMtu;
+  } else {
+    XLOG(INFO) << "  Skipped: Kernel interface fboss" << *interface.vlan
+               << " not found";
+  }
```
benoit-nexthop, line 75: This part of the change was not merged.

PR #888, opened 2026-01-30
[Nexthop] FBOSS Distro init service and default configs

marif-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

# Summary
At boot, start a oneshot `systemd` service that launches a `fboss_init.sh` script to perform FBOSS distro initialization and setup e.g. copying right config files to `/etc/coop` etc.

- Added `systemd` service launches the script `fboss_init.sh`
  - The init script does the following steps for now
     - Copy `/etc/coop` configuration files based on the `dmidecode` output
     - Generate `fruid.json` file
- Added option to add default configs (`qsfp_service` and `wedge_agent`) for different platforms
- Added default configs for `montblanc` platform as an example

# Test Plan

### fboss_init systemd service
```
● fboss_init.service - FBOSS Initialization
     Active: active (exited) since Tue 2026-01-27 12:00:20 PST
     
Jan 27 16:05:34 fboss101 fboss_init[90731]: [fboss_init] Starting FBOSS initialization
Jan 27 16:05:34 fboss101 fboss_init[90732]: [fboss_init] Detected platform: minipack3
Jan 27 16:05:34 fboss101 fboss_init[90732]: [fboss_init] Using platform config directory: /opt/fboss/share/minipack3
Jan 27 16:05:34 fboss101 fboss_init[90731]: [fboss_init] Copied agent.conf: /opt/fboss/share/minipack3/agent.conf -> /etc/coop/agent.conf
Jan 27 16:05:34 fboss101 fboss_init[90731]: [fboss_init] Copied qsfp.conf: /opt/fboss/share/minipack3/qsfp.conf -> /etc/coop/qsfp.conf
Jan 27 16:05:35 fboss101 fboss_init[90731]: [fboss_init] Generated fruid.json: /var/facebook/fboss/fruid.json
Jan 27 16:05:35 fboss101 fboss_init[90731]: [fboss_init] FBOSS initialization complete
```
### when system config files exist (skipped)
```
Jan 27 16:00:03 fboss101 fboss_init[90579]: [fboss_init] Starting FBOSS initialization
Jan 27 16:00:03 fboss101 fboss_init[90580]: [fboss_init] Detected platform: minipack3
[...]

PR #936, opened 2026-02-14
[Nexthop] Use split sw/hw agents in Distro 

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

The mono-mode `wedge_agent` is going to be deprecated 'soon' and it makes more sense to start Distro Image off using the multi-mode split `fboss_sw_agent` and multiple `fboss_hw_agent`s, one per ASIC.

This change implements that. Switching to `fboss_sw_agent` is straightforward.

Enabling a variable number of `fboss_hw_agent`s is slightly trickier because we need some way to determine the correct number of HW agents to start. We need this to be statically configured so a misbehaving ASIC doesn't go undetected.

The way that is solved here is by adding a `num_hw_agents` file to the configurations extracted by fboss_init. As its name implies, it is the number of hardware agents to start on the given platform. The `fboss_init.sh` script will read this file and enable that number of
hardware agents.

To make the systemd unit dependencies works correctly, all the possible hardware agents are groups under an `fboss_hw_agents.target` target. Here up to 4 HW agents are supported, but that is easily extended.

# Test Plan

Load the image on a Minipack3. Then verify that:

1. `fboss_sw_agent` is running; `ps -A|grep fboss_sw`
2. `fboss_hw_agent@0` is enabled: `systemctl status fboss_hw_agent@0.service`
3. `fboss_hw_agent@1` is disabled: `systemctl status fboss_hw_agent@1.service`


travisb-nexthop:
This builds upon https://github.com/facebook/fboss/pull/888

travisb-nexthop:
Note that the fboss2-dev config CLI does not yet support split agents. We may want to delay merging this until that support is finished.

fboss-image/image_builder/templates/centos-09.0/root_files/usr/lib/systemd/system/fboss_hw_agent@.service
```diff
@@ -0,0 +1,34 @@
...
+Type=simple
+LimitNOFILE=10000000
+LimitCORE=32G
+MemoryMax=3.75G
+MemorySwapMax=0
+
+Environment="PATH=/opt/fboss/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin"
+Environment="LD_LIBRARY_PATH=/opt/fboss/lib"
+WorkingDirectory=/opt/fboss
+ExecStart=/bin/bash -c 'source /opt/fboss/bin/setup_fboss_env && exec /opt/fboss/bin/fboss_hw_agent-sai_impl --switchIndex %i'
```
kevin645, line 21: How do we decide on mutli selection of sai sdk's are we going to stick with the assumption there is only 1 sai_impl on the target distro device?

travisb-nexthop: Right now Distro Image is designed with the assumption that any particular image will only be built with one ASIC SAI. That is the logic implemented here.

When multi-SAI SDK support is added, the logic here would be extended to use whatever the new choice mechanism is.

travisb-nexthop:
This PR is stacked on top of https://github.com/facebook/fboss/pull/888

kevin645:
needs rebasing!

kevin645:
fboss_sw_agent is running; ps -A|grep fboss_sw
fboss_hw_agent@0 is enabled: systemctl status fboss_hw_agent@0.service
fboss_hw_agent@1 is disabled: systemctl status fboss_hw_agent@1.service

have you been able to debug this any further? We wouldn't want to ship this without hw agent running properly.

travisb-nexthop:
> fboss_sw_agent is running; ps -A|grep fboss_sw fboss_hw_agent@0 is enabled: systemctl status [fboss_hw_agent@0.service](mailto:fboss_hw_agent@0.service) fboss_hw_agent@1 is disabled: systemctl status [fboss_hw_agent@1.service](mailto:fboss_hw_agent@1.service)
> 
> have you been able to debug this any further? We wouldn't want to ship this without hw agent running properly.

Debug what? The test was run on a single-ASIC system so we only want one HW agent to be enabled. The other HW agents we want to be disabled because there is no accompanying ASIC.

So fboss_hw_agent@0 enabled, fboss_hw_agent@1 disable is expected.

kevin645:
> > fboss_sw_agent is running; ps -A|grep fboss_sw fboss_hw_agent@0 is enabled: systemctl status [fboss_hw_agent@0.service](mailto:fboss_hw_agent@0.service) fboss_hw_agent@1 is disabled: systemctl status [fboss_hw_agent@1.service](mailto:fboss_hw_agent@1.service)
> > have you been able to debug this any further? We wouldn't want to ship this without hw agent running properly.
> 
> Debug what? The test was run on a single-ASIC system so we only want one HW agent to be enabled. The other HW agents we want to be disabled because there is no accompanying ASIC.
> 
> So fboss_hw_agent@0 enabled, fboss_hw_agent@1 disable is expected.

sorry the main issue was the test plan label asic startup errors was my main concern.

travisb-nexthop:
> > > fboss_sw_agent is running; ps -A|grep fboss_sw fboss_hw_agent@0 is enabled: systemctl status [fboss_hw_agent@0.service](mailto:fboss_hw_agent@0.service) fboss_hw_agent@1 is disabled: systemctl status [fboss_hw_agent@1.service](mailto:fboss_hw_agent@1.service)
> > > have you been able to debug this any further? We wouldn't want to ship this without hw agent running properly.
> > 
> > 
> > Debug what? The test was run on a single-ASIC system so we only want one HW agent to be enabled. The other HW agents we want to be disabled because there is no accompanying ASIC.
> > So fboss_hw_agent@0 enabled, fboss_hw_agent@1 disable is expected.
> 
> sorry the main issue was the test plan label asic startup errors was my main concern.

Ah. That's been resolved elsewhere. I've removed that line from test plan as it is no longer valid. The fboss_hw_agent@0 starts fine and stays running in our internal build.

travisb-nexthop:
https://github.com/facebook/fboss/pull/1000 is the PR which adds split-agent support to the CLI. This PR should merge after that one to keep the image usable.

PR #950, opened 2026-02-20
[Nexthop] Create FBOSS command wrappers in PATH 

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

There exist a handful of FBOSS commands which we would like to have in
$PATH, for example fboss2.

We can't simply put /opt/fboss/bin into $PATH for two reasons:

1. /opt/fboss/bin also contains the daemon binaries which we won't want cluttering up tab completion or being executed by accident
2. When the Platform Stack or Forwarding Stack components are updated without rebuilding the entire image, they will not be installed into /opt/fboss, but instead somewhere else. It's too difficult to update the $PATH of all existing shells when this happens.

Instead we introduce wrappers in /usr/local/bin for each of these commands which will later find the latest installation of the matching component and executes that copy of the command. For now they only execute from the default install path /opt/fboss/bin.

# Test Plan

Manual testing:
```
[root@fboss103 ~]# which fboss2-dev
/usr/local/bin/fboss2-dev
[root@fboss103 ~]# fboss2-dev --help
FBOSS CLI
Usage: /opt/fboss/bin/fboss2-dev [OPTIONS] SUBCOMMAND

Options:
  -h,--help                   Print this help message and exit
  -H,--host TEXT ...          Hostname(s) to query
  -l,--loglevel TEXT          Debug log level
  --file TEXT:FILE            filename, queries every host in the file
  --fmt TEXT:{tabular,json}   Output format (tabular, json)
  --agent-port INT:POSITIVE   Agent thrift port to connect to
  --qsfp-port INT:POSITIVE    QsfpService thrift port to connect to
  --color TEXT                color (no, yes => yes for tty and no for pipe)
  --filter TEXT               filter expression. expression must be of the form TERM1&&TERM2||TERM3||TERM4... where each TERMi is of the form <key op value> See specific commands for list of available filters. Please note the specific whitespace requirements for the command
  --aggregate TEXT            Aggregation operation. Must be of the form OPERATOR(column)
  --aggregate-hosts BOOLEAN   whether to perform aggregation across all hosts or not

Subcommands:
  bounce                      Disable/Enable object
  clear                       Clear object info
  config                      Configuration commands
  create                      Create object
  debug                       Debug object
  delete                      Delete object
  get                         Get object
  reload                      Reload object
  set                         Set object
  show                        Show object info
  start                       Start event
  stop                        Stop event
  stream                      Continuously stream
```


travisb-nexthop:
This comes after https://github.com/facebook/fboss/pull/903

kevin645:
can we rebase this one

kevin645:
Is this scenario typically using distro cli to update at runtime?

travisb-nexthop:
> Is this scenario typically using distro cli to update at runtime?

This is preparing for that scenario, yes. Actually locating the newest Platform Stack/Forwarding Stack to find the new CLI binaries installed via `fboss-image device update` is not in this PR. It will be a follow-on to the PRs which implement the `fboss-image device update` command.

PR #987, opened 2026-03-07
[Nexthop] Distro Image install all components

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

This was discovered missing as part of the Omnibus PR. Originally this would have been spread across a handful of PRs, one for each component type.

Here the various components are installed into the final image.

# Test Plan

Proven out in the Omnibus PR.


fboss/oss/scripts/fake-sai-devel.tar.zstd
travisb-nexthop, line 1: Watch out that this binary file merges. It is critical to from_source.json.

PR #995, opened 2026-03-12
[Nexthop] Add bootstrap Distro Image FSDB config

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

This default FSDB configuration is required to support FSDB starting as a service. Without it, FSDB will crash on startup.

# Test Plan

Load a Distro Image with this fix and verify that fsdb stays running for a couple of minutes.


fboss-image/image_builder/templates/centos-09.0/root_files/etc/coop/fsdb/current
```diff
@@ -0,0 +1,67 @@
...
+      "paths": [
+        {
+          "path": { "raw": ["build_info"] },
+          "isExpected": false,
+          "isStats": false
+        }
+      ],
+      "skipThriftStreamLivenessCheck": true
+    },
+    "qsfp_service": {
```
elangovan911-meta, line 34: Sensor service is available in open source as well. It is also one of the publishers for FSDB. so this can be added to the config

"sensor_service": {
      "paths": [
        {
          "isExpected": false,
          "isStats": true,
          "path": {
            "raw": [
              "sensor_service"
            ]
          }
        }
      ],
      "skipThriftStreamLivenessCheck": false
    }
  },

elangovan911-meta (approved):
Approving since the comment is NIT and FSDB should be able to launch either way/ But please make the change before merging

fboss-image/image_builder/templates/centos-09.0/root_files/etc/coop/fsdb/current
travisb-nexthop: I've added that sensor_service snippet.

I do not have permissions to import and merge.

PR #999, opened 2026-03-12
[Nexthop] Fixup Distro Image service dependencies

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

Avoid service startup succeeding through retries.

fboss_init uses weutil to produce fruid.json. weutil first requires that platform_manager is up so it can access the EEPROMs via the devmap.

Thus here we fix the dependency order and make platform_manager a hard dependency of fboss_init.

qsfp_service gets a similar treatment with fboss_init so it had fruid.json available at the first start.

# Test Plan

See that the services start up reliably when booting Distro Image. Without this change, the FBOSS services and agents would occasionally fail to load when the retry timing did not line up.

### DIFF
--- a/fboss-image/image_builder/templates/centos-09.0/root_files/usr/lib/systemd/system/qsfp_service.service
+++ b/fboss-image/image_builder/templates/centos-09.0/root_files/usr/lib/systemd/system/qsfp_service.service
@@ -3,6 +3,8 @@
 Description=FBOSS QSFP Service
 After=platform_manager.service
 After=rc-local.service
+Wants=fboss_init.service
+After=fboss_init.service
 
 [Service]
 Type=simple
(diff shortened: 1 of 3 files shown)


fboss-image/image_builder/templates/centos-09.0/root_files/usr/lib/systemd/system/qsfp_service.service
```diff
@@ -3,6 +3,8 @@
 Description=FBOSS QSFP Service
 After=platform_manager.service
 After=rc-local.service
+Wants=fboss_init.service
```
kevin645, line 6: We definitely should also make sure fboss_init.service is for all forwarding services no ?

travisb-nexthop: The ones which need to read from hardware or fruid.json, yes. The split agent PR contains the necessary dependencies for the non-mono mode https://github.com/facebook/fboss/pull/936/changes#diff-0808dccebe7981510f4d34354832dc0d2f116cd16b23213b8d227aed16a14ffaR9. This PR explicitly didn't touch it to reduce merge conflicts.

PR #1000, opened 2026-03-12
[Nexthop] Add experimental Bazel build with auto-generated BUILD files

benoit-nexthop:
# Summary

Add an experimental Bazel build for FBOSS that auto-generates BUILD files from the existing Meta-internal BUCK files. This eliminates the manual cmake divergence problem: generated files are never committed, so they always reflect the current state of the BUCK files.

### What's included

- **`fboss/oss/scripts/bazelify.py`** — Converter that translates BUCK rules to Bazel equivalents (`cpp_library` → `cc_library`, `thrift_library` → `fboss_thrift_library`, etc.). Resolves internal/external deps, handles OSS file replacements, auto-discovers headers, and supports BUCK annotations (`# bazelify: exclude`, `# bazelify: include foo.bzl:func`) to control generation. Tracks input checksums so `--if-needed` skips regeneration when nothing changed.

- **`fboss/build_defs/getdeps_dep.bzl`** — Starlark repository rule that builds third-party deps on demand via `cached-get-deps.py`. Deps are built lazily during Bazel's loading phase — no separate getdeps step required.

- **`fboss/oss/scripts/cached-get-deps.py`** — Per-dependency build/cache wrapper. Computes a content hash of each dep's manifest + revision + cmake defines, checks an optional HTTP cache for a pre-built tarball, and builds from source on cache miss. Gives per-dependency caching: unchanged deps download in seconds on fresh CI VMs.

- **`fboss/build_defs/thrift_library.bzl`** — `fboss_thrift_library` macro that runs `thrift1 --gen mstch_cpp2` via a genrule and compiles the generated C++ with `cc_library`.

- **`.bazelrc`** — Configures clang from `/usr/local/llvm/bin/`, sets C++20 standard, carries over compiler flags from `CMakeLists.txt`. Sandboxing is disabled (`--spawn_strategy=local`) because FBOSS BUCK files don't declare all headers explicitly.

- **`fboss/oss/scripts/bazel.sh`** — Wrapper that sources site-specific config (`fboss-build.env`), regenerates BUILD files if needed, sets up sccache if configured, assembles `.bazel.d/` config fragments, and execs Bazel.

- **`fboss/oss/scripts/fboss-build.env.example`** — Template for site-specific infrastructure (sccache scheduler, getdeps HTTP cache, Bazel remote cache). Without this file, the build works normally — deps are built from source and compilation runs locally.

- **Hand-written `.bzl` files** for SAI and wedge_agent targets where internal Meta macros have no OSS equivalent (`sai_api_impl.bzl`, `sai_switch_impl.bzl`, `sai_tracer_impl.bzl`, `sai_diag_impl.bzl`, `sai_platform_impl.bzl`).

- **`docs/docs/build/Build_with_Bazel.md`** — Full documentation covering usage, site configuration, architecture, design decisions, converter internals, and rationale for the repository rule approach vs. alternatives.

- **Minimal BUCK file annotations** (`# bazelify: exclude`) for targets that don't build in the OSS context (Meta-internal deps, facebook/ subdirectories, etc.).

### Test coverage

- `fboss/oss/scripts/test_bazelify.py` — 55 tests covering BUCK parsing, dep resolution, BUILD file generation, thrift translation, and `_make_build_file_content` matrix
- `fboss/oss/scripts/test_cached_get_deps.py` — 31 tests covering cache key generation, HTTP operations, and symlink helpers

# Test Plan

- [x] `bazel build //fboss/...` — **978 targets** build successfully
- [x] `bazel test //fboss/...` — **43 tests** pass
- [x] `pytest test_bazelify.py` — **55 tests** pass
- [x] `pytest test_cached_get_deps.py` — **31 tests** pass
- [x] Bazel-built `wedge_agent` binary tested on **Minipack 3** hardware
- [x] Bazel-built `wedge_agent` binary tested on **Wedge 800** hardware
- [x] Bazel-built `wedge_agent` binary tested on **NH-4010** hardware
- [x] Verified no site-specific infrastructure references in committed code
- [ ] More testing is required — additional platforms and longer soak testing planned. No issues found so far.


fboss/lib/platforms/BUCK
```diff
@@ -9,6 +9,7 @@ cpp_library(
     ],
 )
 
+# bazelify: extra_dep = @gflags//:gflags
```
Scott8440, line 12: Why do we need to have a bazel annotation for gflags when it's already in the "exported_external_deps" field here?

docs/docs/build/Build_with_Bazel.md
```diff
@@ -0,0 +1,573 @@
...
+    - OSS
+    - build
+    - bazel
+oncall: fboss_oss
+---
+
+FBOSS has an experimental Bazel build that auto-generates BUILD files from the
+existing Meta-internal BUCK files. This keeps the OSS build in sync with the
+internal build mechanically, avoiding the manual divergence problems that
+plagued the previous cmake-based OSS build.
```
Scott8440, line 17: I fear that this won't actually avoid the manual divergence problem, it just moves it to a different system. Currently the divergence happens when you forget to update a CMake file, but now there is still an opportunity for divergence if you forget to add a `# bazelify ...` annotation.

The counterpoint is of course that for a lot of targets there is no annotation needed and these targets will stay automatically in sync forever unless something changes. So I agree there is a benefit here, but I'm not fully convinced it's worth supporting an entirely custom, complicated build system.

docs/docs/build/Build_with_Bazel.md
```diff
@@ -0,0 +1,573 @@
...
+The Bazel build is a work in progress. Known limitations:
+
+- **Sandboxing is disabled.** FBOSS source files frequently include headers
+  that are not declared in the BUILD files. Bazel's sandboxing is turned off
+  via `--spawn_strategy=local` in `.bazelrc` until all headers are properly
+  declared.
+
+- **Some targets fail due to Meta-internal dependencies.** Targets that
+  transitively depend on code under `facebook/` directories or on internal
+  services (configerator, scribe, etc.) may fail to compile.
```
Scott8440, line 195: Can you give example targets that fail? I would expect that any target that currently builds in OSS with Cmake today should work with bazel, as no code under facebook/ directories get out anyways.

docs/docs/build/Build_with_Bazel.md
```diff
@@ -0,0 +1,573 @@
...
+- Header-only `cc_library` for header-only packages (e.g., range-v3, CLI11)
+- Special-cased targets for googletest, folly, boost (see `_make_build_file_content`)
+
+Inter-dependencies between pre-built libraries are declared statically (e.g.,
+folly depends on fmt, glog, gflags, boost).
+
+#### 2. BUCK Parsing
+
+The parser uses regex-based extraction (not a full Starlark parser) to find
+target definitions in BUCK files. It handles:
```
Scott8440, line 309: What is the reasoning behind this decision? If using a full parser is an option at all I think it would be nice to have that, as doing something as complex as this with regexes feels dangerous.

docs/docs/build/Build_with_Bazel.md
```diff
@@ -0,0 +1,573 @@
...
+
+**Inline** (placed at the end of a source file line inside a target):
+
+```python
+    "Foo.cpp",  # bazelify: exclude  -- Exclude this file from the target
+```
+
+Use `bazelify.py --generate-excludes` to systematically annotate BUCK files
+by comparing against the cmake build. It identifies files, targets, and
+entire BUCK files that are not part of the OSS build.
```
Scott8440, line 484: Is it expected that this functionality will be used continuously moving forward (until cmake is removed entirely)? The issue I see here is similar to the manual divergence comment I raised above:

* This tool makes it easy to keep the bazel build system up-to-date, but only if the Cmake files are (manually) updated 
* When we remove cmake entirely this tool no longer works and we now have to manually update the exclude annotations

fboss/agent/thrift_packet_stream/tests/BUCK
```diff
@@ -27,6 +27,7 @@ cpp_unittest(
     srcs = [
         "BidirectionalPacketStreamTest.cpp",
     ],
+    size = "large",
```
Scott8440, line 30: Why?

fboss/agent/hw/sai/api/sai_api_impl.bzl
```diff
@@ -0,0 +1,79 @@
+"""Hand-written SAI API library targets for OSS Bazel build.
```
Scott8440, line 1: Let's discuss what would be needed to not have to hand-write these files. If it's just bazel macros I bet we can get those open sourced.

Scott8440 (commented):
As we discussed, we don't think it's feasible to introduce this change to fboss as of now. The main reason is that we would need to get every engineer on board with a new way of making builds work in open source. Given that we have already dedicated a lot of work into making Cmake better, introducing this new change could take away from our ability to polish what we're working on.

Additionally, while getdeps has its problems, we know it works and it has been work-hardened over years. This new system is very complex and still incomplete.

I've left some more detailed concerns in the rest of the review. We do think that there is a possibility that we can bring this change in future if Nexthop is willing to keep working on this in the meantime. In general we would be more willing to accept this if there was a no-effort way get the OSS build in sync with internal (e.g. no manual annotation of buck files), and if the system itself gets simplified.

fboss/lib/platforms/BUCK
benoit-nexthop: This wasn't actually needed here, you're right, because `gflags` is assumed to always be installed (whether we build it with getdeps or install it otherwise). The `extra_dep` annotation was intended to force it in case `self._is_available_repo()` return `False` in `bazelify.py` but that cannot happen for those dependencies so I updated the code and I removed the annotation.

docs/docs/build/Build_with_Bazel.md
benoit-nexthop: While it's true that the divergence problem isn't completely eliminated, it's significantly reigned in. The cmake files are very frequently out of sync, with new files and dependencies added to the BUCK build (or sometimes the cmake build) and not on the other side. The Bazelification of the BUCK build pretty much eliminates these problems and makes the maintenance of the remaining deltas much easier to maintain in the long term (assuming we can ditch the cmake build altogether).

Also I disagree with the characterization as a "entirely custom, complicated build system", while the script that does the translation is not as simple as I had wished, the result is a clean, pure Bazel build, that closely mimics the Buck build.

docs/docs/build/Build_with_Bazel.md
benoit-nexthop: What I meant here is that there are many targets referenced in the BUCK files that cannot be built with Bazel because they depend on Meta-internal code that is not OSS. Those targets were never in the cmake build, and are being excluded from the Bazel build with annotations.

docs/docs/build/Build_with_Bazel.md
benoit-nexthop: It felt simpler initially, but it's true that as I've been adding things to it, maybe we should revisit this and see how the code would compare if we used the [official parser](https://github.com/bazelbuild/starlark) code.

docs/docs/build/Build_with_Bazel.md
benoit-nexthop: This is intended to help maintain the annotations during the transition period, in case people update the cmake build without updating the annotations. It's not something we'd keep once we ditch the cmake build. I still contend that if we ditch the cmake build, the resulting maintenance overhead of keeping the bazelify annotations up to date will be significantly smaller than keeping a whole separate parallel build in sync with the BUCK build.

fboss/agent/thrift_packet_stream/tests/BUCK
benoit-nexthop: This test is slow and regularly takes more than 300s to run. Bazel is stricter than BUCK in a few ways, but it doesn't hurt the Buck build to specify things a bit more fully.

fboss/agent/hw/sai/api/sai_api_impl.bzl
benoit-nexthop: Yes, open sourcing more of your `.bzl` files would help a lot.

benoit-nexthop:
> As we discussed, we don't think it's feasible to introduce this change to fboss as of now. The main reason is that we would need to get every engineer on board with a new way of making builds work in open source. Given that we have already dedicated a lot of work into making Cmake better, introducing this new change could take away from our ability to polish what we're working on.

Yes, there is a transition cost that is not free, but ultimately it will be easier to maintain once cmake is out of the picture. There is a lot less cognitive load to think about "Buck + annotations" than think about "Buck + cmake". It's all in one place, you don't need to update totally separate files that have a very different structure etc. The current status quo already shows that mistakes are regularly made as new files and dependencies are frequently added to the BUCK files and not to the cmake files.

> Additionally, while getdeps has its problems, we know it works and it has been work-hardened over years. This new system is very complex and still incomplete.

This change doesn't get rid of getdeps, it ties into it, as the Bazel build still calls to getdeps to build all the third party dependencies. It's just that getdeps is no longer needed beyond that.

> I've left some more detailed concerns in the rest of the review. We do think that there is a possibility that we can bring this change in future if Nexthop is willing to keep working on this in the meantime. In general we would be more willing to accept this if there was a no-effort way get the OSS build in sync with internal (e.g. no manual annotation of buck files), and if the system itself gets simplified.

I will keep working on this as we're going to migrate to Bazel internally and ditch cmake no matter what, because building with Bazel is so much better for us than continuing to use cmake. You guys don't feel the pain of building with cmake because you use Buck.

Having no manual annotations would be hard because the OSS build is inherently different than the Meta-internal build, so something is needed to record what those differences are (e.g. "add this file or dependency for the OSS build").

benoit-nexthop:
As per our discussion today, there is not enough interest in upstreaming this for now, so I'm closing this PR. We can reopen it and revisit later if needed.

PR #977, opened 2026-03-04
[Nexthop] Distro CLI image command 

travisb-nexthop:
**Pre-submission checklist**
- [X] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [X] `pre-commit run`

# Summary

The fboss-image device image command configures the Distro
Infrastructure to PXE boot the given image file for the given device
on the next reboot.

This replaces the manual steps previously necessary with the MVP
Distro Infrastructure.

# Test Plan

```
$ ./fboss-image device dc:da:4d:fc:ad:2d image nh-fboss-broadcom-xgs.tar
[0.00s] Setting image for device dc:da:4d:fc:ad:2d: nh-fboss-broadcom-xgs.tar
[0.07s] Using persistent directory: /home/travisb/work/upstream/fboss/fboss-image/distro_infra/data
[0.07s] Extracting image tarball to /home/travisb/work/upstream/fboss/fboss-image/distro_infra/data/dc-da-4d-fc-ad-2d...
[1.96s] Image extracted successfully to /home/travisb/work/upstream/fboss/fboss-image/distro_infra/data/dc-da-4d-fc-ad-2d
[1.96s] Enabling PXE boot for MAC address: dc-da-4d-fc-ad-2d
[2.09s] Successfully configured device dc:da:4d:fc:ad:2d with image nh-fboss-broadcom-xgs.tar
[2.09s] Device is ready for PXE boot
```

Then watch the switch PXE boot on the next boot:
```
>>Checking Media Presence......
>>Media Present......
>>Start PXE over IPv6 on MAC: DC-DA-4D-FC-AD-2D. Press ESC key to abort PXE boot..
  Station IP address is FC00:33:0:0:0:FB05:5030:774C
  Server IP address is FC00:33:0:0:0:0:0:1000
  NBP filename is ipxev6.efi
  NBP filesize is 1158144 Bytes

>>Checking Media Presence......
>>Media Present......
 Downloading NBP file...

  NBP file downloaded successfully.
iPXE initialising devices...
```

PR #1018, opened 2026-03-18
[Nexthop] Add Distro CLI fboss-image device image-upstream command

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

This is a simple front-end for the image-upstream command which will
combine the various image description options into an URL and then
download it, checking that previously downloaded images are not
re-downloaded if the HTTP server indicates that they have not changed.

Later work will formalize the image publishing backend, but this
provides most of the necessary front-end.

# Test Plan

New unit tests. Also a manual happy-path test that when the cache does
not contain the image it is downloaded and immediately repeating the
command does not redownload the file.


travisb-nexthop:
This is stacked on top of https://github.com/facebook/fboss/pull/977/

kevin645:
rebase please!

PR #1019, opened 2026-03-18
[Nexthop] Add support for build with qsfp_service_sai and hw_agent_sai

ctikku-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

Add support to build binaries with another SAI library (QSFP) as well as existing SAI library (for ASIC from Broadcom). This commit was done in two parts,  "sai" in manifest to "hw_agent_sai" and add support for another tag, qsfp_service_sai.

# Test Plan

Tested fboss image build with existing SAI libraries works without any issues but since we do not have access to any QSFP SAI library, that part could not be validated.


Lewis-Kang:
May I know why the term ‘PAI’ is not used directly? Thanks.

ctikku-nexthop:
> May I know why the term ‘PAI’ is not used directly? Thanks.

Lewis, Just got an update from Travis that there was a discussion (between Meta and Nexthop) where Joseph suggested naming these as follows:

    hw_agent_sai -> npu_sai, 
    qsfp_service_sai -> phy_sai

I will update the code and PR shortly.

PR #1054, opened 2026-03-31
[Nexthop][fboss2-dev] Restart spilt agent for warmboot and coldboot

hillol-nexthop:
# Summary

The fboss2-dev config session commit command needs to correctly restart services after a config change — including in split-agent deployments where fboss_sw_agent and fboss_hw_agent@N run as separate systemd units instead of the monolithic wedge_agent. The old implementation had service restart logic inlined in ConfigSession, couldn't handle split-mode, and was untestable due to direct folly::Subprocess calls.

- Extracted service restart and reload logic into a new FbossServiceUtil class, with a SystemdInterface abstraction over systemd operations for dependency injection
- FbossServiceUtil reads switchInfoMap + multi_switch flag from agent config to determine whether to restart wedge_agent (monolithic) or fboss_sw_agent + fboss_hw_agent@N (split mode)
- Coldboot properly uses AgentDirectoryUtil to find the correct cold_boot_once file per service (sw/hw/monolithic)
- ConfigSession now lazily constructs FbossServiceUtil on first commit (via ensureFbossServiceUtil()), or accepts an injected mock for testing
- applyServiceActions now returns the list of actual systemd unit names that were acted on, so commit output can report e.g. "warmboot: fboss_sw_agent, fboss_hw_agent@0"
- Added unit tests (ConfigSessionSystemdTest) with a mock SystemdInterface and integration tests (ConfigSessionRestartTest) covering warmboot and coldboot end-to-end on real hardware
# Test Plan
This test suite (`ConfigSessionSystemdTest`) contains 9 tests that verify the split agent and monolith logic in `ConfigSession`:

- **IsSplitMode_ReturnsTrueWhenSwAgentEnabled** - Verifies `isSplitMode()` returns `true` when `fboss_sw_agent` service is enabled
- **IsSplitMode_ReturnsFalseWhenSwAgentNotEnabled** - Verifies `isSplitMode()` returns `false` when `fboss_sw_agent` service is not enabled
- **IsSplitMode_ReturnsFalseOnException** - Verifies `isSplitMode()` gracefully handles exceptions and returns `false` (assumes monolithic mode)

- **RestartService_MonolithicMode_Warmboot** - Verifies warmboot restart calls `restartService("wedge_agent")` and waits for service to become active
- **RestartService_MonolithicMode_Coldboot** - Verifies coldboot restart calls `stopService()` then `startService()` (not `restartService()`) for `wedge_agent`

- **RestartService_SplitMode_Warmboot_SingleHwAgent** - Verifies warmboot restart restarts both `fboss_sw_agent` and `fboss_hw_agent@0`
- **RestartService_SplitMode_Coldboot_SingleHwAgent** - Verifies coldboot restart stops then starts both agents in correct sequence

- **RestartService_PropagatesFailure** - Verifies exceptions from systemd operations are propagated to caller
- **RestartService_ServiceFailsToStart** - Verifies exceptions from `waitForServiceActive()` timeout are propagated to caller

**Total: 9 tests** covering split mode detection, service restart logic for both monolithic and split architectures, and error handling.

fboss2_integration_test:
- ConfigSessionRestartTest
The above test would cover both warmboot and coldboot


joseph5wu:
@hillol-nexthop Can you update your summary here cause the sentences look a bit confusing to me:
> For hitless, it will reload fboss_sw_agent.
For coldboot, it will stop and start fboss_sw_agent and fboss_hw_agent@* for warmboot, it will restart fboss_sw_agent and fboss_hw_agent@* For monolith it will be backward compatible.

In our production, we prefer to restart all agent services (sw+N*hw) regardless of warmboot and coldboot.
- For Coldboot, please make sure you write the corresponding flag files before you restart these services
sw_agent: https://github.com/facebook/fboss/blob/main/fboss/agent/AgentDirectoryUtil.h#L116
hw_agent: https://github.com/facebook/fboss/blob/main/fboss/agent/AgentDirectoryUtil.h#L118
- For Warmboot, just restart these services 
What's more, we recommend to start/stop hw_agent before sw_agent.

Btw, can we have a new utility class to handle the service restart logic? I would like to see ConfigSession only handle the config related logic, and have something FbossServiceUtil to handle whether it's split mode, restart the related service logic will be a cleaner implementation.

fboss/cli/fboss2/session/FbossServiceUtil.cpp
```diff
@@ -0,0 +1,225 @@
...
+    std::unique_ptr<SystemdInterface> systemd)
+    : systemd_(std::move(systemd)),
+      switchInfoMap_(std::move(switchInfoMap)),
+      multiSwitch_(multiSwitch) {}
+
+std::string FbossServiceUtil::getServiceName(cli::ServiceType service) {
+  switch (service) {
+    case cli::ServiceType::AGENT:
+      return "wedge_agent";
+  }
```
joseph5wu, line 51: Wouldn't this need to consider the split mode services' names?

fboss/cli/fboss2/session/FbossServiceUtil.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+bool FbossServiceUtil::isSplitMode() const {
+  return multiSwitch_;
+}
+
+std::string FbossServiceUtil::getColdbootFileForService(
+    const std::string& service) {
+  AgentDirectoryUtil dirUtil;
+
+  if (service == "fboss_sw_agent") {
```
joseph5wu: Can we make all these services names as a constexpr at the beginning of this cpp file?

fboss/cli/fboss2/session/FbossServiceUtil.cpp
```diff
@@ -0,0 +1,225 @@
...
+    return dirUtil.getHwColdBootOnceFile(switchIndex);
+  } else if (service == "wedge_agent") {
+    return dirUtil.getColdBootOnceFile();
+  } else {
+    throw std::runtime_error(
+        fmt::format("Unknown service type for coldboot: {}", service));
+  }
+}
+
+void FbossServiceUtil::createColdbootMarkerFile(
```
joseph5wu, line 77: Please consider to reuse the existing common file utility functions from
https://github.com/facebook/fboss/blob/main/fboss/lib/CommonFileUtils.h#L52

Can you also check whether there's any other file management logic in fboss2/config and
please adjust them accordingly

fboss/cli/fboss2/session/FbossServiceUtil.cpp
```diff
@@ -0,0 +1,225 @@
...
+    performRestartAndWait(service);
+    LOG(INFO) << "Warmboot completed for service: " << service;
+  }
+}
+
+std::vector<std::string> FbossServiceUtil::getServicesToRestart(
+    cli::ServiceType service) const {
+  std::vector<std::string> services;
+
+  if (isSplitMode()) {
```
joseph5wu: I know we only support AGENT service for now, but let's also make sure you only need to check split mode when the service type is actually AGENT. Otherwise in the future when we support BGP, you will always return sw_agent/hw_agent even though the cli might only touch BGP config here

joseph5wu (commented):
Overall these new changes look better to me. Please address the comments before sending out as a ready to ship PR.
And please fix the failures on the github action for this PR

fboss/cli/fboss2/session/FbossServiceUtil.cpp
hillol-nexthop: In case of monolith, there is just wedge_agent but in case of split, there are two agents. So which name to return in case of split? Because of this confusion I have not touched the logic of this function. Instead wherever agent name is required, I handled it there itself.
For e.x. Which agent to restart:
Monolith - getServiceName() -> wedge_agent
Split - fboss_sw + N*fboss_hw

fboss/cli/fboss2/session/FbossServiceUtil.cpp
hillol-nexthop: Thanks it helped a lot reduce the code volume.

fboss/cli/fboss2/session/FbossServiceUtil.cpp
hillol-nexthop: done

fboss/cli/fboss2/session/FbossServiceUtil.cpp
hillol-nexthop: Done

fboss/cli/fboss2/session/FbossServiceUtil.cpp
hillol-nexthop: Yes there are multiple other places in ConfigSession and BgpConfigSession where raw file management is done. They can be optimised by leveraging CommonFileUtils. But I would prefer to do them as a separate PR for a better separation of concern. Hope that should be fine.

PR #1055, opened 2026-03-31
[Nexthop][fboss2-dev] Add test to cover default constructor of ConfigSession (#526)

hillol-nexthop:
# Summary
There was no way to test the default constructor of ConfigSession in UT. The change is made to extend the coverge.
Implemented environment variable override (`FBOSS_CONFIG_BASE_DIR`) for `AgentDirectoryUtil` to enable testing of `ConfigSession` default constructor.

### 1. AgentDirectoryUtil.cpp

Added a helper function `getBaseConfigDir()` that checks for the `FBOSS_CONFIG_BASE_DIR` environment variable:

### 2. CmdConfigSessionTest.cpp

Added test `defaultConstructorWithEnvOverride` that:
- Sets `FBOSS_CONFIG_BASE_DIR` to a test directory
- Calls `ConfigSession::getInstance()` which uses the default constructor
- Verifies the session is created with the test paths
- Confirms the default constructor is now fully testable

## Benefits

✅ **Backward Compatible**: Production code unchanged - defaults to `/etc/coop` if env var not set
✅ **Test-Friendly**: Tests can now use the default constructor by setting `FBOSS_CONFIG_BASE_DIR`
✅ **No API Changes**: Existing code continues to work without modifications
✅ **Follows Existing Patterns**: Similar to `FLAGS_volatile_state_dir` and `FLAGS_persistent_state_dir`

## Production Behavior

In production, `FBOSS_CONFIG_BASE_DIR` is not set, so the default `/etc/coop` path is used. No changes to production deployment or configuration required.

**Pre-submission checklist**
- [ ] `pre-commit run`

# Test Plan
The change is primarily to extend testing.

### DIFF
--- a/fboss/cli/fboss2/session/ConfigSession.cpp
+++ b/fboss/cli/fboss2/session/ConfigSession.cpp
@@ -126,6 +126,19 @@ std::string getHomeDirectory() {
   return std::string(home);
 }
 
+// Helper function to get base config directory with environment variable
+// override. This allows tests to redirect config paths to temporary directories
+// by setting FBOSS_CONFIG_BASE_DIR environment variable. If not set, defaults
+// to "/etc/coop"
+std::string getBaseConfigDir() {
+  // NOLINTNEXTLINE(concurrency-mt-unsafe): Used only during initialization
+  const char* envOverride = std::getenv("FBOSS_CONFIG_BASE_DIR");
+  if (envOverride != nullptr && !std::string(envOverride).empty()) {
+    return {envOverride};
+  }
+  return "/etc/coop";
+}
+
 void ensureDirectoryExists(const std::string& dirPath) {
   std::error_code ec;
   fs::create_directories(dirPath, ec);
@@ -284,10 +297,19 @@ ConfigSession::ConfigSession() {
   username_ = getUsername();
   std::string homeDir = getHomeDirectory();
 
-  // Use AgentDirectoryUtil to get the config directory path
-  // getConfigDirectory() returns /etc/coop/agent, so we get the parent to get
-  // /etc/coop
-  AgentDirectoryUtil dirUtil;
+  // Create AgentDirectoryUtil with environment-aware config directory
+  // This allows tests to override config paths via FBOSS_CONFIG_BASE_DIR
+  std::string baseConfigDir = getBaseConfigDir();
+  AgentDirectoryUtil dirUtil(
+      FLAGS_volatile_state_dir,
+      FLAGS_persistent_state_dir,
+      kPkgDir,
+      kSystemdDir,
+      getBaseConfigDir() + "/agent", // configDirectory_
+      getBaseConfigDir() + "/agent_drain"); // drainConfigDirectory_
+
+  // getConfigDirectory() returns /etc/coop/agent (or override), get parent to
+  // get /etc/coop
   std::string coopDir =
       fs::path(dirUtil.getConfigDirectory()).parent_path().string();
 
--- a/fboss/agent/AgentDirectoryUtil.cpp
+++ b/fboss/agent/AgentDirectoryUtil.cpp
@@ -54,13 +54,26 @@ DEFINE_string(
 using namespace std;
 
 namespace {
-static auto constexpr kPkgDir =
-    "/etc/packages/neteng-fboss-wedge_agent/current";
-static auto constexpr kSystemdDir = "/etc/systemd/system";
-static auto constexpr kCoopAgentDir = "/etc/coop/agent";
-static auto constexpr kCoopAgentDrainDir = "/etc/coop/agent_drain";
-static auto constexpr kSwAgentService = "fboss_sw_agent.service";
-static auto constexpr kHwAgentService = "fboss_hw_agent@.service";
+auto constexpr kPkgDir = "/etc/packages/neteng-fboss-wedge_agent/current";
+auto constexpr kSystemdDir = "/etc/systemd/system";
+auto constexpr kCoopAgentDir = "/etc/coop/agent";
+auto constexpr kCoopAgentDrainDir = "/etc/coop/agent_drain";
+auto constexpr kSwAgentService = "fboss_sw_agent.service";
+auto constexpr kHwAgentService = "fboss_hw_agent@.service";
+
+// Helper function to get base config directory with environment variable
+// override This allows tests to redirect config paths to temporary directories
+// by setting FBOSS_CONFIG_BASE_DIR environment variable. If not set, defaults
+// to "/etc/coop"
+std::string getBaseConfigDir() {
+  // NOLINTNEXTLINE(concurrency-mt-unsafe): Used only during initialization
+  const char* envOverride = std::getenv("FBOSS_CONFIG_BASE_DIR");
+  if (envOverride != nullptr && !std::string(envOverride).empty()) {
+    return {envOverride};
+  }
+  return "/etc/coop";
+}
+
 } // namespace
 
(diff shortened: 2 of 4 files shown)


joseph5wu:
Actually do you think instead of asking fboss/agent/AgentDirectoryUtil.cpp to check the system environment, can we actually use the non-default AgentConfigUtil constructor in ConfigSession?
https://github.com/facebook/fboss/blob/main/fboss/agent/AgentDirectoryUtil.h#L34

That's the reason that we have different constructors in this util library. I think if we are changing the default behavior with this extra getenv() logic, it kinda misses the point of having different constructors, right?

hillol-nexthop:
> Actually do you think instead of asking fboss/agent/AgentDirectoryUtil.cpp to check the system environment, can we actually use the non-default AgentConfigUtil constructor in ConfigSession? https://github.com/facebook/fboss/blob/main/fboss/agent/AgentDirectoryUtil.h#L34
> 
> That's the reason that we have different constructors in this util library. I think if we are changing the default behavior with this extra getenv() logic, it kinda misses the point of having different constructors, right?

Updated

fboss/cli/fboss2/session/ConfigSession.cpp
```diff
@@ -284,10 +297,19 @@ ConfigSession::ConfigSession() {
...
+  AgentDirectoryUtil dirUtil(
+      FLAGS_volatile_state_dir,
+      FLAGS_persistent_state_dir,
+      kPkgDir,
+      kSystemdDir,
+      getBaseConfigDir() + "/agent", // configDirectory_
+      getBaseConfigDir() + "/agent_drain"); // drainConfigDirectory_
+
+  // getConfigDirectory() returns /etc/coop/agent (or override), get parent to
+  // get /etc/coop
```
joseph5wu, line 312: @hillol-nexthop The second ConfigSession::ConfigSession() with sessionConfigDir/systemConfigDir is used for TestableConfigSession, which is used by the unit tests
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/test/TestableConfigSession.h#L27

And the default `ConfigSession::ConfigSession()` will be used by the regular fboss2-dev logic?
When you changed the code here instead of the other constructor but your test can still pass, it makes me think we don't even need such change?

joseph5wu (changes requested):
Most of our existing config unit tests are using `TestableConfigSession`, which is not using the default `ConfigSession::ConfigSession()` which uses the default /etc/coop/ config directory. Instead `TestableConfigSession` is using `ConfigSession::ConfigSession()` with its specified `sessionConfigDir` and `systemConfigDir` without using  `AgentDirectoryUtil`.
Do we still need this change if current existing unit tests are already not using the default /etc/coop/agent.conf?

fboss/cli/fboss2/session/ConfigSession.cpp
hillol-nexthop: True. The default constructor is used in the regular fboss2-dev and TestableConfigSession is used in UT. Because of this, the default constructor remains uncovered in the UT. Some times back there was a breakage in the fboss2-dev because of a bug in the default constructor flow (don't remember the exact breakage though!) and then I thought it would be good to have some coverage of the default constructor. And this change is just to make the default constructor testable.
See if it makes sense.

joseph5wu: I'm afraid I still didn't see the value of making the change here on the default ConfigSession::ConfigSession() by changing the coop config directory to the one you prefer in UT, which is exactly the second constructor ConfigSession::ConfigSession() is created for.
If you want to make sure both constructors have high test coverage, why not just do something like:
```
  ConfigSession::ConfigSession()
      : ConfigSession(
            getHomeDirectory() + "/.fboss2",
            fs::path(AgentDirectoryUtil().getConfigDirectory())
                .parent_path()
                .string()) {}

  ConfigSession::ConfigSession(
      std::string sessionConfigDir,
      std::string systemConfigDir)
      : sessionConfigDir_(std::move(sessionConfigDir)),
        systemConfigDir_(std::move(systemConfigDir)),
        username_(getUsername()),
        git_(std::make_unique<Git>(systemConfigDir_)) {
    initializeSession();
  }
```

So you will know besides the `sessionConfigDir_` and `systemConfigDir_` are different b.w regular use of fboss2-dev and UT, everything else will have the same implementation

hillol-nexthop: The coop dir will depend on the env var- so it remains the same in prod and will change in test.
Your approach looks clean as you are delegating the job to parameterised constructor from the default one and the parameterised constructor is already tested in UT.
But my motivation was to touch the default constructor flow where AgentDirectoryUtil instantiates and creates the path for coop dir because this is where I found issue some times back. You are condensing the existing logic to a single line here but this flow will remain untested.

benoit-nexthop: @joseph5wu we can totally have the 2-arg ctor call the default ctor, but it doesn't change anything to the fact that the default ctor will remain untested.

joseph5wu: @hillol-nexthop / @benoit-nexthop : Sure if we're chasing 100% test coverage here, then we can add the test via env variables. But this is never gonna be possible to be 100% test coverage in reality.
So now we have to answer the question, what do we really want to test here?
Do you want to test the default `AgentDirectoryUtil` behavior in this default constructor? But the way you did here also changes the default `AgentDirectoryUtil` behavior, and you'll never actually verify the default `AgentDirectoryUtil` behavior.
Or if you want to verify the two "customized" directories are actually expected. The we can have an easier unit tests for these "default" directories gen loigc.
```
// sessionConfigDir_
getHomeDirectory() + "/.fboss2"

// systemConfigDir_
fs::path(AgentDirectoryUtil().getConfigDirectory())
                .parent_path()
                .string()
```

Once again, I'm not against adding more tests but I just want to make sure we're not over-engineering here to adjust the current logics just to support unnecessary unit tests .

Please if you can, can you share what was the problem of using the default constructors back then?

hillol-nexthop: I landed up on a change in the dafault constructor because of which it didn't call the AgentDirUtil to get the  coop dir path. It was caught in manual testing and then I realised we can cover this in UT too and so added the tests.
It's okay to ignore this change if it is going towards over-engineering direction. And anyway these set of default constructors don't go through a lot of churn so, better close the PR.

joseph5wu: @hillol-nexthop / @benoit-nexthop : please decide whether you guys want to close this PR(I think the UT is not fundamentally adding any strong test coverage). But I do think you can consider to introduce a smaller change:
```
ConfigSession::ConfigSession()
      : ConfigSession(
            getHomeDirectory() + "/.fboss2",
            fs::path(AgentDirectoryUtil().getConfigDirectory())
                .parent_path()
                .string()) {}

  ConfigSession::ConfigSession(
      std::string sessionConfigDir,
      std::string systemConfigDir)
      : sessionConfigDir_(std::move(sessionConfigDir)),
        systemConfigDir_(std::move(systemConfigDir)),
        username_(getUsername()),
        git_(std::make_unique<Git>(systemConfigDir_)) {
    initializeSession();
  }
```
So that you can keep the `ConfigSession` constructor behavior similar b/w test and prod scenarios

hillol-nexthop: I will chose to close the PR.

PR #1072, opened 2026-04-10
[Nexthop][fboss2-dev] Add interface profile cli command

hillol-nexthop:
# Summary
Implements `fboss2-dev config interface <port-list> profile <profile-id>` — the command specified in NOS-4544. Profile-id is a `PortProfileID` enum string (e.g. `PROFILE_100G_4_NRZ_RS528_COPPER`, case-insensitive). Setting a profile also updates `port->speed()` to the corresponding speed derived from PlatformMapping, keeping both fields consistent.

This PR also reverts #635 (which implemented `speed` under NOS-4544 instead of `profile`). The `speed` attribute has been removed from the CLI for now; it will be re-added in a follow-up ticket if needed.

**New command syntax:**
```
fboss2-dev config interface eth1/1/1 profile PROFILE_100G_4_NRZ_RS528_COPPER
fboss2-dev config interface eth1/1/1 profile profile_100g_4_nrz_rs528_copper   # case-insensitive
fboss2-dev config interface eth1/1/1,eth1/2/1 profile PROFILE_400G_8_PAM4_RS544X2N_OPTICAL
fboss2-dev config interface eth1/1/1 profile PROFILE_DEFAULT   # resets to default
```

## Implementation

Added `ProfileValidator` class (`ProfileValidation.h/.cpp`) mirroring the `SpeedValidator` pattern:
- **Constructor** queries `FbossCtrl::getPlatformMapping` + `QsfpService::getAllPortSupportedProfiles` once
- **Validation strategy**: QSFP-primary (optics-aware), PlatformMapping fallback (hardware capabilities)
- **`parseProfile()`**: case-insensitive via `TEnumTraits<cfg::PortProfileID>::findValue(toUpper(str))`
- **`validateProfile()`**: checks requested profile is in supported list; `PROFILE_DEFAULT` always passes
- **`getProfileSpeed()`** called once before the per-port loop (speed is a property of the profile, not the port — avoids rebuilding `PlatformMapping` per port)

Both `port->profileID()` and `port->speed()` are set on apply, keeping them consistent.

  
# Test Plan

**Unit tests** in `CmdConfigInterfaceTest.cpp` (21 new test cases):
- `parseProfileValidUppercase` / `parseProfileValidLowercase` — case-insensitive parsing
- `parseProfileDefault` — PROFILE_DEFAULT accepted
- `parseProfileUnknown` / `parseProfileEmpty` — invalid strings rejected
- `validateProfileSupported` — supported profile validated and returned
- `validateProfileDefault` — PROFILE_DEFAULT bypasses hardware check
- `validateProfileNotSupported` — unsupported profile rejected with helpful error
- `validateProfilePortNotFound` — unknown port throws runtime_error
- `validateProfileQsfpPrimaryFallback` — QSFP primary, PlatformMapping fallback verified
- `getProfileSpeedKnown` / `getProfileSpeedUnknown` — speed lookup from PlatformMapping
- `queryClientSetsProfile` — full command sets both profileID and speed
- `queryClientSetsProfileDefault` — PROFILE_DEFAULT resets both fields
- `queryClientProfileUnsupportedThrows` — unsupported profile rejected
- `queryClientProfileInvalidStringThrows` — bad string rejected before Thrift calls
- `queryClientSetProfileMultiPort` — all ports updated in multi-port command
- `queryClientSetProfileAndDescription` — combined attributes work correctly
- `queryClientNoAttributesErrorMentionsProfile` — error message lists profile

**Integration tests** added in `ConfigInterfaceProfileTest.cpp`:
- `SetValidProfileSuccess`, `SetInvalidProfileFails`, `SetProfileDefaultResetsSpeed`, `SetProfileMultiInterface`

**Pre-submission checklist**
- [x] `pre-commit run`


joseph5wu:
@hillol-nexthop : FBOSS has been using this `PortProfileID` concept to program our port/interface for a long time
https://github.com/facebook/fboss/blob/main/fboss/agent/switch_config.thrift#L129
Basically this port profile idea is to capture multiple port speed and phy configs into one single profile so we can program such interface with the specified port profile all at once.
Therefore, I highly recommend to continue using the term of profiles to program port speed instead of the single speed here. Otherwise, fboss2 won't be able to tell which exactly profile to use based on the speed, like when you want to use 200G, you won't be able to tell it's `PROFILE_200G_4_PAM4_RS544X2N_COPPER` or `PROFILE_200G_4_PAM4_RS544X2N_OPTICAL`. 


Please consider to support the following cli instead
```
fboss2-dev config interface eth1/1/1 profile PROFILE_200G_4_PAM4_RS544X2N_COPPER
```
The cli can still use `getAllPortSupportedProfiles()` to check whether the user input profile is valid or not.

```
> fboss2 -H rsw029.p066.f01.nha6 show interface eth1/8/5 capabilities
 Interface  Profiles
-------------------------------------------------------------------
 eth1/8/5   PROFILE_100G_4_NRZ_RS528_COPPER
            PROFILE_100G_4_NRZ_RS528_OPTICAL
            PROFILE_200G_4_PAM4_RS544X2N_COPPER
            PROFILE_200G_4_PAM4_RS544X2N_OPTICAL
            PROFILE_400G_4_PAM4_RS544X2N_OPTICAL
            PROFILE_400G_4_PAM4_RS544X2N_COPPER


> fboss2 -H rsw029.p066.f01.nha6 show interface eth1/8/5
+-----------+--------+-------+------+------+------------------------------+-------------+
| Interface | Status | Speed | VLAN | MTU  | Addresses                    | Description |
-----------------------------------------------------------------------------------------
| eth1/8/5  | down   | 200G  | 3002 | 9000 | fe80::d849:bfff:fe9c:eae9/64 |             |
-----------------------------------------------------------------------------------------

> fboss2 -H rsw029.p066.f01.nha6 show port eth1/8/5
 ID  Name      AdminState  LinkState  ActiveState  Transceiver  TcvrID  Speed  ProfileID                             HwLogicalPortId  Drained  PeerSwitchDrained  PeerPortDrainedOrDown  Errors  Core Id  Virtual device Id  Cable Len meters
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 42  eth1/8/5  Enabled     Down       --           Present      7       200G   PROFILE_200G_4_PAM4_RS544X2N_OPTICAL  36               No       --                 --                     --      --       --                 --
```

hillol-nexthop:
> @hillol-nexthop : FBOSS has been using this `PortProfileID` concept to program our port/interface for a long time https://github.com/facebook/fboss/blob/main/fboss/agent/switch_config.thrift#L129 Basically this port profile idea is to capture multiple port speed and phy configs into one single profile so we can program such interface with the specified port profile all at once. Therefore, I highly recommend to continue using the term of profiles to program port speed instead of the single speed here. Otherwise, fboss2 won't be able to tell which exactly profile to use based on the speed, like when you want to use 200G, you won't be able to tell it's `PROFILE_200G_4_PAM4_RS544X2N_COPPER` or `PROFILE_200G_4_PAM4_RS544X2N_OPTICAL`.
> 
> Please consider to support the following cli instead
> 
> ```
> fboss2-dev config interface eth1/1/1 profile PROFILE_200G_4_PAM4_RS544X2N_COPPER
> ```
> 
> The cli can still use `getAllPortSupportedProfiles()` to check whether the user input profile is valid or not.
> 
> ```
> > fboss2 -H rsw029.p066.f01.nha6 show interface eth1/8/5 capabilities
>  Interface  Profiles
> -------------------------------------------------------------------
>  eth1/8/5   PROFILE_100G_4_NRZ_RS528_COPPER
>             PROFILE_100G_4_NRZ_RS528_OPTICAL
>             PROFILE_200G_4_PAM4_RS544X2N_COPPER
>             PROFILE_200G_4_PAM4_RS544X2N_OPTICAL
>             PROFILE_400G_4_PAM4_RS544X2N_OPTICAL
>             PROFILE_400G_4_PAM4_RS544X2N_COPPER
> 
> 
> > fboss2 -H rsw029.p066.f01.nha6 show interface eth1/8/5
> +-----------+--------+-------+------+------+------------------------------+-------------+
> | Interface | Status | Speed | VLAN | MTU  | Addresses                    | Description |
> -----------------------------------------------------------------------------------------
> | eth1/8/5  | down   | 200G  | 3002 | 9000 | fe80::d849:bfff:fe9c:eae9/64 |             |
> -----------------------------------------------------------------------------------------
> 
> > fboss2 -H rsw029.p066.f01.nha6 show port eth1/8/5
>  ID  Name      AdminState  LinkState  ActiveState  Transceiver  TcvrID  Speed  ProfileID                             HwLogicalPortId  Drained  PeerSwitchDrained  PeerPortDrainedOrDown  Errors  Core Id  Virtual device Id  Cable Len meters
> ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>  42  eth1/8/5  Enabled     Down       --           Present      7       200G   PROFILE_200G_4_PAM4_RS544X2N_OPTICAL  36               No       --                 --                     --      --       --                 --
> ```

Okay I will implement 
config interface <> profile <profile-id> 
instead of speed.
In that case do we need to query qsfp? I think if we just query agent for supportedProfiles, we can see valid/invalid.

fboss/cli/fboss2/commands/config/interface/CmdConfigInterface.cpp
```diff
@@ -113,8 +115,55 @@ InterfacesConfig::InterfacesConfig(std::vector<std::string> v)
...
 }
 
+std::string applyProfile(
+    const HostInfo& hostInfo,
+    const utils::InterfaceList& interfaces,
+    const std::string& value) {
+  // Parse first — fast error for completely invalid strings
+  cfg::PortProfileID requestedProfile = ProfileValidator::parseProfile(value);
+
+  // PROFILE_DEFAULT: skip hardware validation, reset both fields to defaults
```
joseph5wu: Currently the only usecase for PROFILE_DEFAULT is hyperport:
https://github.com/facebook/fboss/blob/main/fboss/agent/platforms/sai/SaiPlatformPort.cpp#L61
while the regular ports will always have a working profile thanks to our platform_mapping_v2.

I'd suggest to remove this extra logic to deal with the `PROFILE_DEFAULT`.
And always use `ProfileValidator` to validate whether a requested profile is allowed for such interface

fboss/cli/fboss2/commands/config/interface/CmdConfigInterface.cpp
```diff
@@ -113,8 +115,55 @@ InterfacesConfig::InterfacesConfig(std::vector<std::string> v)
...
+      if (port) {
+        port->profileID() = cfg::PortProfileID::PROFILE_DEFAULT;
+        port->speed() = cfg::PortSpeed::DEFAULT;
+      }
+    }
+    return "profile=PROFILE_DEFAULT, speed=auto";
+  }
+
+  // Non-default: validate against hardware + optics.
+  // Build validator once (queries Agent + QSFP); reuse across all ports.
```
joseph5wu, line 126: Like the comments above. No need for special treatment for DEFAULT profile

fboss/cli/fboss2/commands/config/interface/ProfileValidation.cpp
```diff
@@ -0,0 +1,166 @@
...
+    const std::string& profileStr) {
+  cfg::PortProfileID result{};
+  if (!apache::thrift::TEnumTraits<cfg::PortProfileID>::findValue(
+          toUpper(profileStr), &result)) {
+    throw std::invalid_argument(
+        fmt::format(
+            "Invalid profile value '{}': not a valid PortProfileID. "
+            "Valid values include: PROFILE_DEFAULT, PROFILE_10G_1_NRZ_NOFEC, "
+            "PROFILE_100G_4_NRZ_RS528_COPPER, PROFILE_400G_8_PAM4_RS544X2N_OPTICAL, "
+            "... (61 values total, see PortProfileID enum in switch_config.thrift)",
```
joseph5wu: Don't mention "61 values total" since the PortProfileID enum will keep growing

fboss/cli/fboss2/commands/config/interface/ProfileValidation.cpp
```diff
@@ -0,0 +1,166 @@
...
+    throw std::invalid_argument(
+        fmt::format("No supported profiles found for port {}", portName));
+  }
+
+  // Extract profile IDs from the map (keys only)
+  const auto& profileMap = portConfig.supportedProfiles().value();
+  std::vector<cfg::PortProfileID> profiles;
+  profiles.reserve(profileMap.size());
+  for (const auto& [profileId, _] : profileMap) {
+    profiles.push_back(profileId);
```
joseph5wu: Please consider to use the PlatformMapping::getAllPortProfiles() for the same functionality 
https://github.com/facebook/fboss/blob/main/fboss/agent/platforms/common/PlatformMapping.h#L143C3-L143C57

joseph5wu (changes requested):
Please address the comments.
Overall it looks pretty good for me.

fboss/cli/fboss2/commands/config/interface/ProfileValidation.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/interface/ProfileValidation.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/interface/CmdConfigInterface.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/interface/CmdConfigInterface.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/interface/ProfileValidation.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  try {
+    auto qsfpClient =
+        utils::createClient<apache::thrift::Client<QsfpService>>(hostInfo);
+    qsfpClient->sync_getAllPortSupportedProfiles(qsfpSupportedProfiles_, true);
+  } catch (const std::exception&) {
+    // QsfpService unavailable — fall back to PlatformMapping in
+    // getSupportedProfiles()
+  }
+
```
joseph5wu, line 50: Let's only use agent to fetch the cfg::PlatformMapping thrift content here.
We have common lib to get `std::map<std::string, std::vector<cfg::PortProfileID>>` for all supported profiles for ports in a platform.

https://github.com/facebook/fboss/blob/main/fboss/agent/platforms/common/PlatformMapping.h#L143

joseph5wu (commented):
Not sure why this comment didn't send out last time.
Let's remove the dependency of qsfp in the ProfileValidation.cpp. I think using agent to fetch the cfg::PlatformMapping and then create a PlatformMapping obj to deal with the validation logic should be sufficient.

fboss/cli/fboss2/commands/config/interface/ProfileValidation.cpp
hillol-nexthop: Are we sure about this @joseph5wu ? What if the agent returns a subset of profiles which is not supported by the optics in that particular port?

joseph5wu: Now thinking about it. It's safe to query qsfp thrift api here since you're always fetching agent first. Let's keep it then

PR #1095, opened 2026-04-17
[Nexthop][fboss2-dev] Add config and delete interface ipv6 ndp CLI commands

hillol-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

Based on https://github.com/facebook/fboss/pull/1072

# Summary

Adds two new fboss2-dev CLI commands for configuring and deleting IPv6 Neighbor Discovery (NDP) settings on interfaces:

- `config interface <intf> ipv6 ndp <attr> [<value>] ...` — set NDP attributes
- `delete interface <intf> ipv6 ndp <attr> ...` — reset NDP attributes to defaults

**Supported attributes:** `ra-interval`, `ra-lifetime`, `hop-limit`, `prefix-valid-lifetime`, `prefix-preferred-lifetime`, `managed-config-flag`, `other-config-flag`, `ra-address`

## Implementation

- Added `CmdConfigInterfaceIpv6` and `CmdConfigInterfaceIpv6Ndp` under `commands/config/interface/ipv6/` to handle NDP configuration via `cfg::Interface::ndp()`.
- Added `CmdDeleteInterface`, `CmdDeleteInterfaceIpv6`, and `CmdDeleteInterfaceIpv6Ndp` under `commands/delete/interface/` to reset NDP attributes to their defaults.
- `NdpConfigAttrs` parses multi-attribute token lists; `NdpDeleteAttrs` validates a list of attribute names to reset.
- Both commands use `ConfigSession` for config persistence. No Thrift RPCs are made — NDP is a pure config operation.
- Registered both command trees in `CmdListConfig.cpp` and added `OBJECT_ARG_TYPE_ID_NDP_CONFIG_ATTRS` / `OBJECT_ARG_TYPE_ID_NDP_DELETE_ATTRS` to `CmdUtilsCommon.h`.
- Updated both `BUCK` and `cmake/CliFboss2.cmake` build files.

# Test Plan

**Unit tests:** `test/config/CmdConfigInterfaceIpv6NdpTest.cpp`, `test/config/CmdDeleteInterfaceIpv6NdpTest.cpp`
- Valid attribute set/reset per attribute
- Invalid value format → `std::invalid_argument`
- Out-of-range hop-limit (0-255)
- Multi-port update
- Multiple attributes in one command
- Idempotent delete (no NDP config on interface)
- Case-insensitive attribute names

**Integration tests:** `test/integration_test/ConfigInterfaceIpv6NdpTest.cpp`, `test/integration_test/DeleteInterfaceIpv6NdpTest.cpp`
```
/tmp/fboss2_integration_test --gtest_filter=ConfigInterfaceIpv6Ndp*:DeleteInterfaceIpv6Ndp*
```


joseph5wu (changes requested):
@hillol-nexthop : Please consider to keep the ndp cli similar to arp:
https://github.com/facebook/fboss/pull/1104
```
fboss2-dev config arp <attr> <value>
```
Our FBOSS ndp settings is per system and not per interface. Right now this implementation here treats ndp for interface.

Also please use the keyword `ndp` in the cli instead of `nd`

hillol-nexthop:
> @hillol-nexthop : Please consider to keep the ndp cli similar to arp: #1104
> 
> ```
> fboss2-dev config arp <attr> <value>
> ```
> 
> Our FBOSS ndp settings is per system and not per interface. Right now this implementation here treats ndp for interface.
> 
> Also please use the keyword `ndp` in the cli instead of `nd`

@joseph5wu The switch_config.thrift defines NDP config at per interface level and not global. If we want to make a model like global default/template + per interface override then the thrift schema has to change which needs bigger design discussion.

hillol-nexthop:
Updated the PR with the latest revision of this change, rebased onto current main:

- Stronger argument validation in the `config`/`delete` handlers (shared attribute-name set via `ndpAttrNames()`, range checks, duplicate-attribute rejection)
- Integration-test helpers (`getInterfaceIdForPort()`, `getNdpConfig()`); tests now restore the NDP attributes they touch in TearDown
- clang-tidy fixes

The CLI keyword is `ndp` (`config interface <intf> ipv6 ndp <attr> <value>`). Per the earlier discussion, the config stays per-interface to match `cfg::Interface::ndp` in switch_config.thrift.

fboss/cli/fboss2/BUCK
```diff
@@ -406,6 +406,8 @@ cpp_library(
         "commands/delete/config/facebook/CmdDeleteConfigPatcherAgent.cpp",
         "commands/delete/config/facebook/CmdDeleteConfigUtils.cpp",
         "commands/facebook/rage/CmdRage.cpp",
+        "commands/delete/interface/ipv6/CmdDeleteInterfaceIpv6.cpp",
```
joseph5wu, line 409: Please make sure all the config related changes are only included in the `fboss2-config-lib` rather than `fboss2-lib` so that we don't have to worry config changes are included in our production fboss2 tool

fboss/cli/fboss2/BUCK
```diff
@@ -639,6 +641,8 @@ cpp_library(
         "commands/debug/facebook/bgp/neighbors/CmdDebugBgpNeighborsStart.h",
         "commands/delete/config/CmdDeleteConfig.h",
         "commands/delete/config/facebook/CmdDeleteConfigPatcherAgent.h",
+        "commands/delete/interface/ipv6/CmdDeleteInterfaceIpv6.h",
```
joseph5wu, line 644: You missed the parent header(same as the cpp in the src directory) here:
```
"commands/delete/interface/CmdDeleteInterface.h",
```

joseph5wu (commented):
I've fixed the comments below in my internal diff. 
Please make sure to not make the same mistakes in the future diffs

joseph5wu:
@hillol-nexthop :
fbcode/fboss/github/cmake/CliFboss2TestIntegrationTest.cmake: Two new integration test files (ConfigInterfaceIpv6NdpTest.cpp and DeleteInterfaceIpv6NdpTest.cpp) were added to BUCK but not to the CMake target, so they will not be compiled into fboss2_integration_test in OSS builds.

In your test plan, you explicitly said you've tested the integration tests which this PR won't be able to do so without properly updating the cmake. Please make sure this is just a mistake from rebasing instead of not executing a proper testing

PR #1104, opened 2026-04-18
[Nexthop][fboss2-dev] Add fboss2-dev config arp CLI subcommands

benoit-nexthop:
# Summary

Adds a new `fboss2-dev config arp <attr> <value>` command family for tuning the FBOSS agent's ARP/NDP neighbor-table timers at runtime. Four attributes are supported, all can be set in a hitless config reload (no agent restart required):

- `timeout`        -> SwitchConfig.arpTimeoutSeconds
- `age-interval`   -> SwitchConfig.arpAgerInterval
- `max-probes`     -> SwitchConfig.maxNeighborProbes
- `stale-interval` -> SwitchConfig.staleEntryInterval

Note: I was going to implement a `refresh` subcommand as well, to set `arpRefreshSeconds`. That field is declared in `switch_config.thrift` but is **not** read by `ApplyThriftConfig.cpp` and is absent from `switch_state.thrift`, so setting it would be a no-op in the agent. It is explicitly excluded from `kArpValidAttrs` and ignored for now.

# Test Plan

## Unit tests

```
$ ./fboss/oss/scripts/bazel.sh test \
    //fboss/cli/fboss2/test/config:cmd_config_test \
    --test_filter='CmdConfigArpTestFixture.*'

[==========] Running 10 tests from 1 test suite.
[ RUN      ] CmdConfigArpTestFixture.argValidation_valid
[       OK ] CmdConfigArpTestFixture.argValidation_valid (84 ms)
[ RUN      ] CmdConfigArpTestFixture.argValidation_badArity
[       OK ] CmdConfigArpTestFixture.argValidation_badArity (30 ms)
[ RUN      ] CmdConfigArpTestFixture.argValidation_unknownAttr
[       OK ] CmdConfigArpTestFixture.argValidation_unknownAttr (34 ms)
[ RUN      ] CmdConfigArpTestFixture.argValidation_nonInteger
[       OK ] CmdConfigArpTestFixture.argValidation_nonInteger (32 ms)
[ RUN      ] CmdConfigArpTestFixture.argValidation_negative
[       OK ] CmdConfigArpTestFixture.argValidation_negative (31 ms)
[ RUN      ] CmdConfigArpTestFixture.setTimeout
[       OK ] CmdConfigArpTestFixture.setTimeout (82 ms)
[ RUN      ] CmdConfigArpTestFixture.setAgeInterval
[       OK ] CmdConfigArpTestFixture.setAgeInterval (103 ms)
[...]

PR #1105, opened 2026-04-18
[Nexthop][fboss2-dev] Add fboss2-dev config load-balancing CLI subcommands

benoit-nexthop:
# Summary

Adds two sibling command families under `fboss2-dev config load-balancing`:

- `fboss2-dev config load-balancing ecmp <attr> <value>` operates on the `LoadBalancerID::ECMP` entry of SwitchConfig.loadBalancers
- `fboss2-dev config load-balancing lag <attr> <value>` operates on the `LoadBalancerID::AGGREGATE_PORT` entry

Both accept the same six attributes, applied hitlessly:

- `hash-fields-ipv4 <src-ip[,dst-ip]|none>`
- `hash-fields-ipv6 <src-ip,dst-ip,flow-label in any combination, or none>`
- `hash-fields-transport <src-port[,dst-port]|none>`
- `hash-fields-mpls <top,second,third in any combination, or none>`
- `hash-algorithm <crc16-ccitt|crc32-lo|crc32-hi|crc32-ethernet-lo|crc32-ethernet-hi|crc32-koopman-lo|crc32-koopman-hi|crc>`
- `hash-seed <integer|default>` (`default` unsets the optional seed, falling back to the deterministic-by-MAC default)

If the user targets a `LoadBalancerID` that isn't already present in the config, an entry is created on the fly (`CRC16_CCITT` default, empty field sets). This makes the CLI usable on freshly-initialized switches that only have one side configured.

All edits are classified HITLESS: the SAI layer dispatches LoadBalancersDelta changes through `SaiSwitchManager::changeLoadBalancer` without any *ChangeProhibited guard.

Both ECMP and LAG share the same argument parsing and apply helper (`LoadBalancingConfigArgs` and `applyLoadBalancerConfig`) so the two handler classes are thin wrappers that only select the target LoadBalancerID.

Note: this PR is stacked on top of PR #1104, which must be merged first.

## LAG integration tests — temporarily disabled

The three LAG integration tests (`DISABLED_LagHashAlgorithm`, `DISABLED_LagHashSeed`, `DISABLED_LagHashFieldsIpv4`) are disabled in this PR and re-enabled in the follow-up **#1309**.

During integration testing we found two bugs in `SaiSwitchManager::addOrUpdateLagLoadBalancer` that are fixed in #1309:

1. **UDF group ID drop** — `getOrCreate` was called without passing `udfGroupIds`, so any hash update on a DUT with LAG UDF groups would silently drop them.
2. **Empty hash-fields no-op** — clearing all IPv4/IPv6 fields left the old hash object still attached to the switch attribute instead of detaching and releasing it.

All three LAG tests call `addOrUpdateLagLoadBalancer` and are affected by bug 1 on DUTs with UDF groups configured. They are re-enabled in #1309 once the agent-side fix is in place.

**Merge order: this PR first, then #1309.**

# Test Plan

## Unit tests

```
$ ./fboss/oss/scripts/bazel.sh test //fboss/cli/fboss2/test/config:cmd_config_test \
    --test_filter='CmdConfigLoadBalancingTestFixture.*' --test_output=errors

//fboss/cli/fboss2/test/config:cmd_config_test                           PASSED in 0.8s
Executed 1 out of 1 test: 1 test passes.
```

18 tests in `CmdConfigLoadBalancingTestFixture` covering:

- arg-parsing (arity / unknown attrs)
- `applyLoadBalancerConfig` attribute dispatch (all six attrs, including
  the `none` / `default` sentinel values and the create-missing-entry
  branch)
- `queryClient()` end-to-end through `ConfigSession` for both ECMP and LAG

## Integration tests

ECMP integration tests pass. LAG integration tests are disabled here and verified in #1309.

```
$ ssh $DUT "/tmp/fboss2_it_nos6212e --gtest_filter='ConfigLoadBalancingTest.*'"

[----------] 6 tests from ConfigLoadBalancingTest
[  PASSED  ] 6 tests.
```

`EcmpHashAlgorithm`, `EcmpHashSeed`, `EcmpHashFieldsIpv4`, `EcmpHashFieldsIpv6`, `EcmpHashFieldsTransport`, `EcmpHashFieldsMpls`.

## Sample session

```
$ fboss2-dev config load-balancing ecmp hash-algorithm crc32-hi
Set ecmp hash-algorithm to crc32-hi
$ fboss2-dev config load-balancing ecmp hash-seed 42
Set ecmp hash-seed to 42
$ fboss2-dev config load-balancing ecmp hash-fields-ipv4 src-ip,dst-ip
Set ecmp hash-fields-ipv4 to src-ip,dst-ip
$ fboss2-dev config load-balancing ecmp hash-fields-mpls none
Set ecmp hash-fields-mpls to none
$ fboss2-dev config load-balancing lag hash-algorithm crc
Set lag hash-algorithm to crc
$ fboss2-dev config load-balancing lag hash-seed default
Cleared lag hash-seed (using deterministic default)
```


fboss/cli/fboss2/utils/CmdUtilsCommon.h
```diff
@@ -91,6 +91,7 @@ enum class ObjectArgTypeId : uint8_t {
   OBJECT_ARG_TYPE_L2_LEARNING_MODE,
   OBJECT_ARG_TYPE_ID_INTERFACES_CONFIG,
   OBJECT_ARG_TYPE_ARP_CONFIG,
+  OBJECT_ARG_TYPE_LOAD_BALANCING_CONFIG,
```
joseph5wu: Can you use the new way per-Traits addCliArg method instead of defining the new enum?

joseph5wu (changes requested):
You might need to rebase and replace the central ObjectArgTypeId enum approach

fboss/cli/fboss2/utils/CmdUtilsCommon.h
benoit-nexthop: Done.

PR #1106, opened 2026-04-18
[Nexthop][fboss2-dev] Skip kernel-interface MTU check when fboss<vlan> tap is absent

benoit-nexthop:
# Summary

`ConfigInterfaceMtuTest.SetAndVerifyMtu` asserts the kernel-side MTU of `fboss<vlan>` matches what `config interface ... mtu <N>` just committed. The test is already written to *skip* the kernel check when the tap doesn't exist:

```cpp
int kernelMtu = getKernelInterfaceMtu(*interface.vlan);
if (kernelMtu > 0) {
  EXPECT_EQ(kernelMtu, newMtu);
} else {
  XLOG(INFO) << "  Skipped: Kernel interface fboss" << *interface.vlan
             << " not found";
}
```

— but `getKernelInterfaceMtu()` throws on `ip link show` failure rather than returning 0, so the else-branch is unreachable and the test always fails on platforms where the first eth interface's VLAN has no L3 RIF.

On NH-4010-F in our lab with default config, eth1/1/1 carries VLAN 9 but only `fboss10` and `fboss2001+` kernel taps exist, so every full run of `fboss2_integration_test` reports this one test as failed even though the agent-side MTU set + commit + reload worked fine.

## Fix

Return 0 from `getKernelInterfaceMtu()` when the interface isn't found, which makes the existing `if (kernelMtu > 0)` skip branch reachable. The test now passes on platforms without the matching tap while still enforcing the kernel-MTU invariant where a tap does exist.

Also pull in the extra direct `#include`s that clang-include-cleaner flagged on the file while we're here.

# Test Plan

On NH-4010-F, full `fboss2_integration_test` with this change goes from 22/23 to **23/23 green**:

```
$ /tmp/fboss2_integration_test
[==========] 23 tests from 9 test suites ran. (97770 ms total)
[  PASSED  ] 23 tests.
```

Before the fix the single failure was:

```
[  FAILED  ] ConfigInterfaceMtuTest.SetAndVerifyMtu
C++ exception with description "Kernel interface fboss9 not found
('ip link show' returned 1)" thrown in the test body.
```

### DIFF
--- a/fboss/cli/fboss2/test/integration_test/ConfigInterfaceMtuTest.cpp
+++ b/fboss/cli/fboss2/test/integration_test/ConfigInterfaceMtuTest.cpp
@@ -69,10 +69,15 @@ TEST_F(ConfigInterfaceMtuTest, SetAndVerifyMtu) {
   XLOG(INFO) << "[Step 5] Verifying kernel interface MTU...";
   ASSERT_TRUE(interface.vlan.has_value());
   int kernelMtu = getKernelInterfaceMtu(*interface.vlan);
-  EXPECT_EQ(kernelMtu, newMtu)
-      << "Kernel MTU is " << kernelMtu << ", expected " << newMtu;
-  XLOG(INFO) << "  Verified: Kernel interface fboss" << *interface.vlan
-             << " has MTU " << kernelMtu;
+  if (kernelMtu > 0) {
+    EXPECT_EQ(kernelMtu, newMtu)
+        << "Kernel MTU is " << kernelMtu << ", expected " << newMtu;
+    XLOG(INFO) << "  Verified: Kernel interface fboss" << *interface.vlan
+               << " has MTU " << kernelMtu;
+  } else {
+    XLOG(INFO) << "  Skipped: Kernel interface fboss" << *interface.vlan
+               << " not found";
+  }
 
   // Step 6: Restore original MTU
   XLOG(INFO) << "[Step 6] Restoring original MTU (" << originalMtu << ")...";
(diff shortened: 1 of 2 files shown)


joseph5wu:
@benoit-nexthop The current `ConfigInterfaceMtuTest.SetAndVerifyMtu` always `EXPECT_EQ(kernelMtu, newMtu)` rather than if-else. Can you confirm whether this PR is still needed?
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/test/integration_test/ConfigInterfaceMtuTest.cpp#L71

benoit-nexthop:
> @benoit-nexthop The current `ConfigInterfaceMtuTest.SetAndVerifyMtu` always `EXPECT_EQ(kernelMtu, newMtu)` rather than if-else. Can you confirm whether this PR is still needed? https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/test/integration_test/ConfigInterfaceMtuTest.cpp#L71

Yes it's needed. In fact more is needed. The problem is that in [PR 893](https://github.com/facebook/fboss/pull/893#pullrequestreview-4144247842) some of the diff in the PR was not merged upstream, I don't know why. But there was a check originally in the code for `if (kernelMtu > 0) {` but that is missing in the [commit merged upstream](https://github.com/facebook/fboss/commit/06ff27b07202b5198eaace19dd0ef2dd056a3ba9#diff-e9f4370b2829a4c1b2944d176da2f5b95b31b469e7c72c4e9d02d54f1e433b2a). We need to reintroduce that check.

PR #1107, opened 2026-04-18
[Nexthop][Distro] Fix fboss_init.sh startup and package missing shared libs

benoit-nexthop:
# Summary

`fboss_init.sh` wasn't sourcing `setup_fboss_env`, so `weutil` couldn't find its shared libraries and silently failed to generate `fruid.json`. Downstream services would then crash on startup because `fruid.json` was missing.

`setup_fboss_env`: derive FBOSS root from the script's own location instead of `$PWD`, and add an idempotency guard to prevent duplicate `PATH` and `LD_LIBRARY_PATH` prepends when the script is sourced multiple times.

`fboss_init.sh`: `source setup_fboss_env` at startup so `weutil` resolves its shared libraries; make `generate_fruid()` propagate failure with `return 1`; make `main()` exit 1 if fruid generation fails so systemd marks `fboss_init.service` as failed rather than succeeded.

`fboss_cmd_find.sh`: `source setup_fboss_env` before exec so every FBOSS command runs with the correct `PATH` and `LD_LIBRARY_PATH`.

`package.py`: the thrift-python migration (commit f9d8d65) switched FBOSS's manifest dependencies from the base variants (folly, wangle, fizz, mvfst, fmt, zstd, fbthrift) to their `-python` variants, which are built with `BUILD_SHARED_LIBS=ON`. As a result, every FBOSS binary (`platform_manager`, `weutil`, `fsdb`, ...) now links against `libfolly.so`, `libwangle.so`, `libfmt.so`, `libfizz.so`, `libmvfst.so` at runtime, not just Python extensions. Bundle these shared libraries in the forwarding-stack and platform-stack tarballs via a new `_find_getdeps_libs()` helper that locates `.so` files under each package's getdeps install prefix. `LIB_NAME_OVERRIDES` handles packages whose getdeps name differs from the library filename (e.g. `fmt-python` builds `libfmt.so`). Also add a `lib64` -> `lib` symlink in each tarball so binaries with `RPATH`/`RUNPATH` pointing to `lib64/` resolve correctly, and bundle `libunwind.so` from the LLVM build container since CentOS does not ship a compatible version and getdeps does not install it.

# Test Plan

These fixes were needed to properly image DUTs with the latest distro image build. Without them no FBOSS agent would come up.

### DIFF
--- a/fboss-image/image_builder/templates/centos-09.0/root_files/usr/local/bin/fboss_init.sh
+++ b/fboss-image/image_builder/templates/centos-09.0/root_files/usr/local/bin/fboss_init.sh
@@ -3,6 +3,9 @@
 
 set -e
 
+# shellcheck source=/opt/fboss/bin/setup_fboss_env
+source /opt/fboss/bin/setup_fboss_env
+
 FBOSS_SHARE="/opt/fboss/share"
 COOP_DIR="/etc/coop"
 FRUID_FILE="/var/facebook/fboss/fruid.json"
@@ -61,11 +64,12 @@ generate_fruid() {
 
   mkdir -p "$(dirname "$FRUID_FILE")"
 
-  if /opt/fboss/bin/weutil --json >"$FRUID_FILE" 2>/dev/null; then
+  if weutil --json >"$FRUID_FILE"; then
     log "Generated fruid.json: $FRUID_FILE"
   else
     error "Failed to generate fruid.json"
     rm -f "$FRUID_FILE"
+    return 1
   fi
 }
 
@@ -125,7 +129,9 @@ main() {
 
   create_distro_base_snapshot
   setup_coop_configs "$platform_dir"
-  generate_fruid
+  if ! generate_fruid; then
+    exit 1
+  fi
   enable_hw_agents "$platform_dir"
 
   log "FBOSS initialization complete"
--- a/fboss-image/image_builder/templates/centos-09.0/root_files/usr/local/lib/fboss_cmd_find.sh
+++ b/fboss-image/image_builder/templates/centos-09.0/root_files/usr/local/lib/fboss_cmd_find.sh
@@ -40,10 +40,8 @@ if [ -x "$update_path" ]; then
   exec "$update_path" "$@"
 else
   if [ -f /opt/fboss/bin/setup_fboss_env ]; then
-    pushd /opt/fboss >/dev/null
     # shellcheck source=/opt/fboss/bin/setup_fboss_env
-    source ./bin/setup_fboss_env
-    popd >/dev/null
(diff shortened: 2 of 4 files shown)


benoit-nexthop:
Chain of events that led us here:
1. 518c49ff889 (Jan 22) — Created `-python` manifest variants with `BUILD_SHARED_LIBS=ON` because Python extension wheels need `-fPIC` C++ libs. At this point, FBOSS was unaffected.
2. 108845a0fbf (Mar 7) — Unconditionally enabled `BUILD_SHARED_LIBS` on `gflags` and `glog` (previously conditional on `--shared_libs=on`), and added `-fPIC` to `libsodium`, `libunwind`, `xxhash`, `xz` so they could be linked into shared libs. This made `gflags`/`glog` always shared, which is why `libunwind` is packaged in `package.py`.
3. f9d8d654a0b (Apr 6) — At this point FBOSS binaries are going dynamic linked. To support the `thrift-py` to `thrift-python` migration (Cython extensions need `folly-python` and `fbthrift-python` for ABI compatibility), FBOSS's manifest was switched to the `-python` variants wholesale. Consequence: `platform_manager`, `weutil`, `fsdb`, etc. now all link against shared libs, therefore those shared libs now need to be bundled in the distro image.

kevin645:
likely needs to be rebased. I think @shiva-menta found a similar issue in his testing recently as well.

benoit-nexthop:
Change rebased. I don't know about your `/updates/` stuff but now for the main `setup_fboss_env` there is no need to `pushd` / `popd` anymore, it works independently of the current working directory.

KevinYakar:
Wondering why we use package.py instead of package-fboss.py? We have not had problems with libraries using the latter, we copy libs from the container rather than the host machine which ends up satisfying the required libs noted by ldd. I know package.py is probably more specific to distro, just wanted some context here

travisb-nexthop:
> Wondering why we use package.py instead of package-fboss.py? We have not had problems with libraries using the latter, we copy libs from the container rather than the host machine which ends up satisfying the required libs noted by ldd. I know package.py is probably more specific to distro, just wanted some context here

@KevinYakar The primary reason is that Distro requires packaging Platform Stack and Forwarding Stack separately. Further each of those need to package 'product' binaries and 'test' binaries separately. To make that distinction and to further control what ends up in the image requires explicitly listing the contents explictly.

package-fboss.py doesn't do that and we were hesitant to perform major backwards-incompatible work to make it do that. Further, package-fboss.py packaged a lot of things (like .a files) which we don't want in the Distro at all.

Beyond even that, package.py prioritizes speed more highly because it is on the critical path for constructing Distro images incrementally. It is substantially faster than package-fboss.py by solving a slightly different problem.

PR #1133, opened 2026-04-26
[Nexthop][fboss2-dev] Add fboss2-dev config copp CLI subcommands

benoit-nexthop:
# Summary

Adds a `config copp` command family for control-plane-policing
(CoPP) state. All edits are applied hitlessly by the SAI layer
(`SaiHostifManager::processHostifDelta` → `changeCpuQueue()` /
trap-group update, no `*ChangeProhibited()` guard):

- `fboss2-dev config copp cpu-queue <id>` — ensure the queue exists
- `fboss2-dev config copp cpu-queue <id> name <string>` — set
  `PortQueue.name`
- `fboss2-dev config copp cpu-queue <id> rate-limit kbps <max>` —
  set `PortQueue.portQueueRate.kbitsPerSec.maximum`
- `fboss2-dev config copp cpu-queue <id> rate-limit pps <max>` —
  set `PortQueue.portQueueRate.pktsPerSec.maximum`
- `fboss2-dev config copp reason <reason-name> queue <id>` —
  upsert into `cpuTrafficPolicy.rxReasonToQueueOrderedList`

Both handler classes (`CmdConfigCoppCpuQueue`, `CmdConfigCoppReason`)
live in a single pair of `CmdConfigCopp.{h,cpp}` files under
`commands/config/copp/` since the two families operate on
closely-related CoPP state.

Two non-obvious details called out in the commit message:

- A trivial `CmdConfigCopp` parent handler ("Incomplete command, please
  use 'cpu-queue' or 'reason' subcommand") is required so that
  `addCommandBranch()` increments depth between the parent and the
  leaves — otherwise both leaves would register their positional args
  at the same `CmdArgsLists::data_[0]` slot and collide.
- `OBJECT_ARG_TYPE_COPP_REASON_CONFIG` is registered with
  `->required()->expected(3)` in `CmdSubcommands.cpp` to force CLI11 to
  route positionals to the option instead of reclassifying a reason
  value (`arp`, `ndp`, `bgp`, ...) as a sibling subcommand via the
  last-gasp subcommand lookup. The trigger was `config copp reason arp
  queue 0` — CLI11 was matching "arp" against `config arp`.

# Test Plan

Unit tests — 20 new tests covering arg validation (arity, unknown
attrs, non-integer, out-of-range, missing `queue` keyword) and
`queryClient()` for each attribute. Seed config mirrors the
`cpuQueues` + `cpuTrafficPolicy` shape from a production
`agent.conf` on a deployed RSW device:

```
$ ./fboss/oss/scripts/bazel.sh test \
    //fboss/cli/fboss2/test/config:cmd_config_test \
    --test_filter='CmdConfigCoppTestFixture.*' --test_output=all
...
[  PASSED  ] 20 tests.
```

Integration tests on NH-4010-F:

```
$ /tmp/fboss2_integration_test --gtest_filter='ConfigCoppTest.*'
[==========] Running 4 tests from 1 test suite.
[ RUN      ] ConfigCoppTest.CpuQueueEnsureExistsIsNoOp
[       OK ] ConfigCoppTest.CpuQueueEnsureExistsIsNoOp (59 ms)
[ RUN      ] ConfigCoppTest.CpuQueueSetName
[       OK ] ConfigCoppTest.CpuQueueSetName (111 ms)
[ RUN      ] ConfigCoppTest.CpuQueueSetRateLimitPps
[       OK ] ConfigCoppTest.CpuQueueSetRateLimitPps (107 ms)
[ RUN      ] ConfigCoppTest.ReasonToQueueUpdate
[       OK ] ConfigCoppTest.ReasonToQueueUpdate (115 ms)
[==========] 4 tests from 1 test suite ran. (397 ms total)
[  PASSED  ] 4 tests.
```

Sample CLI session on the same DUT (commit + restore each change so
the DUT is left unchanged):

```
$ fboss2-dev config copp cpu-queue 1 name cpuQueue-sample-demo
Set cpu-queue 1 name to 'cpuQueue-sample-demo'
$ fboss2-dev config copp cpu-queue 1 rate-limit pps 2500
Set cpu-queue 1 rate-limit pps max to 2500
$ fboss2-dev config copp reason bgp queue 0
Updated reason BGP -> queue 0
$ fboss2-dev config session diff
--- current live config
+++ session config
@@ cpuQueues id=1:
-        "name": "cpuQueue-default",
+        "name": "cpuQueue-sample-demo",
         "portQueueRate": {
           "pktsPerSec": {
-            "maximum": 1000,
+            "maximum": 2500,
@@ rxReasonToQueueOrderedList[BGP]:
-          "queueId": 9,
+          "queueId": 0,
Config session committed successfully as c99d21b and config reloaded
for fboss_sw_agent.
```


fboss/cli/fboss2/CmdSubcommands.cpp
```diff
@@ -353,6 +353,34 @@ CLI::App* CmdSubcommands::addCommand(
               "<port-list> [<attr> <value> ...] where <attr> is one "
               "of: description, mtu");
           break;
+        case utils::ObjectArgTypeId::OBJECT_ARG_TYPE_COPP_CPU_QUEUE_CONFIG:
```
joseph5wu: Let's use the new `addCliArgs` approach from https://github.com/facebook/fboss/pull/1112

joseph5wu (commented):
Please rebase on top of https://github.com/facebook/fboss/pull/1112

fboss/cli/fboss2/CmdSubcommands.cpp
benoit-nexthop: Done.

fboss/cli/fboss2/commands/show/aggregateport/CmdShowAggregatePort.cpp
joseph5wu, line 1: Can we remove the change on this class for this PR because it's irrelevant. 
Besides it also stops me to import because it detects there'll be conflicts on this cpp file

joseph5wu (changes requested):
Please rebase to the latest master

fboss/cli/fboss2/CmdListConfig.cpp
```diff
@@ -117,6 +118,26 @@ const CommandTree& kConfigCommandTree() {
        commandHandler<CmdConfigAppliedInfo>,
        argRegistrar<CmdConfigAppliedInfoTraits>},
 
+      {
+          "config",
+          "copp",
+          "Configure CoPP (control-plane policing) settings",
+          commandHandler<CmdConfigCopp>,
+          argTypeHandler<CmdConfigCoppTraits>,
```
joseph5wu, line 126: This should be `argRegistrar` and the following two subcommands should also use `argRegistrar`.
I noticed this due to our on-diff fboss2_integration_tests failed
Not sure why your internal CI didn't catch this issue

joseph5wu (commented):
I've fixed this fboss2_integration_tests internally.
Please make sure next time when rebasing commits, please run comprehensive tests before requesting PR reviews

PR #1144, opened 2026-04-30
[Nexthop] Distro tajo support

travisb-nexthop:
**Pre-submission checklist**
- [X] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [X] `pre-commit run`

# Summary

Add support for building Distro Images with the Tajo SAI.

Please see the individual commits for the sequential details. Unfortunately all the changes are required to build successfully.

# Test Plan

Modify the from_source.json manifest to separately provide a Broadcom SAI and a Tajo SAI and verify that Distro images are successfully built for both.

PR #1147, opened 2026-05-02
[Nexthop] Convert distro build from build-helper.py to use npu flags

travisb-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

When run-getdeps.py gained the --npu-* flags, it began to default to
configuring the FakeSAI by default if those flags are not passed.

In the Distro build which was continuing to use the older
build-helper.py mechanism, this results in BOTH the FakeSAI and real
SAI being defined in the single build. This causes problems and fails.

Resolve this by moving the Distro build to the newer --npu-*
run-getdeps.py flags.

Note that this requires a slight change to the way that the SAI paths
are symlinked. run-getdeps.py with the --npu-* options uses /opt/sdk
inside the build container so the Distro build script cannot use that
path for the NPU SAI. Conveniently this symlink isn't necessary at all
since the extracted NPU SAI is already available under /deps. Use
those paths directly instead.

# Test Plan

These contents have been pulled out of https://github.com/facebook/fboss/pull/1144 where full images using a real SAI were built.


travisb-nexthop:
This is built on top of https://github.com/facebook/fboss/pull/1019

kevin645:
SAI_DIR="/deps/deps/phy_sai-extracted" is this a typo ?

PR #1091, opened 2026-04-16
[Nexthop][fboss2-dev] Add VlanManager for auto-VLAN creation in config vlan subcommands

benoit-nexthop:
# Summary

Adds a `VlanManager` class that auto-creates a VLAN (and its required `cfg::Interface`) in the session config whenever a `config vlan <id>` subcommand references a VLAN that does not yet exist.

Previously, `config vlan <id> static-mac add` and
`config vlan <id> port taggingMode` would throw an error if the VLAN was not already present in the running config, forcing operators to create the VLAN manually first. Now these commands create the VLAN on demand and print a one-line notice when they do:

```
Created VLAN <id>
```

The agent requires every VLAN to have a matching `cfg::Interface` entry or it rejects the commit. `VlanManager::createVlan` therefore creates both objects atomically:
- `cfg::Vlan` with `id`, `name = "Vlan<id>"`, `routable = false`
- `cfg::Interface` with `intfID = id`, `vlanID = id`, `name = "fboss<id>"`, `routerID = 0`, `ipAddresses = []`

For `static-mac delete` on a non-existent VLAN the command now returns an idempotent "does not exist" message instead of throwing, consistent with the delete-on-non-existent pattern used elsewhere.

# Test Plan

Unit tests (`fboss2_cmd_config_test --gtest_filter='VlanManager*'`):

```
[==========] Running 12 tests from 1 test suite.
[----------] 12 tests from VlanManagerTestFixture
[ RUN      ] VlanManagerTestFixture.createVlanNewVlan
[       OK ] VlanManagerTestFixture.createVlanNewVlan (44 ms)
[ RUN      ] VlanManagerTestFixture.createVlanAlreadyExists
[       OK ] VlanManagerTestFixture.createVlanAlreadyExists (40 ms)
[ RUN      ] VlanManagerTestFixture.createVlanSetsCorrectId
[       OK ] VlanManagerTestFixture.createVlanSetsCorrectId (40 ms)
[ RUN      ] VlanManagerTestFixture.createVlanSetsAutoGeneratedName
[       OK ] VlanManagerTestFixture.createVlanSetsAutoGeneratedName (49 ms)
[ RUN      ] VlanManagerTestFixture.createVlanSetsRoutableFalse
[       OK ] VlanManagerTestFixture.createVlanSetsRoutableFalse (44 ms)
[ RUN      ] VlanManagerTestFixture.createVlanDoesNotSaveConfig
[...]

PR #1153, opened 2026-05-05
[Nexthop][fboss2-dev] Add fboss2-dev config vlan default CLI command

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`
Based on upstream PR -> [#1091](https://github.com/facebook/fboss/pull/1091)
# Why I did it

`sw.defaultVlan` (thrift: `SwitchConfig.defaultVlan`) is always
initialized to 4094 in the base agent config and was not previously
configurable via CLI. This command exposes it so operators can redirect
untagged traffic to a different VLAN at bring-up without hand-editing
the JSON config.

# How I did it

**New command:** `fboss2-dev config vlan default <vlan-id>`, registered
in `CmdListConfig.cpp` as a sibling of the `config vlan <id>` subtree
(separate node so the integer arg resolves at the right depth).

**Implementation (`CmdConfigVlanDefault.cpp`):** The command implements
the following logic:

1. **No-op guard:** if `sw.defaultVlan` already equals the target,
return immediately with a clear message.
2. **Old-default safety check:** if the old default VLAN is used as
`ingressVlan` for any port but has no matching `cfg::Interface` entry,
refuse with a descriptive error (this is the config state that causes
`fboss_sw_agent` to crash on startup).
3. **Target VLAN lookup:** search `sw.vlans` by ID. If the target VLAN
already exists, reuse it as-is. If not, auto-create a non-routable
placeholder: `cfg::Vlan{id=vlanId, name="default_<id>", routable=false}`.
4. **Old VLAN cleanup:** if the old default VLAN has no interface and no
port still uses it as `ingressVlan`, remove it from the VLAN table.
5. **Commit:** set `sw.defaultVlan\` to the new ID and call
`session.saveConfig()`.

**New files:**
- `fboss/cli/fboss2/commands/config/vlan/CmdConfigVlanDefault.h/.cpp`
  — command implementation
- `fboss/cli/fboss2/test/config/CmdConfigVlanDefaultTest.cpp` — unit
  tests
- `fboss/cli/fboss2/test/integration_test/ConfigVlanDefaultTest.cpp` —
  integration tests
- Wired into `cmake/CliFboss2.cmake`, `cmake/CliFboss2TestConfig.cmake`,
  `cmake/CliFboss2TestIntegrationTest.cmake`, `fboss/cli/fboss2/BUCK`,
  `fboss/cli/fboss2/test/config/BUCK`,
  `fboss/cli/fboss2/test/integration_test/BUCK`

# Test Plan

**Unit tests (`CmdConfigVlanDefaultTest`)** cover the core logic paths:
- `setDefaultVlanAlreadySet` — no-op when target equals current default
- `setDefaultVlanSuccess` — happy path: creates new target VLAN, updates
  `defaultVlan`
- `setDefaultVlanUpdatesOnlyOneEntry` — no duplicate VLAN entries created
- `changeDefaultVlanMultipleTimes` — sequential changes accumulate
  correctly
- `createsNewVlanWhenNeitherExists` — auto-creates target when VLAN table
  lacks it
- `reusesExistingTargetVlan` — no duplication when target VLAN already
  present
- `succeedsWithNoDefaultVlanField` — handles Thrift zero-default when
  `defaultVlan` absent from config

**Integration tests (`ConfigVlanDefaultTest`)** run on fboss-sim:
- `SetDefaultVlanTo300` — moves eth ports off old default, sets new
  default, commits, verifies `sw.defaultVlan` via `getRunningConfig()`
  thrift, restores original state
- `NoOpWhenDefaultVlanUnchanged` — idempotency: re-setting current
  default returns a clear message with no config change
- `RefuseWhenPortOnDefaultVlanWithNoInterface` — validation guard:
  refuses change when old default VLAN is used as `ingressVlan` but has
  no interface (skipped if precondition not met in running config)
- `ChangeDefaultVlanWithPortInNonDefaultVlan` — regression: command does
  not crash when ports have already been moved off the current default
  VLAN


fboss/cli/fboss2/commands/config/vlan/CmdConfigVlanDefault.cpp
```diff
@@ -0,0 +1,116 @@
...
+        return *intf.vlanID();
+      });
+
+  bool oldVlanUsedAsIngress =
+      hasElemWithId(*swConfig.ports(), currentDefault, [](const auto& port) {
+        return *port.ingressVlan();
+      });
+
+  if (oldVlanUsedAsIngress && !oldVlanHasInterface) {
+    return fmt::format(
```
joseph5wu, line 61: The safety check returns the error message as a normal RetType string, which means the CLI exits with code 0 even when the command is refused. This is inconsistent with other config commands — for example, CmdConfigInterfaceSwitchportAccessVlan throws std::invalid_argument for validation failures, giving callers a non-zero exit code to distinguish success from refusal.

  Fix: Change return fmt::format(...) to throw std::invalid_argument(fmt::format(...)) and add #include <stdexcept>. The integration test RefuseWhenPortOnDefaultVlanWithNoInterface should then check result.exitCode != 0 and result.stderr (instead of result.stdout).

```
  // Before:
  if (oldVlanUsedAsIngress && !oldVlanHasInterface) {
    return fmt::format("Refusing to change default VLAN from {} to {} ...",
        currentDefault, vlanId, currentDefault);
  }

  // After:
  if (oldVlanUsedAsIngress && !oldVlanHasInterface) {
    throw std::invalid_argument(fmt::format(
        "Refusing to change default VLAN from {} to {} ...",
        currentDefault, vlanId, currentDefault));
  }
```

fboss/cli/fboss2/test/integration_test/ConfigVlanDefaultTest.cpp
```diff
@@ -0,0 +1,365 @@
...
+ protected:
+  folly::dynamic getRunningConfig() const {
+    HostInfo hostInfo("localhost");
+    auto client =
+        utils::createClient<apache::thrift::Client<FbossCtrl>>(hostInfo);
+    std::string configStr;
+    client->sync_getRunningConfig(configStr);
+    return folly::parseJson(configStr);
+  }
+};
```
joseph5wu, line 33: ConfigVlanDefaultTest defines its own getRunningConfig() that is identical to the one already provided by the Fboss2IntegrationTest base class. This adds unnecessary code and extra includes (FbossCtrlAsyncClient.h, CmdClientUtilsCommon.h, HostInfo.h) that aren't needed when using the inherited version.

  Fix: Remove the method override and the 3 unnecessary includes. The class body becomes just class ConfigVlanDefaultTest : public Fboss2IntegrationTest {};

```
  // Before:
  #include "fboss/agent/if/gen-cpp2/FbossCtrlAsyncClient.h"
  #include "fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.h"
  #include "fboss/cli/fboss2/utils/CmdClientUtilsCommon.h"
  #include "fboss/cli/fboss2/utils/HostInfo.h"

  class ConfigVlanDefaultTest : public Fboss2IntegrationTest {
   protected:
    folly::dynamic getRunningConfig() const {
      HostInfo hostInfo("localhost");
      auto client = utils::createClient<...>(hostInfo);
      std::string configStr;
      client->sync_getRunningConfig(configStr);
      return folly::parseJson(configStr);
    }
  };

  // After:
  #include "fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.h"

  class ConfigVlanDefaultTest : public Fboss2IntegrationTest {};
```

fboss/cli/fboss2/test/integration_test/ConfigVlanDefaultTest.cpp
```diff
@@ -0,0 +1,365 @@
...
+  // to load the config (not full ASIC init, which takes much longer).
+  commitConfig();
+  waitForAgentReady();
+  // Step 4: Verify via running config.
+  // sync_getRunningConfig reflects the committed config as soon as the agent
+  // loads it, independently of ASIC programming progress.
+  auto config = getRunningConfig();
+  EXPECT_EQ(config["sw"]["defaultVlan"].asInt(), std::stoi(targetVlan));
+  XLOG(INFO) << "[Step 4] Verified defaultVlan="
+             << config["sw"]["defaultVlan"].asInt();
```
joseph5wu, line 127: Post-commit verification accesses config["sw"]["defaultVlan"] directly. The base class now provides getSwConfigField<T>(field) which handles the sw object lookup with proper error messages. Consider using it for cleaner verification:

```
  // Before:
  auto config = getRunningConfig();
  EXPECT_EQ(config["sw"]["defaultVlan"].asInt(), std::stoi(targetVlan));

  // After:
  EXPECT_EQ(getSwConfigField<int>("defaultVlan"), std::stoi(targetVlan));
```

joseph5wu (commented):
I've addressed all the comments internally. But would like to leave them here for future reference.
Please consider to reuse the following comments for your future PRs

joseph5wu:
@vybhav-nexthop : When I was running the fboss2-integration-test on Montblanc:
```
================================================================================
Test Results for Config: montblanc/brcm-13.3.0.0_odp
--------------------------------------------------------------------------------
[  FAILED] ConfigVlanDefaultTest.SetDefaultVlanTo300 (6.530 secs)
[  PASSED] ConfigVlanDefaultTest.NoOpWhenDefaultVlanUnchanged (16.506 secs)
[ SKIPPED] ConfigVlanDefaultTest.ChangeDefaultVlanWithPortInNonDefaultVlan (16.549 secs)
[ SKIPPED] ConfigVlanDefaultTest.RefuseWhenPortOnDefaultVlanWithNoInterface (16.444 secs)
================================================================================

Ran 4 tests in 26 mins 56.536 secs
[   ERROR] 0
[  FAILED] 1
[ TIMEOUT] 0
[  PASSED] 1
[ SKIPPED] 2
[ OMITTED] 0
[ RETRIED] 0
```

  1 FAILED: SetDefaultVlanTo300

  The test tries to create VLAN 300 by running config vlan 300 static-mac add 02:00:00:00:01:2c eth1/8/1, but it fails with Invalid argument: VLAN 300 does not exist in configuration. The test assumes static-mac add will auto-create the VLAN, but it doesn't — VLAN 300 needs to exist first(https://github.com/facebook/fboss/pull/1179). This is a bug in the test's "Step 0" setup logic. The device has defaultVlan=4094 and no VLAN 300 in the table, so the test can't create its target VLAN using this approach.


  2 SKIPPED (expected, by design):

  - ChangeDefaultVlanWithPortInNonDefaultVlan — Skipped because "No eth ports on the default VLAN to move". The device's ports don't have ingressVlan=4094 (the default), so there's nothing to reassign. The GTEST_SKIP is working as intended.
  - RefuseWhenPortOnDefaultVlanWithNoInterface — Skipped because "No port uses the default VLAN as ingressVlan — cannot exercise guard". Same root cause — ports aren't on VLAN 4094. The test explicitly skips when the precondition isn't met.

I would prefer to properly adjust an interface to use the default vlan first before these two tests rather than skipping them. Try to use https://github.com/facebook/fboss/pull/791  to change an existing interface to use the default vlan at the beginning of the two tests

fboss/cli/fboss2/commands/config/vlan/CmdConfigVlanDefault.cpp
vybhav-nexthop: thanks for the fix, would be taken care in future PRs for any failures on the command

fboss/cli/fboss2/test/integration_test/ConfigVlanDefaultTest.cpp
vybhav-nexthop: thanks again, will ensure to check for duplication before introducing new functions

fboss/cli/fboss2/test/integration_test/ConfigVlanDefaultTest.cpp
vybhav-nexthop: thanks, will use the get/set config fields

vybhav-nexthop:
>
@joseph5wu so this PR was dependent on the PR mentioned in the descripton-> #1091. This PR helps us creates a vlan id of 300 , which then allows us to change the access vlans for all the ports.

with regards to the two tests getting skipped, we could go about doing it by changing the vlan for the interface. Will try to test and commit this in an upcoming PR. In our setup, we had a simulator with the required configs tested here

PR #1170, opened 2026-05-07
[Accton][minipack3bta] Gate L3_MTU_ERROR in BCM COPP test utils when trap not supported

jeankuo:
**Pre-submission checklist**
- [v] `pre-commit run`

# Summary

- Update `getCoppRxReasonToQueuesForBcm()` to conditionally include
  `L3_MTU_ERROR` mapping only when the ASIC supports
  `L3_MTU_ERROR_TRAP`.

# Test Plan

- Run COPP-related tests and confirm `L3_MTU_ERROR` mapping appears
  only when the capability is supported.
- Testing restricted to SAI_15.1.0.2_EA environment; will expand in
  future iterations.

### DIFF
--- a/fboss/agent/test/utils/CoppTestUtils.cpp
+++ b/fboss/agent/test/utils/CoppTestUtils.cpp
@@ -1286,11 +1286,14 @@ std::vector<cfg::PacketRxReasonToQueue> getCoppRxReasonToQueuesForBcm(
           std::pair(cfg::PacketRxReason::ARP, coppHighPriQueueId),
           std::pair(cfg::PacketRxReason::DHCP, coppMidPriQueueId),
           std::pair(cfg::PacketRxReason::BPDU, coppMidPriQueueId),
-          std::pair(cfg::PacketRxReason::L3_MTU_ERROR, kCoppLowPriQueueId),
           std::pair(cfg::PacketRxReason::L3_SLOW_PATH, kCoppLowPriQueueId),
           std::pair(cfg::PacketRxReason::L3_DEST_MISS, kCoppLowPriQueueId),
           std::pair(cfg::PacketRxReason::TTL_1, kCoppLowPriQueueId),
           std::pair(cfg::PacketRxReason::CPU_IS_NHOP, kCoppLowPriQueueId)};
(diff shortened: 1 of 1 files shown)


ashutoshgrewal:
Hey team, the SaiHostifManager changes need to be reworked. The FBOSS pattern is to exclude unsupported traps at config generation time, not skip them silently in the SAI layer. If an unsupported trap reaches SaiHostifManager, that means something upstream built the config wrong, and we want that to fail loudly rather than get swallowed.

  Look at how getCoppRxReasonToQueuesForSai() handles this - EAPOL, MPLS_TTL_1, SAMPLEPACKET, PORT_MTU_ERROR, HOST_MISS are all gated with hwAsic->isSupported() so they never enter switch state if the ASIC can't handle them. Your CoppTestUtils  change does exactly this for the BCM path, which is the right approach.

  For the next version:
  - Keep the isSupported check in getCoppRxReasonToQueuesForBcm() but delete the commented-out line entirely (no // std::pair(...))
  - Remove all SaiHostifManager.cpp changes
  - Remove the SaiApiError.h include (unused)

jenna-kuo:
Hi @ashutoshgrewal,

Thanks for the feedback. Reworked the PR following the architecture principle.

- Reverted all `SaiHostifManager.cpp` changes — the silent-skip blocks in `processRxReasonToQueueDelta()` and the `SaiApiError.h` include are gone.
- Kept the `getCoppRxReasonToQueuesForBcm()` `isSupported()` gate.
- Deleted the `// std::pair(cfg::PacketRxReason::L3_MTU_ERROR, ...)` commented-out line entirely.
- Updated the PR title and description to reflect the narrowed scope.

On the production COPP config path: `getCoppRxReasonToQueuesForSai()` already gates `L3_MTU_ERROR` via `isSupported(L3_MTU_ERROR_TRAP)`. We also found that link-test materialized JSON configs bypass that builder and feed `rxReasonToQueueOrderedList` directly into switch state — for our minipack3bta case we removed the `L3_MTU_ERROR` entry in the link config (PR #1171). We will keep an eye on any other JSON config sources during bring-up and gate at the source the same way.

Pushed the update.

Thank you.

jenna-kuo:
Merged internally. Closing. Thank you.

PR #1173, opened 2026-05-07
[Cisco][Q200, G200 and G202X] Fix AgentQueuePerHostTests for Cisco Asics

ram316583:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Fix AgentQueuePerHostTests for Cisco Asics

# Test Plan

Manually verified the test - AgentQueuePerHostTest/2.VerifyHostToQueueMappingClassIDsAfterResolve

### DIFF
--- a/fboss/agent/test/agent_hw_tests/AgentQueuePerHostTests.cpp
+++ b/fboss/agent/test/agent_hw_tests/AgentQueuePerHostTests.cpp
@@ -305,8 +305,23 @@ class AgentQueuePerHostTest : public AgentHwTest {
           if (qid == 0) {
             EXPECT_EVENTUALLY_GE(pktsOnQueue, 1);
           } else {
-            EXPECT_EVENTUALLY_EQ(
-                pktsOnQueue, 2 /* 1 pkt each for ttl < 128 and ttl >= 128 */);
+            /*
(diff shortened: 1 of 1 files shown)


Tianyu-Meta (approved):

vvasavada-nexthop:
https://github.com/facebook/fboss/pull/1448 does the same fix. Given that we have a few places where we have same checks for ASICs and expected pktCount, it is better to put this into a helper function like PR 1448 does and use it everywhere (here and in other tests).

ram316583:
> #1448 does the same fix. Given that we have a few places where we have same checks for ASICs and expected pktCount, it is better to put this into a helper function like PR 1448 does and use it everywhere (here and in other tests).

Sure , Makes sense.  Let me know when #1448 PR gets merged, I can close this PR. Thanks for Info.

ram316583:
The main reason we currently expect 4 packets for Cisco ASICs is due to packets being looped back on the egress side. This is an unintended side effect of the loopbacks in the test setup and is causing test flakiness.

We are planning to drop these looped-back packets on the egress port using an ACL. Once that is in place, the expected packet count should be 2 for all ASICs.

Closing this PR as it won't be needed when we add acl changes

PR #1178, opened 2026-05-11
[Nexthop][fboss2-dev] Add fboss2-dev config mac aging-time CLI command

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Adds `fboss2-dev config mac aging-time <seconds>` to set the L2 MAC
address aging timer (`l2AgeTimerSeconds` in `SwitchSettings`, default
300s).

The change is applied **HITLESSLY** by the SAI layer
(`SaiSwitch.cpp::setMacAgingSeconds()`, no `ChangeProhibited` guard).

**New files:**
- `commands/config/mac/CmdConfigMac.{h,cpp}` — parent branch node
- `commands/config/mac/aging_time/CmdConfigMacAgingTime.{h,cpp}` — leaf handler with positive-integer validation
- `test/config/CmdConfigMacAgingTimeTest.cpp` — unit tests (arg validation, set, already-set)
- `test/integration_test/ConfigMacTest.cpp` — E2E set/restore test

**Modified:** `CmdListConfig.cpp`, `cmake/CliFboss2.cmake`, `cmake/CliFboss2TestConfig.cmake`, `cmake/CliFboss2TestIntegrationTest.cmake`, `fboss2/BUCK`, `test/config/BUCK`, `test/integration_test/BUCK`

# Test Plan

Unit tests — 3 test cases covering arg validation, set, and already-set no-op:

```
$ bazel test //fboss/cli/fboss2/test/config:cmd_config_test \
    --test_filter='CmdConfigMacAgingTimeTestFixture.*' --test_output=all
...
[  PASSED  ] 3 tests.
```

Integration test — `ConfigMacTest.SetAndRestoreAgingTime` on a DUT (set aging-time, verify, restore).


fboss/cli/fboss2/commands/config/mac/aging_time/CmdConfigMacAgingTime.h
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+ private:
+  int32_t seconds_{300};
+};
+
+struct CmdConfigMacAgingTimeTraits : public WriteCommandTraits {
+  using ParentCmd = CmdConfigMac;
+  static void addCliArg(CLI::App& cmd, std::vector<std::string>& args) {
+    cmd.add_option("seconds", args, "MAC aging time in seconds (1-1000000)");
```
joseph5wu, line 34: Problem: The CLI help string says "MAC aging time in seconds (1-1000000)" but the validation in MacAgingTimeArg only checks seconds_ <= 0 — values above 1,000,000 are silently accepted. This was flagged by both AI reviewers.

  Fix: Remove the range from the help string since SAI doesn't impose a hard upper limit:
```
  // Before
  cmd.add_option("seconds", args, "MAC aging time in seconds (1-1000000)");

  // After
  cmd.add_option("seconds", args, "MAC aging time in seconds");
```

  Rule of thumb: If you advertise a range in the help text, enforce it in validation. If you don't want to enforce an upper bound, don't document one.

fboss/cli/fboss2/commands/config/mac/aging_time/CmdConfigMacAgingTime.cpp
```diff
@@ -0,0 +1,77 @@
...
+  swConfig.switchSettings()->l2AgeTimerSeconds() = newSeconds;
+
+  session.saveConfig(cli::ServiceType::AGENT, cli::ConfigActionLevel::HITLESS);
+
+  return fmt::format(
+      "Successfully set MAC aging time to {} seconds", newSeconds);
+}
+
+void CmdConfigMacAgingTime::printOutput(const RetType& output) {
+  std::cout << output << std::endl;
```
joseph5wu, line 70: Problem: CmdConfigMacAgingTime.cpp::printOutput() uses std::endl, while other config commands (e.g., CmdConfigPtpTransparentClock) use "\n". std::endl forces a flush which is unnecessary here.

  Fix:
```
  // Before
  std::cout << output << std::endl;

  // After
  std::cout << output << "\n";
```

fboss/cli/fboss2/test/integration_test/ConfigMacTest.cpp
```diff
@@ -0,0 +1,97 @@
...
+  } else {
+    XLOG(INFO) << "[Step 2] Setting aging-time to minimum (1s)...";
+    setAgingTime(1);
+    EXPECT_EQ(getAgingTime(), 1);
+
+    XLOG(INFO) << "[Step 3] Restoring original aging-time " << originalSeconds
+               << "s...";
+    setAgingTime(originalSeconds);
+    EXPECT_EQ(getAgingTime(), originalSeconds);
+  }
```
joseph5wu, line 94: Problem: SetMinimumAgingTime has a branch where originalSeconds == 1 that logs "Already at minimum — skipping set step" and exits without executing any assertions. The test reports PASSED without testing anything, hiding potential
  regressions.

  Fix: Always exercise a real write by picking a different target value:
```
  // Before
  if (originalSeconds == 1) {
      XLOG(INFO) << "  Already at minimum — skipping set step.";
  } else {
      setAgingTime(1);
      EXPECT_EQ(getAgingTime(), 1);
      ...
  }

  // After
  int64_t targetSeconds = (originalSeconds == 1) ? 2 : 1;
  setAgingTime(targetSeconds);
  EXPECT_EQ(getAgingTime(), targetSeconds);
  setAgingTime(originalSeconds);
  EXPECT_EQ(getAgingTime(), originalSeconds);
```

  Rule of thumb: Every test path should execute at least one assertion. If a test might skip its core logic, either use GTEST_SKIP() to make the skip explicit, or restructure so a real write is always exercised (see
  SetAndRestoreAgingTime in the same file for the correct pattern).

joseph5wu (commented):
I've addressed all the comments internally but wants to leave them here for future reference.
Please consider these comments for your future PRs

fboss/cli/fboss2/commands/config/mac/aging_time/CmdConfigMacAgingTime.h
vybhav-nexthop: thanks for this insight, didnt wanna enforce a validation since wasn't sure what the upper limit should be

PR #1179, opened 2026-05-11
[Nexthop][fboss2-dev] fboss2-dev switchport access vlan: validate VLAN exists before commit

hillol-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

`fboss2-dev config interface <name> switchport access vlan <id>` blindly wrote `ingressVlan` on the port and rewrote the matching `vlanPorts` entry without checking that the target VLAN existed in `config.sw()->vlans()`. The CLI then printed `Config session committed successfully` and triggered a warmboot; `fboss_sw_agent` SIGABRTed during initialization with:

```
switch initialization failed: VLAN <id> has no interface, even when corresp port <port> is enabled
```

Validate that the VLAN exists up front in `queryClient()` and throw `std::invalid_argument` with a descriptive message before touching the config or saving.

- `CmdConfigInterfaceSwitchportAccessVlan.cpp`: pre-flight check for VLAN existence.
- `CmdConfigInterfaceSwitchportAccessVlanTest.cpp`: fixture now seeds `sw.vlans`; added a negative test for the rejection path and assert the running config is unchanged after the throw.

Note: the redundant `ConfigSessionRestartTest` removal that previously rode along with this PR has been split out into #1202.

## Test plan

- [x] Unit tests: `queryClientThrowsWhenVlanDoesNotExist`, plus updated positive case still passes against the seeded VLAN.


fboss/cli/fboss2/commands/config/interface/switchport/access/vlan/CmdConfigInterfaceSwitchportAccessVlan.cpp
```diff
@@ -31,6 +31,36 @@ CmdConfigInterfaceSwitchportAccessVlan::queryClient(
...
+      if (*intf.vlanID() == vlanId) {
+        vlanHasInterface = true;
+        break;
+      }
+    }
+    if (!vlanHasInterface) {
+      throw std::invalid_argument(
+          "VLAN " + std::to_string(vlanId) +
+          " has no interface. Add an interface to the VLAN before assigning ports to it.");
+    }
```
joseph5wu: I'm a bit confused by the limitation here? Why do we need such vlan has an interface before you can assign a port/interface to it?

joseph5wu (commented):
The vlan existence check looks good to me but I'm not sure why do we need the vlan to have interfaces before you can set a port with an access vlan? Wouldn't this cli be used to assign the vlan to a port/interface?

fboss/cli/fboss2/commands/config/interface/switchport/access/vlan/CmdConfigInterfaceSwitchportAccessVlan.cpp
hillol-nexthop: You are right @joseph5wu . I have seen a VLAN with no interface crashes the agents so took this path. But I realise it is redundant as this crash will be flagged during vlan creation itself. No point in dragging this check here.

hillol-nexthop:
> The vlan existence check looks good to me but I'm not sure why do we need the vlan to have interfaces before you can set a port with an access vlan? Wouldn't this cli be used to assign the vlan to a port/interface?

Yes you are right. I have changed it.

PR #1181, opened 2026-05-12
[Accton][minipack3bta] Add --extra-gflags pass-through to run_test.py

jeankuo:
**Pre-submission checklist**
- [v] `pre-commit run`

# Summary

Add --extra-gflags argument to run_test.py for forwarding arbitrary gflags to test binaries without modifying run_test.py itself. This enables passing runtime flags (e.g., --asic_feature_support_overrides) directly to binaries.

Use = to attach value when it starts with --:
  --extra-gflags="--asic_feature_support_overrides=73=false --v=2"

Changes:
- Add _get_common_gflags() method in TestRunner base class
- Inject common gflags into _get_test_run_cmd() for all test runners
- Add --extra-gflags argument to the argument parser

# Test Plan

- Run run_test.py sai without --extra-gflags: no behavioral change
- Run with --extra-gflags="--v=2": confirm flag appears in binary command
- Run with --extra-gflags="--asic_feature_support_overrides=73=false": confirm override flag forwarded correctly
- Verify --extra-gflags can be placed at any position in the command line

### DIFF
--- a/fboss/oss/scripts/run_scripts/run_test.py
+++ b/fboss/oss/scripts/run_scripts/run_test.py
@@ -266,6 +266,15 @@ class TestRunner(abc.ABC):
         self._known_bad_test_regexes = None
         self._unsupported_test_regexes = None
 
+    def _get_common_gflags(self) -> list[str]:
+        """
+        Return pass-through gflags appended to every test binary invocation.
+        No parsing or validation - just forward whatever the user provides.
+        """
+        if hasattr(args, "extra_gflags") and args.extra_gflags:
+            return args.extra_gflags.split()
+        return []
+
     @abc.abstractmethod
     def _get_config_path(self):
         pass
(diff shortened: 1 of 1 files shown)


ashutoshgrewal:
Hi team, This is tied to the other PR which should not land. If you have a separate use case for --extra-gflags beyond the runtime override, please let us know.

jenna-kuo:
> Hi team, This is tied to the other PR which should not land. If you have a separate use case for --extra-gflags beyond the runtime override, please let us know.

Hi @ashutoshgrewal ,

Thanks for the question.
We agree with your feedback on #1182 and will close that PR.

For this PR, we think --extra-gflags can be useful on its own. In the past, when we needed to pass specific gflags through run_test.py (for example --platform_mapping_override_path and --bsp_platform_mapping_override_path), we asked the team and Meta added them one by one. Looking at the git history, the platform_mapping flags alone were added in three separate commits across different subcommands (c93e65ef, 4d3c074b, 0ca1e8a0). --extra-gflags is a generic version of the same idea, so similar requests in the future do not need a new wrapper each time.

The change only affects run_test.py and does not touch any production code. If you are open to it, we would like to land this PR. We can update the description to remove the reference to #1182 and explain the standalone purpose. If you prefer to close it, we are also OK with that. Please let us know.

Thank you.

PR #1182, opened 2026-05-12
[Accton][minipack3bta] Add runtime asic feature support override via gflag

jeankuo:
**Pre-submission checklist**
- [v] `pre-commit run`

# Summary

Add runtime feature support override mechanism via gflag --asic_feature_support_overrides for TomahawkUltra1 ASIC.

This allows overriding individual HwAsic::Feature support at runtime without code changes, enabling faster platform bring-up and SDK compatibility testing. The override is specified as comma-separated FEATURE_ID=true|false pairs.

Changes:
- Add getFeatureSupportOverride() protected method in HwAsic base class
- Add logFeatureSupportOverrides() to log active overrides at startup
- Integrate override check in TomahawkUltra1Asic::isSupported()

# Test Plan

- Build and run SAI agent with no flag: no behavioral change, no log output
- Run with --asic_feature_support_overrides 73=false: feature reported as unsupported and FeatureOverride log lines appear at startup
- Run with invalid/empty values: graceful handling, no crash
- Verify removing the flag restores original feature support behavior

[test_log.txt](https://github.com/user-attachments/files/27634398/test_log.txt)

### DIFF
--- a/fboss/agent/hw/switch_asics/HwAsic.cpp
+++ b/fboss/agent/hw/switch_asics/HwAsic.cpp
@@ -8,7 +8,12 @@
  *
  */
 #include "fboss/agent/hw/switch_asics/HwAsic.h"
+#include <folly/logging/xlog.h>
 #include <thrift/lib/cpp/util/EnumUtils.h>
+#include <algorithm>
+#include <cctype>
+#include <cstdlib>
+#include <sstream>
 #include "fboss/agent/FbossError.h"
 #include "fboss/agent/hw/switch_asics/Agera3PhyAsic.h"
 #include "fboss/agent/hw/switch_asics/Chenab2Asic.h"
@@ -35,6 +40,10 @@
 
 DEFINE_int32(acl_gid, -1, "Content aware processor group ID for ACLs");
 DEFINE_int32(teFlow_gid, -1, "Exact Match group ID for TeFlows");
+DEFINE_string(
+    asic_feature_support_overrides,
+    "",
+    "Comma-separated feature overrides in the format FEATURE_ID=true|false");
 
 namespace {
 constexpr auto kDefaultACLGroupID = 128;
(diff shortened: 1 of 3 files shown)


daiwei1983:
Can you tell me more about the use case, when you might want to dynamically turn on/off hw features at run time pretty often?

Ideally, these HwAsic features just tell whether the ASIC is capably of supporting this feature from hardware point of view. Whether this feature is really enabled or not is determined by separate feature gflag. So, HwAsic feature support is more like static hardware spec information never changes. It makes more sense to dynamically guard feature on/off using gflag.

jenna-kuo:
> Add runtime feature support override mechanism via gflag --asic_feature_support_overrides for TomahawkUltra1 ASIC.
> 
> This allows overriding individual HwAsic::Feature support at runtime without code changes, enabling faster platform bring-up and SDK compatibility testing. The override is specified as comma-separated FEATURE_ID=true|false pairs.

Hi  @daiwei1983,

Thanks for the review! Let me clarify the use case.
 
You're right that HwAsic::isSupported() represents static hardware capability and shouldn't change in normal operation. This override is not intended for production feature toggling.
 
The use case is during new platform bring-up and SDK compatibility testing. When we bring up a new ASIC/SDK version (e.g., TomahawkUltra1 with a newer Memory SDK), we sometimes encounter situations where:
 
1. A feature is declared as supported in the ASIC class, but the SDK version we're testing against doesn't actually implement it yet (or has a regression).
2. We need to quickly verify whether a failure is caused by a specific feature being incorrectly reported as supported/unsupported, without rebuilding the binary.

Rebuilding and redeploying the agent binary on lab switches takes significant time. This gflag lets us temporarily mask a feature to isolate issues during debugging, without code changes or recompilation.
 
Also worth noting: currently the override is only integrated into TomahawkUltra1Asic::isSupported() — it does not affect any other ASIC classes. The scope is intentionally limited to the platform we're actively bringing up, minimizing any risk to existing platforms.

Once bring-up is complete and the correct feature set is confirmed, the flag is not used — it defaults to empty string (no-op). The WARNING-level logging also makes it very visible when overrides are active, so it won't silently affect production behavior.

ashutoshgrewal:
Hi team,

 This is fine as a local development hack for your private builds during bring-up - we understand the need to iterate quickly when discovering what TU1 supports. But it shouldn't land in the codebase. FBOSS requires ASIC feature support to be hardcoded and reviewed - it's the contract the rest of the system relies on to build correct configs and program hardware safely.

  The right approach for landing changes is what you're already doing in other PRs - update the return true / return false blocks in TomahawkUltra1Asic::isSupported() directly.

jenna-kuo:
> Hi team,
> 
> This is fine as a local development hack for your private builds during bring-up - we understand the need to iterate quickly when discovering what TU1 supports. But it shouldn't land in the codebase. FBOSS requires ASIC feature support to be hardcoded and reviewed - it's the contract the rest of the system relies on to build correct configs and program hardware safely.
> 
> The right approach for landing changes is what you're already doing in other PRs - update the return true / return false blocks in TomahawkUltra1Asic::isSupported() directly.

Hi @ashutoshgrewal ,
Thanks for the clear feedback, this makes sense. The contract argument is fair — we'll keep this patch local for our bring-up work and stick with the hardcoded isSupported() approach for upstream changes (as in our other PRs). Closing this PR.

jenna-kuo:
Note: This branch has been rebased to latest upstream/main and includes an additional utility script (parse_feature_enum.py) for looking up feature IDs. The GitHub diff above may not reflect the latest state. To get the full commit (4 files), use: git fetch https://github.com/jeankuo/fboss.git dev-20260511 && git cherry-pick FETCH_HEAD

PR #1183, opened 2026-05-12
[Nexthop][fboss2-dev] CLI for config and delete ACL rule and action subcommands

hillol-nexthop:
## Summary

Full ACL rule CLI surface in `fboss2-dev` — match fields, actions, and whole-rule delete:

```
fboss2-dev config acl rule <table> <rule> <attr> <value> [<ttl-mask>]
fboss2-dev config acl rule <table> <rule> action <subattr> [<value>]
fboss2-dev delete acl rule <table> <rule>
```

`config acl rule` upserts — if the named rule does not exist in the table it is created with the supplied attribute, otherwise the attribute is applied to the existing entry. `delete acl rule` removes the entire `AclEntry` (and its `MatchAction`) from the table.

### Match-field attributes

16 setters on `AclEntry`: `source-ip`, `destination-ip`, `protocol`, `source-port`, `destination-port`, `dscp`, `tcp-flags`, `icmp-type`, `icmp-code`, `ip-fragment`, `ttl`, `destination-mac`, `ethertype`, `vlan`, `ip-type`, `packet-lookup-result`. The 5-token `ttl <value> <mask>` form is unique to the setter; without `<mask>` it defaults to `0xFF`.

### Action subcommands

```
action permit | deny                            # AclEntry.actionType
action send-to-queue <0-32767>                  # MatchAction.sendToQueue
action set-dscp <0-63>                          # MatchAction.setDscp
action set-tc <0-7>                             # MatchAction.setTc
action mirror-ingress <name>                    # MatchAction.ingressMirror
action mirror-egress <name>                     # MatchAction.egressMirror
action counter <name>                           # MatchAction.counter
action trap-to-cpu                              # MatchAction.toCpuAction = TRAP
action copy-to-cpu                              # MatchAction.toCpuAction = COPY
action redirect nexthop <ip>                    # MatchAction.redirectToNextHop
```

`permit` / `deny` mutate `AclEntry.actionType` in place. The rest land on a `MatchAction` attached via `dataPlaneTrafficPolicy.matchToAction` keyed by rule name; the CLI finds or creates the `MatchToAction` entry on demand.

### Common

All operations target `sw.aclTableGroups` (field 56) — the deprecated `aclTableGroup` (field 45) is not supported. All are **HITLESS**: AclEntry + MatchAction mutations apply at runtime via `processAclTableGroupDelta` in `SaiAclTableManager`; no warmboot-prohibited guard in `SaiSwitch`.

## How to use the CLIs

```text
# Create + set a match field (rule auto-created if absent)
$ fboss2-dev config acl rule AclTable1 my-rule source-ip 10.250.0.0/24
Created and set acl rule 'my-rule' (table 'AclTable1', group 'acl-table-group-ingress') source-ip to 10.250.0.0/24

# Set an action on the same rule
$ fboss2-dev config acl rule AclTable1 my-rule action set-dscp 46
Set acl rule 'my-rule' (table 'AclTable1', group 'acl-table-group-ingress') action to set-dscp

# Redirect form (6-token)
$ fboss2-dev config acl rule AclTable1 my-rule action redirect nexthop 10.10.10.1
Set acl rule 'my-rule' (table 'AclTable1', group 'acl-table-group-ingress') action to redirect

# Remove the whole rule
$ fboss2-dev delete acl rule AclTable1 my-rule
Deleted acl rule 'my-rule' from table 'AclTable1' (group 'acl-table-group-ingress')

# Inspect + commit
$ fboss2-dev config session diff       # shows entry + matchToAction add/remove
$ fboss2-dev config session commit     # HITLESS — no agent restart
```

### Unit tests

`fboss2_cmd_config_test --gtest_filter='*AclRule*'` → **49/49 PASS**

`CmdConfigAclRuleTestFixture` (44):
- Match-field arg validation: `argValidation_badArity`, `argValidation_unknownAttr`, `argValidation_extraTokenOnlyForTtl`, `argValidation_outOfRange`, `argValidation_badIp`, `argValidation_badMac`, `argValidation_protocolKeyword`, `argValidation_ipFragmentEnum`, `argValidation_etherTypeNumericOrName`
- Action arg validation: `argValidation_unknownActionSubattr`, `argValidation_actionPermitNoExtra`, `argValidation_actionRequiresValue`, `argValidation_actionRangeChecks`, `argValidation_actionRedirectShape`
- Match-field mutation: `ruleAutoCreated`, `tableNotFound`, `setSourceIp`, `setDestinationIp`, `setProtocol`, `setSourcePort`, `setDestinationPort`, `setDscp`, `setTcpFlags`, `setIcmpType`, `setIcmpCode`, `setIpFragment`, `setTtlDefaultMask`, `setTtlExplicitMask`, `setDestinationMac`, `setEtherType`, `setVlan`, `setIpType`, `setPacketLookupResult`
- Action mutation: `setActionPermit`, `setActionDeny`, `setActionSendToQueue`, `setActionSetDscp`, `setActionSetTc`, `setActionMirrorIngress`, `setActionMirrorEgress`, `setActionCounter`, `setActionTrapToCpu`, `setActionCopyToCpu`, `setActionRedirectNexthop`

`CmdDeleteAclRuleTestFixture` (5):
- `argValidation_badArity`, `argValidation_validArity`, `tableNotFound`, `ruleNotFound`, `deleteRule`

### Integration tests

`fboss2_integration_test --gtest_filter='*AclRule*'` on a real DUT → **25/25 PASS**

`SetUpTestSuite` snapshots `/etc/coop/agent.conf` and unions the first AclTable's `qualifiers` list with every match-field qualifier exercised by the suite, then restarts `fboss_sw_agent`. `TearDownTestSuite` restores the snapshot and restarts. Each test creates the rule on demand via `config acl rule`; per-test `TearDown` deletes it via `delete acl rule`.

Match-field (17):
- `SetSourceIp`, `SetDestinationIp`, `SetProtocol`, `SetSourcePort`, `SetDestinationPort`, `SetDscp`, `SetTcpFlags`, `SetIcmpType`, `SetIcmpCode`, `SetIpFragment`, `SetTtlDefaultMask`, `SetTtlExplicitMask`, `SetDestinationMac`, `SetEtherType`, `SetVlan`, `SetIpType`, `SetPacketLookupResult`

Action (7):
- `SetActionPermit`, `SetActionDeny`, `SetActionSendToQueue`, `SetActionSetDscp`, `SetActionSetTc`, `SetActionTrapToCpu`, `SetActionCopyToCpu`

Delete (1):
- `DeleteRule` — creates the entry via `config acl rule`, commits, asserts present, then `delete acl rule`, commits, asserts gone.

`action mirror-ingress`, `action mirror-egress`, `action counter`, and `action redirect nexthop` are intentionally unit-test-only — SAI rejects entries that reference an undefined mirror session, unprovisioned counter, or unresolvable nexthop, and provisioning that supporting state on the DUT is out of scope for this PR. The CLI is verified to construct the right config delta in unit tests.


fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
```diff
@@ -0,0 +1,544 @@
...
+  }
+  if (s == "udp") {
+    return 17;
+  }
+  if (s == "icmp") {
+    return 1;
+  }
+  if (s == "icmpv6") {
+    return 58;
+  }
```
joseph5wu: What are these magic numbers?

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
```diff
@@ -0,0 +1,544 @@
...
+      } else if (
+          sub == kAclRuleActionMirrorIngress ||
+          sub == kAclRuleActionMirrorEgress || sub == kAclRuleActionCounter) {
+        if (av.empty()) {
+          throw std::invalid_argument(
+              fmt::format("Action '{}' requires a non-empty name", sub));
+        }
+        actionStringValue_ = av;
+      }
+    }
```
joseph5wu: Can we make this huge action related code block as a separate function?

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
```diff
@@ -0,0 +1,544 @@
...
+    intValue_ = parseIntInRange(attr, value, 0, 255);
+  } else if (attr == kAclRuleAttrIcmpType || attr == kAclRuleAttrIcmpCode) {
+    intValue_ = parseIntInRange(attr, value, 0, 255);
+  } else if (attr == kAclRuleAttrIpFragment) {
+    ipFrag_ = parseIpFragment(value);
+  } else if (attr == kAclRuleAttrTtl) {
+    intValue_ = parseIntInRange(attr, value, 0, 255);
+    if (v.size() == 5) {
+      ttlMask_ =
+          static_cast<int16_t>(parseIntInRange("ttl-mask", v[4], 0, 255));
```
joseph5wu: Please consider to use constants for all these magic numbers for the max values of each type.

fboss/cli/fboss2/test/config/BUCK
```diff
@@ -5,7 +5,8 @@ oncall("fboss_oss")
 cpp_unittest(
     name = "cmd_config_test",
     srcs = [
-        "BgpConfigSessionTest.cpp",
+        "BgpConfigSessionTest.cpp",  # bazelify: exclude
+        "CmdConfigAclRuleTest.cpp",
```
joseph5wu, line 10: Please don't include `# bazelify: exclude` as we don't use Nexthop implemented bazelify.sh in FBOSS

fboss/cli/fboss2/test/integration_test/ConfigAclRuleTest.cpp
```diff
@@ -0,0 +1,594 @@
...
+      17, // IP_TYPE
+      18, // TTL
+      22, // ETHER_TYPE
+      23, // OUTER_VLAN
+  };
+  return kIds;
+}
+} // namespace
+
+class ConfigAclRuleTest : public Fboss2IntegrationTest {
```
joseph5wu, line 115: Recently I introduced this per-test agent cold boot to Fboss2IntegrationTestRunner 
https://github.com/facebook/fboss/commit/b108a8750dc01f1bb3075eb75a60bd70824f5e8a
to make sure each single fboss2_integration_test run will alwasy start with a clean state.
And in our FBOSS internal test tool(Netcastle), I also implemented something similar

The goal is to keep our gtest here only focus on testing the functionality of our services/fboss2,
while our test infrastructure/orchestrator(like Netcastle or run_test.py) can deal with the setup/cleanup test suite logic. Please consider to simplify the test to use what we provide in run_test.py

joseph5wu (changes requested):

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
hillol-nexthop: changed to standard constants.

fboss/cli/fboss2/test/config/BUCK
hillol-nexthop: done

fboss/cli/fboss2/test/integration_test/ConfigAclRuleTest.cpp
hillol-nexthop: I have removed the snapshotting part but I need to keep the residue for injecting ACL qualifiers. I know run_test.py accepts custom config for a test but when I want to run all the integration tests, I can't specify custom config for a particular suite. The current method takes whatever base config you provide and adds the qualifiers to facilitate ACL tests to run which I find much more convenient.
It deviates a bit from the convention but makes the test self-sustainable. Let me know if you are okay with the tradeoff.

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
hillol-nexthop: refactored.

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/acl/rule/AclRuleAttrs.cpp
```diff
@@ -0,0 +1,137 @@
...
+  if (sub == kAclRuleActionPermit || sub == kAclRuleActionDeny ||
+      sub == kAclRuleActionTrapToCpu || sub == kAclRuleActionCopyToCpu) {
+    return {4, 4}; // no value
+  }
+  if (sub == kAclRuleActionRedirect) {
+    return {6, 6}; // redirect nexthop <ip>
+  }
+  return {5, 5}; // single value
+}
+
```
joseph5wu: Let's define all these magic number as constants in the header.
I noticed you also directly use these numbers in the CmdConfigAclRule.cpp which makes me concern if we need to change the syntax of one of the attributes, it might not be as easy to locate the places we need to adjust

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
```diff
@@ -0,0 +1,611 @@
...
+  if (attr == kAclRuleAttrTtl) {
+    throw std::invalid_argument(
+        fmt::format(
+            "Attribute 'ttl' takes a value and an optional mask "
+            "('ttl <value> [<mask>]'); got {} value token(s)",
+            v.size() - 3));
+  }
+  // Everything else is exactly one value; a 5th token is only meaningful for
+  // `ttl`, so call that out specifically.
+  if (v.size() == 5) {
```
joseph5wu: This is why we need to define these magic numbers.

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
```diff
@@ -0,0 +1,611 @@
...
+  const auto arity = aclRuleActionArity(sub);
+  if (v.size() >= arity.min && v.size() <= arity.max) {
+    return;
+  }
+  // Phrase the error in terms of what's wrong: too few tokens means a missing
+  // value, too many means trailing junk. v[0..2] are table/rule/attr, so the
+  // value-token count is v.size() - 4 ('action' itself is v[3]).
+  const bool tooFew = v.size() < arity.min;
+  // Re-derive the action category from its arity to phrase the error.
+  if (arity.max == 4) {
```
joseph5wu: Magic numbers!!!

fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
```diff
@@ -475,10 +472,61 @@ void Fboss2IntegrationTest::waitForAgentReady(
     // NOLINTNEXTLINE(facebook-hte-BadCall-sleep_for)
     std::this_thread::sleep_for(std::chrono::seconds(10));
   }
+  dumpAgentDiagnostics();
   FAIL() << "Agent did not become ready within " << timeout.count()
          << " seconds";
 }
 
+void Fboss2IntegrationTest::dumpAgentDiagnostics() const {
```
joseph5wu: I highly recommend not to include this logic in the integration tests.
Instead we should consider to use the test orchestrator like run_test.py or our internal Netcastle 
to handle the agent services or collecting logs since these should be the test orchestrator's responsibility.

Let's remove this logic out from this PR and have this PR only focus on ACL related implementation

joseph5wu (changes requested):
Please address the magic number and acl unrelated test services orchestration logic comments.

fboss/cli/fboss2/commands/config/acl/rule/AclRuleAttrs.cpp
hillol-nexthop: I have taken most of the numbers to constants. Some of them are still left and looks to me trivial to use constants like:         ->expected(4, 6).
Also I have changed the design as I felt a lot of if-else was done in parsing. I changed it to a more table-lookup based approach.

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
hillol-nexthop: removed

fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
hillol-nexthop: Removed this.

PR #1198, opened 2026-05-14
[Nexthop][NPI] "asic_config_v3" part 1 - basic skeleton

marif-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Implements `asic_config_v3` design. 

Add `fboss/lib/asic_config_v3/` Python package skeleton along with a fully defined CMake build target. This creates the basic infra of the data-driven ASIC config generator (`asic_config_v3`) that will replace the code-driven `asic_config_v2` path.

The `gen.py` entry point is included with an empty `_GENERATOR_REGISTRY` so the `fboss-asic-config-v3-gen` target is buildable immediately; registry entries for concrete ASIC families will be added in follow-up commits.

# Test Plan
Ran `fboss/lib/asic_config_v3/run-helper.sh`.

Note: `fboss/lib/asic_config_v3/run-helper.sh` will not generate any output until "asic_config_v3" part 4 PR.

----

**This PR is created on top of**
- https://github.com/facebook/fboss/pull/1131
- https://github.com/facebook/fboss/pull/1149 

**so it contains diff from it. When that PR will merge, this PR will have a shorter diff.**

### DIFF
--- a/cmake/AsicConfigV3ConfigCli.cmake
+++ b/cmake/AsicConfigV3ConfigCli.cmake
@@ -1,0 +1,37 @@
+# CMake to build libraries and binaries in fboss/lib/asic_config_v3
+# Data-driven ASIC config generation
+
+include(FBPythonBinary)
+
+set(
+    ASIC_CONFIG_V3_PY_SRCS
+    "fboss/lib/asic_config_v3/__init__.py"
+    "fboss/lib/asic_config_v3/base_generator.py"
+    "fboss/lib/asic_config_v3/gen.py"
+    "fboss/lib/asic_config_v3/generators/__init__.py"
+    "fboss/lib/platform_mapping_v2/asic_vendor_config.py"
+    "fboss/lib/platform_mapping_v2/gen.py"
+    "fboss/lib/platform_mapping_v2/helpers.py"
+    "fboss/lib/platform_mapping_v2/platform_mapping_v2.py"
+    "fboss/lib/platform_mapping_v2/port_profile_mapping.py"
+    "fboss/lib/platform_mapping_v2/profile_settings.py"
+    "fboss/lib/platform_mapping_v2/read_files_utils.py"
+    "fboss/lib/platform_mapping_v2/si_settings.py"
+    "fboss/lib/platform_mapping_v2/static_mapping.py"
+)
+
+add_fb_thrift_python_executable(
+    fboss-asic-config-v3-gen
+    MAIN_MODULE fboss.lib.asic_config_v3.gen:generate_all_asic_configs
+    SOURCES ${ASIC_CONFIG_V3_PY_SRCS}
+    DEPENDS
+        platform_config_python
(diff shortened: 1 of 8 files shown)


cmake/AsicConfigV3ConfigCli.cmake
```diff
@@ -0,0 +1,38 @@
...
+    "fboss/lib/platform_mapping_v2/asic_vendor_config.py"
+    "fboss/lib/platform_mapping_v2/gen.py"
+    "fboss/lib/platform_mapping_v2/helpers.py"
+    "fboss/lib/platform_mapping_v2/integrated_transceiver_mapping.py"
+    "fboss/lib/platform_mapping_v2/platform_mapping_v2.py"
+    "fboss/lib/platform_mapping_v2/port_profile_mapping.py"
+    "fboss/lib/platform_mapping_v2/profile_settings.py"
+    "fboss/lib/platform_mapping_v2/read_files_utils.py"
+    "fboss/lib/platform_mapping_v2/si_settings.py"
+    "fboss/lib/platform_mapping_v2/static_mapping.py"
```
joseph5wu: Let's only include them when you actually need them
Your current PR doesn't need any `platform_mapping_v2` files.
I would like to decouple these two tools

cmake/AsicConfigV3ConfigCli.cmake
```diff
@@ -0,0 +1,38 @@
...
+    MAIN_MODULE fboss.lib.asic_config_v3.gen:generate_all_asic_configs
+    SOURCES ${ASIC_CONFIG_V3_PY_SRCS}
+    DEPENDS
+        platform_config_python
+        switch_config_python
+        transceiver_python
+        phy_python
+        platform_mapping_config_python
+        fboss_common_python
+        python-pyyaml::python-pyyaml
```
joseph5wu: Same comments for the dependency. asic config shouldn't use transceiver_python etc.
Please only uses the necessary dependency

joseph5wu (changes requested):
This base PR looks pretty straightforward to me.
But please make sure not to include unused files or dependencies

cmake/AsicConfigV3ConfigCli.cmake
marif-nexthop: Done. "Part 1" now declares only the skeleton's stdlib-only sources (__init__.py, base_generator.py, gen.py, generators/__init__.py) with no `platform_mapping_v2` files and no thrift DEPENDS. The `platform_mapping_v2` sources and the thrift dependencies will be added in Parts 3 and 4, in the same PRs as the code that actually imports them. Verified the `fboss-asic-config-v3-gen` target still builds and runs from a clean build.

cmake/AsicConfigV3ConfigCli.cmake
marif-nexthop: Done.

PR #1212, opened 2026-05-20
[Nexthop] Restrict stable_hashes tarball extraction to hashes and manifests only

benoit-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

The tarball at `fboss/oss/stable_commits/latest_stable_hashes.tar.gz` contains stale copies of `build/fbcode_builder/getdeps/*.py` and other build infrastructure files. Extracting it wholesale was silently overwriting the repo's current versions, including `build/fbcode_builder/getdeps/cli.py`, reverting local changes and fixes that might exist in the current branch / fork.

Scope the extraction to the two directories it is actually meant to update: `build/deps/github_hashes/` and `build/fbcode_builder/manifests/`.

Had to fix various ruff-related complaints in docker-build.py to keep pre-commit check happy.

# Test Plan

Not needed.

### DIFF
--- a/fboss/oss/scripts/docker-build.py
+++ b/fboss/oss/scripts/docker-build.py
@@ -9,10 +9,8 @@ import re
 import shlex
 import subprocess
 import sys
-import tempfile
 from pathlib import Path
-from typing import List, Optional, Tuple
-
+from typing import Optional
 
 OPT_ARG_SCRATCH_PATH = "--scratch-path"
 OPT_ARG_CMAKE_TARGET = "--target"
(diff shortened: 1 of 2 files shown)


KevinYakar:
overwriting the current build infra with what is contained in the stable commits is the intended behavior. the point of a stable commit is that the build should be guaranteed to pass the dependency build phase. if we only extract manifests and hashes, we can run into issues that fail the build, so we decided to pin both the hashes/manifests and the infra used to build it.

e.g. getdeps.py is updated to use a new variable defined in manifests, new getdeps.py expects it but pinned manifests are missing it, getdeps.py fails the build

benoit-nexthop:
Closing then. We'll keep this change as in our private fork, because reverting the Python files breaks our builds as it undoes some other of our changes.

PR #1232, opened 2026-05-25
[Nexthop] Fix fbthrift-python setup.py bdist_wheel build: sccache + parallel

benoit-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

The cmake custom command that invokes `setup.py bdist_wheel` was not forwarding the compiler launcher (`CMAKE_C/CXX_COMPILER_LAUNCHER=sccache`) to the subprocess, and set CXX but not CC even though Python's distutils on this system uses CC (gcc) for C++ extension files. The build also had no parallelism: all ~40+ Cython extensions compiled one-at-a-time without sccache, causing a significant build-time regression.

The patch to fbthrift's `thrift/lib/python/CMakeLists.txt`:
- Sets `CC` and `CXX` to `"$LAUNCHER $COMPILER"` for both the `--api-only` (Cython header generation) and `bdist_wheel` commands.
- Extracts the parallel job count from `CMAKE_JOB_POOLS` (compile=N), falling back to `CMAKE_BUILD_PARALLEL_LEVEL` or `nproc`, and passes `build_ext --parallel N` before `bdist_wheel` so setuptools compiles extensions in parallel with sccache.

Note: This patch could also be applied upstream in fbthrift directly as well.

# Test Plan

Code still builds. Albeit faster.

### DIFF
--- a/build/fbcode_builder/manifests/fbthrift-python
+++ b/build/fbcode_builder/manifests/fbthrift-python
@@ -16,6 +16,7 @@ fbthrift = thrift/lib/rust
 [build]
 builder = cmake
 job_weight_mib = 2048
+patchfile = fbthrift-python-sccache-launcher.patch
 
 [build.not(os=linux)]
 builder = nop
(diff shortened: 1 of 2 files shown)


AarjunC:
Thank you @benoit-nexthop. I pushed this change to fbthrift.

https://github.com/facebook/fbthrift/commit/09c056775c162ae6e91768810a31d811b3c6f282

PR #1233, opened 2026-05-26
[Nexthop] clean up ACL counter warmboot handles when recreating ACL table

andrew-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary
_UPDATE: This fix addresses a warm-boot ACL-table recreation bug. When an existing ACL table is found with attributes that differ from the newly requested attributes, the current path can trigger a crash. Existing warm-boot SAI tests no longer appear to trigger the condition naturally, but the problematic product code path remains. The issue was reproduced by forcing `FLAGS_force_recreate_acl_tables=true`; after reapplying this fix, the crash no longer reproduced._

During warmboot, when the desired ACL table create-attributes differ from what was recovered, the table is torn down and rebuilt. The old teardown path missed `SaiAclCounter` warmboot handles, which still reference the table OID — so `sai_remove_acl_table` returned `SAI_STATUS_OBJECT_IN_USE`. This was fatal on Cisco SAI, which surfaced on wedge800cact warmboot sai_agent tests.
- Adds the missing step: remove unclaimed warmboot handles from `SaiAclCounterStore` for the table being recreated.
- Replaces the old `setObject → reset()` with direct `removeUnclaimedWarmbootHandlesIf()` calls on each store, so the SAI `remove_*` APIs run in a controlled, dependency-respecting order (entries → counters → group member → table).
- Recreates the group member via `setObject` (normal `objects_` path) instead of `addWarmbootHandle`, matching that we're now in steady state for this table.

# Test Plan
wedge800cact before:
```
[       OK ] cold_boot.SaiPortAdminStateTest.assertAdminState (14221 ms)
[   FAILED ] warm_boot.SaiPortAdminStateTest.assertAdminState (0 ms)
```
after:
```
[       OK ] cold_boot.SaiPortAdminStateTest.assertAdminState (15745 ms)
[       OK ] warm_boot.SaiPortAdminStateTest.assertAdminState (7181 ms)
```

### DIFF
--- a/fboss/agent/hw/sai/switch/SaiAclTableManager.cpp
+++ b/fboss/agent/hw/sai/switch/SaiAclTableManager.cpp
@@ -2146,68 +2146,78 @@ void SaiAclTableManager::recreateAclTable(
   XLOG(DBG2) << "refreshing acl table schema";
   auto adapterHostKey = aclTable->adapterHostKey();
   auto& aclEntryStore = saiStore_->get<SaiAclEntryTraits>();
+  auto tableOid = static_cast<sai_object_id_t>(aclTable->adapterKey());
 
-  std::map<
-      SaiAclEntryTraits::AdapterHostKey,
-      SaiAclEntryTraits::CreateAttributes>
-      entries{};
-  // remove acl entries from acl table, retain their attributes
-  for (const auto& entry : aclEntryStore) {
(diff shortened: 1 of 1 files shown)


srikrishnagopu:
Closing PR.

PR #1090, opened 2026-04-16
[Celestica] Minerva: fan_service: Implement Minerva rack FSC new requ…

QiuyunXie:
…irement

**Pre-submission checklist**
- [ ] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [ ] `pre-commit run`
image

# Summary

Quanta thermal team and Meta thermal team have reached an agreement on the new requirements for Minerva rack FSC. Modify code and fan_service config file to implement the requirement as below:
image

image

# Test Plan

Quanta thermal team has verified this modification on Minerva rack and it meets their requirements.
image

Also attach the Fan_service unit test logs on IceCube, SantaBarbara and Anacapa25:
[Icecube_fan_service_log.txt](https://github.com/user-attachments/files/26785250/Icecube_fan_service_log.txt)
[Santa_fan_service_log.txt](https://github.com/user-attachments/files/26785258/Santa_fan_service_log.txt)
[Anacapa25_fan_service_log.txt](https://github.com/user-attachments/files/26785260/Anacapa25_fan_service_log.txt)

PR #1236, opened 2026-05-27
[Celestica] Minerva: fan_service: Implement and improve Minerva rack FSC new requirement

QiuyunXie:
**Pre-submission checklist**
- [ ] `pre-commit run`
image

# Summary

Quanta thermal team and Meta thermal team have reached an agreement on the new requirements for Minerva rack FSC. Modify code and fan_service config file to implement the requirement as below:
image
image
Following this PR [https://github.com/facebook/fboss/pull/1090] review comments, I reworked the "specialMode and COMeShutdown" into a config-driven deadFanShutdownCondition, where fan_service only implements a generic policy: when the number of failed fans reaches a configured threshold, optionally force a PWM and execute configured shutdown commands. Janga/Tahan opt into this behavior via their fan_service JSON configs; other platforms are unaffected unless they configure the condition.

# Test Plan

Quanta thermal team has verified this modification and it meets their requirements.
Attach the Fan_service unit test logs on Minerva and SantaBarbara:
[Minerva_FSC_log.txt](https://github.com/user-attachments/files/28308335/Minerva_FSC_log.txt)
[Santa_FSC_LOG.txt](https://github.com/user-attachments/files/28308340/Santa_FSC_LOG.txt)


fboss/platform/fan_service/ControlLogic.cpp
```diff
@@ -743,6 +771,11 @@ void ControlLogic::updateControl(std::shared_ptr<SensorData> pS) {
     fb303::fbData->setCounter(kHasFanPwmWriteFailure, hasAnyPwmFailure ? 1 : 0);
     fb303::fbData->setCounter(kHasFanAbsent, hasAnyFanAbsent ? 1 : 0);
   });
+
+  if (deadFanShutdownMode_) {
+    pBsp_->runFanDeadShutdownCmds(
```
mikechoifb, line 776: Why do we have two different guards on this (the first if statement here is one guard, and then the definition of runFanDeadShutdownCmds also has the enable boolean which is another guard, that is set to true). I suggest removing the enable parameter and then just using deadFanShutdownMode_ to gate this behavior.

fboss/platform/fan_service/ControlLogic.cpp
```diff
@@ -743,6 +771,11 @@ void ControlLogic::updateControl(std::shared_ptr<SensorData> pS) {
...
+        *config_.deadFanShutdownCondition()->fanDeadShutdownCmds(), true);
```
mikechoifb: There may be an unchecked access to an optional Thrift field. Make sure you check if the field has a value before dereferencing it or calling value(). Otherwise these will throw an exception if the field doesn't have a value. You can suppress this warning by wrapping the expression in apache::thrift::can_throw, e.g. can_throw(*obj.field()) or can_throw(obj.field()->func()) For more information, see: https://fburl.com/thrift-field-access#lint-for-optional-field-access this->config_.deadFanShutdownCondition is unchecked.

mikechoifb (changes requested):

fboss/platform/fan_service/ControlLogic.cpp
QiuyunXie: @mikechoifb Thanks for the comment. I have modified the code accordingly. Please take another look.

fboss/platform/fan_service/ControlLogic.cpp
QiuyunXie: @mikechoifb Thanks for the comment. I have modified the code accordingly. Please take another look.

PR #1237, opened 2026-05-27
[Meta] Add GitHub workflow for run_scripts pytest unit tests

AarjunC:
Summary: Adds a lightweight GHA workflow that runs the pytest unit tests under `fboss/oss/scripts/run_scripts/tests/` on PRs that touch `run_scripts/**`. The existing `docker-unittest.py` only discovers compiled gtest binaries, so these ~120 pure-Python tests currently run on no PR. Runs in seconds with no getdeps/docker.

Differential Revision: D106542800

### DIFF
--- a/.github/workflows/run-scripts-unittests.yml
+++ b/.github/workflows/run-scripts-unittests.yml
@@ -1,0 +1,42 @@
+name: run_scripts unit tests
+
+on:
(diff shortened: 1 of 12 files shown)


AarjunC:
Changes are committed.

PR #1240, opened 2026-05-28
[Nexthop][fboss2-dev] profile-change validation + subsumed-port handling

hillol-nexthop:
## Summary

Brings `fboss2-dev config interface <intfs> profile <profile>` to parity with [`patcher.py:change_speed`](fboss/cli/fboss2/test/integration_test/patcher.py)'s validation but makes it more reliant on platform mapping and avoiding per platform logic. Profile changes apply only to the explicitly-listed target port(s); subsumed siblings are stripped on upshift; freed sibling lanes on downshift are **not** auto-created.

**Validation**: per-port profile-membership + ASIC-family adjacent-port rules (12T/25T/51T) + 51T 800G-only-on-lane-1 + subsumed-port warnings + batch atomicity + aggregated errors.

**Upshift** (subsumed-port handling): when a larger-lane profile claims neighboring lanes, the sibling ports listed in `PlatformPortConfig.subsumedPorts` are stripped from `cfg::ports`, `vlanPorts`, and `cfg::Interface` entries that reference them — otherwise the agent aborts on apply.

**Downshift**: only the user-listed port(s) change profile. Freed sibling lanes stay unconfigured; the operator brings them up explicitly.

**Apply path**: profile changes go through **coldboot** instead of `reloadConfig` (rationale below).

# Test Plan

**Unit** 
- fboss/cli/fboss2/test/ProfileValidationTest.cpp

**Integration** 
- ConfigInterfaceProfileTest.cpp

### DIFF
--- a/cmake/CliFboss2Test.cmake
+++ b/cmake/CliFboss2Test.cmake
@@ -108,6 +108,7 @@ add_executable(fboss2_cmd_test
   fboss/cli/fboss2/test/CmdStopPcapTest.cpp
   fboss/cli/fboss2/test/GitTest.cpp
   fboss/cli/fboss2/test/PortMapTest.cpp
+  fboss/cli/fboss2/test/ProfileValidationTest.cpp
 )
 
 target_link_libraries(fboss2_cmd_test
--- a/fboss/cli/fboss2/commands/config/interface/CmdConfigInterface.cpp
+++ b/fboss/cli/fboss2/commands/config/interface/CmdConfigInterface.cpp
@@ -22,6 +22,8 @@
 #include <exception>
 #include <iostream>
 #include <optional>
+#include <ostream>
+#include <set>
 #include <stdexcept>
 #include <string>
 #include <unordered_set>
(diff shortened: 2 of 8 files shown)


joseph5wu:
@hillol-nexthop: I decided to implement the full adjusting port profile and creating/deleting an interface by myself cause I'd like to reuse some of our existing config utility functions from our existing sai_hw_test.
If you're interested, you can take a look at the following 4 commits:
- https://github.com/facebook/fboss/commit/1643e6846a0d0b87bef1fb3f81c9d51b4e18684b
- https://github.com/facebook/fboss/commit/2ceb4524ef894a6c737c4d7950c3a09fc19fa688
- https://github.com/facebook/fboss/commit/890ca34a5f318e7fdf819a0c49bd40d3a3c30d44
- https://github.com/facebook/fboss/commit/3101a5a3069bc2d672a105e11e267544f903d4cd

Please take a look at `fboss/cli/fboss2/test/integration_test/ConfigInterfaceProfileTest.cpp` as I implemented a way to find a candidate interface to adjust based on the test requirements rather than just skip the tests 

You can drop this PR now.

cc: @benoit-nexthop

PR #1221, opened 2026-05-21
[Nexthop][NPI] "asic_config_v3" part 4 - Add platform configurations

marif-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

# Summary

Implements `asic_config_v3` design.

In this PR, add platform configurations for the four core XGS platforms which were originally generated by `asic_config_v2`:
- montblanc
- icecube800bc
- icecube800banw
- wedge800bact

After this commit the `asic_config_v3` generator produces four XGS output files which should be identical to the `asic_config_v2` output.

**Note: After review comments, changed variant name `base` to `default`.**
v2: montblanc_base.yml
v3: montblanc_defaultyml

v2: wedge800bact_base.yml
v3: wedge800bact_default.yml

# Test Plan

Ran `fboss/lib/asic_config_v3/run-helper.sh` to generate `asic_config_v3` output:
```
Completed build for fboss-asic-config-v3-gen.GEN_PY_EXE
...
Writing to /var/FBOSS/fboss/fboss/lib/asic_config_v3/generated_asic_configs/icecube800banw_default.yml
...
Writing to /var/FBOSS/fboss/fboss/lib/asic_config_v3/generated_asic_configs/wedge800bact_default.yml
...
Writing to /var/FBOSS/fboss/fboss/lib/asic_config_v3/generated_asic_configs/montblanc_default.yml
...
Writing to /var/FBOSS/fboss/fboss/lib/asic_config_v3/generated_asic_configs/icecube800bc_default.yml
Configs have been written to specified output directory

$ ls -1 fboss/lib/asic_config_v3/generated_asic_configs
icecube800banw_default.yml
icecube800bc_default.yml
montblanc_default.yml
wedge800bact_default.yml
```

The generated output files are byte-identical with `asic_config_v2`
output
```
$ diff -r fboss/lib/asic_config_v3/generated_asic_configs/ fboss/lib/asic_config_v2/generated_asic_configs/
<no diff>
```

Schema validation also passed for all files
```
$ python3 ./fboss/lib/asic_config_v3/validation/validate_asic_configs.py
Vendor common configs (vendor_common.schema.json):
  vendors/broadcom/xgs/sai_common.json: PASSED
  vendors/broadcom/xgs/sdk_common.json: PASSED

[...]

PR #1199, opened 2026-05-14
[Nexthop][NPI] "asic_config_v3" part 2 - JSON schemas and validation script

marif-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

# Summary

Implements `asic_config_v3` design.

Add strict JSON schemas for the Broadcom `XGS` ASIC config, the platform config, and the shared vendor common JSONs. Include a validation script that scans the `asic_config_v3` directory tree and validates every matching file against its schema.

# Test Plan

Install required package `python3 -m pip install jsonschema`
Run schema validation
```
python3 fboss/lib/asic_config_v3/validation/validate_asic_configs.py
Vendor common configs (vendor_common.schema.json):
  No files found.
Broadcom XGS ASIC configs (broadcom_xgs_asic_config.schema.json):
  No files found.
Platform configs (platform_config.schema.json):
  No files found.
0 passed, 0 failed
```
_Note: There are no json files to validate at this point. Will be added in the next PR._

----

**This PR is created on top of**
- https://github.com/facebook/fboss/pull/1131
- https://github.com/facebook/fboss/pull/1149 
- https://github.com/facebook/fboss/pull/1198 

**so it contains diff from these PRs. When these PRs will merge, this PR will have a shorter diff.**

PR #1261, opened 2026-06-04
[Nexthop][NPI] "asic_config_v3" - Add icetea800bc platform

marif-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Implements `asic_config_v3` design.

In this PR, add platform configurations for `icetea800bc` variant (with no variant name) of `icetea800bc` platform.

After this commit the `asic_config_v3` generator produces the following extra output file
- `fboss/lib/asic_config_v3/generated_asic_configs/icetea800bc.yml`

# Test Plan
The generated output file is similar to `asic_config_v2` synced config
except a few lines out-of-order:

```
diff fboss/lib/asic_config_v3/generated_asic_configs/icetea800bc.yml fboss/lib/asic_config_v2/synced_asic_configs/icetea800bc.yml
1400d1399
<       sai_stats_disable_mask: 0x800
1402,1404c1401
<       l3_ecmp_member_secondary_mem_size: 24576
<       l3_ecmp_member_first_lkup_mem_size: 8192
<       dlb_quality_adjustment_default_value_select: 1
---
>       sai_stats_disable_mask: 0x800
1412a1410,1412
>       l3_ecmp_member_secondary_mem_size: 24576
>       l3_ecmp_member_first_lkup_mem_size: 8192
>       dlb_quality_adjustment_default_value_select: 1
```

---

**This PR is created on top of**
- https://github.com/facebook/fboss/pull/1131
- https://github.com/facebook/fboss/pull/1149 
- https://github.com/facebook/fboss/pull/1198 
- https://github.com/facebook/fboss/pull/1199
- https://github.com/facebook/fboss/pull/1220
- https://github.com/facebook/fboss/pull/1221

**so it contains diff from these PRs. When these PRs will merge, this PR will have a shorter diff.**


fboss/configs/platforms/celestica/icetea800bc/asic_config_v3/asic_config.json
```diff
@@ -0,0 +1,55 @@
...
+        "800000": "PC_FEC_RS544_2XN_IEEE"
+      },
+      "enable": 1
+    },
+    "cpu_port": {
+      "speed": 10000,
+      "num_lanes": 1
+    },
+    "mgmt_port": {
+      "enabled": true,
```
joseph5wu, line 28: Do we still need "enabled" if a user is trying to specify `mgmt_port` fields here?
Based on https://github.com/facebook/fboss/blob/main/fboss/lib/asic_config_v3/generators/broadcom_xgs_generator.py#L381
"enabled" seems to be implicit. Let's consider removing "enabled" option for `mgmt_port` like all the other categories

fboss/lib/asic_config_v3/platforms/icetea800bc/asic_config.json
```diff
@@ -0,0 +1,55 @@
...
+      "l3_ecmp_member_first_lkup_mem_size": "8192",
+      "dlb_quality_adjustment_default_value_select": "1"
+    },
+    "features": {
+      "generate_autoload_board_settings": true,
+      "generate_dlb_ecmp_config": true
+    }
+  },
+  "variants": {
+    "internal": {}
```
joseph5wu: Why do we need to name it `internal` while the synced is just 
https://github.com/facebook/fboss/blob/main/fboss/lib/asic_config_v2/synced_asic_configs/icetea800bc.yml

joseph5wu (commented):
I wonder whether we can still control the order of the fields so that we can keep the two files exactly the same as our synced versions.
In our asic_config_v2, we don't shuffle the fields per platform. I wonder we can also control that for v3

marif-nexthop:
@joseph5wu Please check the latest diff between the following asic config files. It only has minor ordering diff.
```
 diff fboss/lib/asic_config_v3/generated_asic_configs/icetea800bc_internal.yml fboss/lib/asic_config_v2/synced_asic_configs/icetea800bc.yml
1400d1399
<       sai_stats_disable_mask: 0x800
1402,1404c1401
<       l3_ecmp_member_secondary_mem_size: 24576
<       l3_ecmp_member_first_lkup_mem_size: 8192
<       dlb_quality_adjustment_default_value_select: 1
---
>       sai_stats_disable_mask: 0x800
1412a1410,1412
>       l3_ecmp_member_secondary_mem_size: 24576
>       l3_ecmp_member_first_lkup_mem_size: 8192
>       dlb_quality_adjustment_default_value_select: 1
```

fboss/lib/asic_config_v3/platforms/icetea800bc/asic_config.json
marif-nexthop: It was a placeholder until we supported generating `<platform>.yml` without any variant name. 
Now it is a supported feature and we can generate `<platform>.yml` or `<platform>_<variant>.yml` (when variant name is given).

fboss/configs/platforms/celestica/icetea800bc/asic_config_v3/asic_config.json
marif-nexthop: That is a good point for platforms that already specify other `mgmt_port fields`. There is one case where the `enabled` flag is still useful. When a platform wants the management port generated entirely from the ASIC `mgmt_port_defaults` without any platform specific overrides.

In that situation there are no other `mgmt_port` fields that could convey the intent, so the block ends up being just `{ "enabled": true }`. `tahan800bc` is a good example. It sets only that one flag and still emits a full management port that is sourced from the `mgmt_port_defaults` in `tomahawk5.json`, with `PORT_ID: 76, SPEED: 100000, NUM_LANES: 4, and FEC_MODE: PC_FEC_RS528`.

PR #1268, opened 2026-06-08
[Nexthop][fboss2-dev] Add config icmpv4-unavailable-src-addr CLI command

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

## Summary

Adds a new write command that sets the optional
`sw.icmpV4UnavailableSrcAddress` field in the agent switch config:

```
fboss2-dev config switch icmpv4-unavailable-src-addr <ipv4-address>

# examples
fboss2-dev config switch icmpv4-unavailable-src-addr 10.0.0.1
fboss2-dev config switch icmpv4-unavailable-src-addr 192.0.0.8   # RFC 7600 default
```

The command is nested under the `config switch` command group (introduced by
the admin-distance command) rather than at the top level, grouping
switch-level settings that do not belong to a specific network object.

FBOSS uses this address as the source IP in ICMP error replies (TTL exceeded,
destination unreachable) when no interface address is available.

- **Action level**: HITLESS — no agent restart required.
- **Upsert**: no-op if the field is already set to the requested value.
- **Validation**: the argument is validated via `folly::IPAddressV4::tryFromString`.
- **Field**: `switch_config.thrift:SwitchConfig.icmpV4UnavailableSrcAddress`
  (optional string). When unset, the agent falls back to `192.0.0.8` (the
  RFC 7600 IANA dummy address) at read time, so the value is always populated
  in switch state:

  ```cpp
  // SwitchSettings.h:397 — SwitchState::getIcmpV4UnavailableSrcAddress()
  if (srcAddress.addr()->size() == 0) {
    return folly::IPAddressV4({192, 0, 0, 8});  // default when unset
  }
  ```

### Unit tests (`CmdConfigIcmpV4UnavailableSrcAddr{,Absent}TestFixture`)

- `argValidation` — valid IPs accepted; invalid strings, out-of-range octets,
  and IPv6 rejected.
- `setAddress` — sets the value, verifies the in-memory config is updated,
  confirms the success path.
- `alreadySet` — no-op when the value is unchanged; config untouched.
- `setAddressWhenFieldAbsent` — first-ever set when the field is not present in
  config (the real device default state).

### Integration test (`ConfigIcmpV4UnavailableSrcAddrTest`)

Because the field is optional, the test never reads it before the first write
(a single-arg read throws when the field is absent). It sets a known value,
reads it back via the running config to confirm the round-trip, then restores
the well-known RFC 7600 default (`192.0.0.8`) — behaviorally identical to the
field being absent. This means the test needs no default-value `getSwConfigField`
overload: `Fboss2IntegrationTest.h` is unchanged from upstream.

Results from the internal CLI Integration Test (DUT) run on 2026-07-03
(device gold414):

```
[----------] 1 test from ConfigIcmpV4UnavailableSrcAddrTest
[ RUN      ] ConfigIcmpV4UnavailableSrcAddrTest.SetAndRestoreAddress
  [Step 1] Setting address to 10.0.0.1...
  Running CLI command: config switch icmpv4-unavailable-src-addr 10.0.0.1
  Verified: address is 10.0.0.1
  [Step 2] Restoring address to default 192.0.0.8...
  Running CLI command: config switch icmpv4-unavailable-src-addr 192.0.0.8
  TEST PASSED
[       OK ] ConfigIcmpV4UnavailableSrcAddrTest.SetAndRestoreAddress (125 ms)
[----------] 1 test from ConfigIcmpV4UnavailableSrcAddrTest (125 ms total)
```

Ran as part of the 60-test fboss2 CLI integration suite; all 60 passed
(total 225 s).

## Review Findings

Pre-publication review (single-pass automated reviewer). One item raised —
unchecked `*config.sw()` dereference in `CmdConfigIcmpV4UnavailableSrcAddr.cpp`
— reviewed and dismissed: it is the established pattern across all
`fboss2 config` commands (e.g. `CmdConfigArp.cpp`, `CmdConfigCopp.cpp`); `sw`
is reliably present for a loaded agent config. No issues above threshold.

## Presubmission Checks

- **clang-format** (`--style=file --dry-run --Werror`): pass on all added/modified C++ sources.
- **pre-commit**: pass (no violations on changed files).


joseph5wu:
@vybhav-nexthop : Please consider to follow the review comments as https://github.com/facebook/fboss/pull/1274 to introduce the CmdConfigSwitch.h/cpp to handle the global SwitchSetting or SwitchConfig attributes

PR #1278, opened 2026-06-10
[Nexthop] Add fboss-sim: a minimal fake-SAI FBOSS runtime container

hillol-nexthop:
## Summary
Adds `fboss-sim`, a minimal fake-SAI FBOSS runtime container. It packages the fake-SAI FBOSS agents and the `fboss2` CLI into a small (~1.2 GB) CentOS Stream 9 Docker image, so FBOSS can be run and exercised end-to-end without any switching hardware.

## What's included
- **`fboss-sim/docker/Dockerfile.runtime`** — builds the runtime image from the CentOS Stream 9 base plus the existing FBOSS `centos-09.0` rootfs overlay (`fboss-image/image_builder/templates/centos-09.0/root_files/`: systemd agent units, login env hook, eth0 IPv6 sysctl) and the collected binaries + shared libraries.
- **`fboss-sim/scripts/fboss-sim-docker-package.py`** — collects the fake-SAI binaries and their `ldd`-resolved shared libraries and builds the image. Two phases:
  - `--collect-only`: resolve binaries + libs into `tmp_build_dir/`. Must run where `ldd` resolves the image-base libraries (CentOS Stream 9), e.g. inside the CentOS build image.
  - `--build-only`: `docker build` the image from `tmp_build_dir/` on the host that owns the Docker daemon.
  - A flagless run does both (when the host is already CentOS Stream 9).
- **`fboss-sim/scripts/fboss-sim-docker-run.py`** — creates an IPv6 network and starts the runtime container in split or monolithic agent mode.
- **`.github/workflows/fboss-sim.yml`** — a GitHub Actions job that builds the fake-SAI agent + `fboss2` targets via `docker-build.py`, packages and starts the runtime image (collect inside the CentOS `fboss_image`, build on the host), and runs `fboss2_integration_test`. Mirrors the existing `agent.yml` / `fboss2-cli.yml` pattern.

## Build naming
The `BUILD_SAI_FAKE` CMake arm now names the fake-SAI agents `wedge_agent-sai_impl` / `fboss_hw_agent-sai_impl` (instead of `-fake`), so they land at the `/opt/fboss/bin` paths the systemd units and `package.py` already reference — the runtime no longer needs `-fake` → `-sai_impl` symlinks.

## Testing
Validated locally with a fake-SAI build:
- `--collect-only` inside the build container → `--build-only` on the host → image built from the `centos-09.0` template rootfs.
- Container starts; both `fboss_sw_agent` and `fboss_hw_agent@0` go active; `fboss2 show interface` serves fake Montblanc ports.
- `fboss2_integration_test`: 38 tests / 18 suites, all passed.


hillol-nexthop:
there is a list of tests failing which need to be looked at separately but this is the fboss-sim infra and GHA.

PR #328, opened 2025-01-30
[Darwin]: Led_manager use sys leds paths and add missing fan module

aalamsi22:
# Description

- Sync Darwin led_manager paths to follow the same `/sys/class/leds` used on all other platforms.
- Add missing rackmon fan as a fru associated with the FAN led.

# Testing

The following sw_tests passed:
- async_logger_test
- fan_service_sw_test
- fsdb_client_test
- platform_config_lib_config_lib_test
- platform_data_corral_sw_test
- platform_helpers_platform_name_lib_test
- platform_manager_config_validator_test
- platform_manager_data_store_test
- platform_manager_device_path_resolver_test
- platform_manager_i2c_explorer_test
- platform_manager_platform_explorer_test
- platform_manager_presence_checker_test
- platform_manager_utils_test
- rackmon_test
- sensor_service_sw_test
- sensor_service_utils_test
- thrift_cow_visitor_tests
- thrift_node_tests
- weutil_crc16_ccitt_test
- weutil_fboss_eeprom_parser_test

The following hw_tests passed:
- data_corral_service_hw_test
- sensor_service_hw_test
- fan_service_hw_test
- weutil_hw_test

### Led Manager Testing
```
Jan 31 00:49:20 ...: I0131 00:49:20.856894  5161 FruPresenceExplorer.cpp:26] Detecting presence of FRUs
Jan 31 00:49:20 ...: I0131 00:49:20.856950  5161 FruPresenceExplorer.cpp:34] Detecting presence of FAN1 (via sysfs)
Jan 31 00:49:20 ...: I0131 00:49:20.857852  5161 FruPresenceExplorer.cpp:49] Detected that FAN1 is present
Jan 31 00:49:20 ...: I0131 00:49:20.857892  5161 FruPresenceExplorer.cpp:34] Detecting presence of FAN2 (via sysfs)
Jan 31 00:49:20 ...: I0131 00:49:20.858802  5161 FruPresenceExplorer.cpp:49] Detected that FAN2 is present
Jan 31 00:49:20 ...: I0131 00:49:20.858819  5161 FruPresenceExplorer.cpp:34] Detecting presence of FAN3 (via sysfs)
Jan 31 00:49:20 ...: I0131 00:49:20.859799  5161 FruPresenceExplorer.cpp:49] Detected that FAN3 is present
Jan 31 00:49:20 ...: I0131 00:49:20.859815  5161 FruPresenceExplorer.cpp:34] Detecting presence of FAN4 (via sysfs)
[...]

PR #1302, opened 2026-06-16
[Nexthop][fboss2-dev] Add config admin-distance CLI command

vybhav-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

# Summary

Adds a new write command:

    fboss2-dev config switch admin-distance <client-id> <distance>

Sets the administrative distance for a routing client in the agent running config via a COLDBOOT config session.

### What

- New command CmdConfigAdminDistance with AdminDistanceArg validating two positional args: client-id (non-negative integer) and distance (0-255)
- Read-modify-write on clientIdToAdminDistance map - all other client entries survive untouched
- No-op detection - if the value is already set, returns early without saving
- Registered under `config switch` as a switch-level setting
- Blocks client-ids 1/2/3/4 (STATIC_ROUTE, INTERFACE_ROUTE, LINKLOCAL_ROUTE, REMOTE_INTERFACE_ROUTE) whose admin distances are hardcoded in the agent and cannot be changed via config
- Wired into CMake (fboss2_config_lib, fboss2_cmd_config_test, fboss2_integration_test) and BUCK

# Test Plan

Validated by a full CI run of this PR rebased onto the latest stable commit (`ac0f8c3e61`): complete unit-test suite plus the fboss2 CLI integration suite against a fake-SAI simulator running the split sw/hw agents.

### Unit tests — full suite: 1592/1592 passed

    Start  325: CmdConfigAdminDistanceTestFixture.argValidation
    Start  326: CmdConfigAdminDistanceTestFixture.updateExistingClient
    Start  327: CmdConfigAdminDistanceTestFixture.addNewClient
    Start  328: CmdConfigAdminDistanceTestFixture.alreadySet
    325/1592 Test  #325: CmdConfigAdminDistanceTestFixture.argValidation ........... Passed 2.30 sec
    326/1592 Test  #326: CmdConfigAdminDistanceTestFixture.updateExistingClient .... Passed 2.29 sec
    327/1592 Test  #327: CmdConfigAdminDistanceTestFixture.addNewClient ............ Passed 2.29 sec
    328/1592 Test  #328: CmdConfigAdminDistanceTestFixture.alreadySet .............. Passed 2.29 sec
    ...
    100% tests passed, 0 tests failed out of 1592

### Integration tests — fboss2_integration_test on fake-SAI sim: 99/99 passed (31 suites)

    [----------] 2 tests from ConfigAdminDistanceTest
    [ RUN      ] ConfigAdminDistanceTest.SetAndRestoreAdminDistance
    I0702 13:58:56] [Step 1] Reading current admin distances...
    I0702 13:58:56]   client-id 0 -> 20
    I0702 13:58:56] [Step 2] Setting client-id 0 to 42...
    I0702 13:58:56] Running CLI command: config switch admin-distance 0 42
    I0702 13:58:57] Running CLI command: config session commit
    ...
    [       OK ] ConfigAdminDistanceTest.SetAndRestoreAdminDistance (59146 ms)
    [ RUN      ] ConfigAdminDistanceTest.ForbiddenClientIdsRejected
    I0702 13:59:56]   Verifying client-id 1 (STATIC_ROUTE) is rejected...
    E0702 13:59:56] Error in command execution: Invalid argument: FBOSS does not allow changing admin distance for client-id 1: STATIC_ROUTE is hardcoded to AdminDistance::STATIC_ROUTE in the agent
    ...
    [       OK ] ConfigAdminDistanceTest.ForbiddenClientIdsRejected
    ...
    [==========] 99 tests from 31 test suites ran. (184487 ms total)
    [  PASSED  ] 99 tests.

### DIFF
--- a/fboss/cli/fboss2/commands/config/switch/admin_distance/CmdConfigAdminDistance.cpp
+++ b/fboss/cli/fboss2/commands/config/switch/admin_distance/CmdConfigAdminDistance.cpp
@@ -1,0 +1,139 @@
+/*
+ *  Copyright (c) 2004-present, Facebook, Inc.
+ *  All rights reserved.
+ *
+ *  This source code is licensed under the BSD-style license found in the
+ *  LICENSE file in the root directory of this source tree. An additional grant
+ *  of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+#include "fboss/cli/fboss2/commands/config/switch/admin_distance/CmdConfigAdminDistance.h"
+
+#include "fboss/cli/fboss2/CmdHandler.cpp"
+
+#include <fmt/format.h>
+#include <folly/Conv.h>
+#include <iostream>
+#include "fboss/cli/fboss2/session/ConfigSession.h"
+
+namespace facebook::fboss {
+
+namespace {
+constexpr int32_t kMaxAdminDistance = 255;
+
+// ClientIDs whose admin distance is hardcoded in the agent and cannot be
+// overridden via clientIdToAdminDistance config:
+//   STATIC_ROUTE (1)         -> AdminDistance::STATIC_ROUTE
+//   INTERFACE_ROUTE (2)      -> AdminDistance::DIRECTLY_CONNECTED
+//   LINKLOCAL_ROUTE (3)      -> AdminDistance::DIRECTLY_CONNECTED
+//   REMOTE_INTERFACE_ROUTE (4) -> AdminDistance::DIRECTLY_CONNECTED
+const std::unordered_map<int32_t, std::string>& forbiddenClients() {
+  static const std::unordered_map<int32_t, std::string> kForbidden = {
+      {1,
+       "STATIC_ROUTE is hardcoded to AdminDistance::STATIC_ROUTE in the agent"},
+      {2,
+       "INTERFACE_ROUTE is hardcoded to AdminDistance::DIRECTLY_CONNECTED in the agent"},
+      {3,
+       "LINKLOCAL_ROUTE is hardcoded to AdminDistance::DIRECTLY_CONNECTED in the agent"},
+      {4,
+       "REMOTE_INTERFACE_ROUTE is hardcoded to AdminDistance::DIRECTLY_CONNECTED in the agent"},
+  };
+  return kForbidden;
+}
+} // namespace
+
+AdminDistanceArg::AdminDistanceArg(std::vector<std::string> v) {
+  if (v.size() != 2) {
+    throw std::invalid_argument(
+        fmt::format(
+            "Expected exactly two arguments: <client-id> <distance>, got {}",
+            v.size()));
+  }
+
+  try {
+    clientId_ = folly::to<int32_t>(v[0]);
+  } catch (const folly::ConversionError&) {
+    throw std::invalid_argument(
+        fmt::format("Invalid client-id '{}': must be an integer", v[0]));
+  }
+  if (clientId_ < 0) {
+    throw std::invalid_argument(
+        fmt::format("client-id must be a non-negative integer, got {}", v[0]));
+  }
+
+  const auto& forbidden = forbiddenClients();
+  auto it = forbidden.find(clientId_);
+  if (it != forbidden.end()) {
+    throw std::invalid_argument(
+        fmt::format(
+            "FBOSS does not allow changing admin distance for client-id {}: {}",
+            clientId_,
+            it->second));
+  }
+
+  try {
+    distance_ = folly::to<int32_t>(v[1]);
+  } catch (const folly::ConversionError&) {
+    throw std::invalid_argument(
+        fmt::format("Invalid distance '{}': must be an integer", v[1]));
+  }
+  if (distance_ < 0 || distance_ > kMaxAdminDistance) {
+    throw std::invalid_argument(
+        fmt::format(
+            "distance must be in the range [0, {}], got {}",
+            kMaxAdminDistance,
+            v[1]));
+  }
+
+  data_ = std::move(v);
+}
+
+CmdConfigAdminDistanceTraits::RetType CmdConfigAdminDistance::queryClient(
+    const HostInfo& /* hostInfo */,
+    const ObjectArgType& arg) {
+  auto& session = ConfigSession::getInstance();
+  auto& config = session.getAgentConfig();
+  auto& swConfig = *config.sw();
+
+  int32_t clientId = arg.getClientId();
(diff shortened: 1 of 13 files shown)


fboss/cli/fboss2/commands/config/admin_distance/CmdConfigAdminDistance.cpp
```diff
@@ -0,0 +1,104 @@
...
+          "Admin distance for client-id {} is already set to {}",
+          clientId,
+          distance);
+    }
+    it->second = distance;
+  } else {
+    adminDistanceMap.emplace(clientId, distance);
+  }
+
+  session.saveConfig(cli::ServiceType::AGENT, cli::ConfigActionLevel::HITLESS);
```
joseph5wu: As you mentioned in the summary, clientIdToAdminDistance is consulted only at route-program time, in the Thrift RPC path (RouteUpdateWrapper::programRoutesImpl → SwSwitch::clientIdToAdminDistance reads live getConfig()). The looked-up value is then stamped onto each client's RouteNextHopEntry and stored. Best-path selection compares the stored admin distances — it never re-derives from config. Nothing re-stamps existing routes when the config map changes (there's no old-vs-new diff of this map anywhere).
Therefore, for BGP/OpenR/TE: we actually need to trigger a coldboot. A hitless reload/warmboot alone won't do it.

fboss/cli/fboss2/commands/config/switch/admin_distance/CmdConfigAdminDistance.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+CmdConfigAdminDistanceTraits::RetType CmdConfigAdminDistance::queryClient(
+    const HostInfo& /* hostInfo */,
+    const ObjectArgType& arg) {
+  auto& session = ConfigSession::getInstance();
+  auto& config = session.getAgentConfig();
+  auto& swConfig = *config.sw();
+
+  int32_t clientId = arg.getClientId();
```
joseph5wu, line 99: Btw there're some subtle hardcoding in the current admin distance for certain client:
- For Static routes from config (STATIC_ROUTE), ConfigApplier hardcodes AdminDistance::STATIC_ROUTE (RouteNextHopEntry.cpp:619)
- For Interface / link-local routes, Hardcoded AdminDistance::DIRECTLY_CONNECTED (ConfigApplier.cpp:151,163).

Therefore, we should stop any user to change the admin distance for these client ids:
- STATIC_ROUTE
- INTERFACE_ROUTE
- LINKLOCAL_ROUTE
- REMOTE_INTERFACE_ROUTE
As when you change the admin distance for them, it won't actually apply the new admin distance any way.
In this case, if a user wants to use config cli to change the admin distance for any of these client ids, just throw an error to inform the user that FBOSS doesn't allow to change the admin distance for such clients. And feel free to use the AdminDistance::STATIC_ROUTE and AdminDistance::DIRECTLY_CONNECTED in the log

fboss/cli/fboss2/test/config/CmdConfigAdminDistanceTest.cpp
```diff
@@ -0,0 +1,142 @@
...
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include "fboss/cli/fboss2/commands/config/admin_distance/CmdConfigAdminDistance.h"
+#include "fboss/cli/fboss2/session/ConfigSession.h"
+#include "fboss/cli/fboss2/test/config/CmdConfigTestBase.h"
+
+using namespace ::testing;
+
+namespace facebook::fboss {
```
joseph5wu, line 20: Since we won't be able to update the admin distance for certain client ids, please add unit tests to verify that

fboss/cli/fboss2/test/integration_test/ConfigAdminDistanceTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+         "admin-distance",
+         std::to_string(clientId),
+         std::to_string(distance)});
+    ASSERT_EQ(result.exitCode, 0)
+        << "admin-distance CLI failed: " << result.stderr;
+    commitConfig();
+  }
+};
+
+TEST_F(ConfigAdminDistanceTest, SetAndRestoreAdminDistance) {
```
joseph5wu, line 70: Make sure we have both allowed and forbidden client ids in the integration tests and make sure it's using coldboot when supported

joseph5wu (changes requested):
This should be a switch level setting, and please consider move this cli header/files under the switch/ folder

fboss/cli/fboss2/commands/config/admin_distance/CmdConfigAdminDistance.cpp
vybhav-nexthop: will change to coldboot to ensure the all the routes for the given client are updated(vs with hitless , we only updated the new routes)

fboss/cli/fboss2/commands/config/switch/admin_distance/CmdConfigAdminDistance.cpp
vybhav-nexthop: will add these in the next commit. thanks !

fboss/cli/fboss2/test/config/CmdConfigAdminDistanceTest.cpp
vybhav-nexthop: yes will verify those 4 client ids arent acceptable

fboss/cli/fboss2/test/integration_test/ConfigAdminDistanceTest.cpp
vybhav-nexthop: will add tests for the same

vybhav-nexthop:
@joseph5wu , could you kindly have a review of this PR, will rebase the others (with regards to the switch structure) off of this

fboss/cli/fboss2/commands/config/switch/admin_distance/CmdConfigAdminDistance.cpp
vybhav-nexthop: done in the latest commit

fboss/cli/fboss2/test/config/CmdConfigAdminDistanceTest.cpp
vybhav-nexthop: added 4 unit tests for the same

PR #1277, opened 2026-06-10
[Nexthop] WEDGE800CNHP will be an alias for WEDGE800CACT

andrew-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

# Summary

Wedge800CNHP and WEDGE800CACT should use the same platform code. Wedge800CNHP platform config updates will come from https://github.com/facebook/fboss/pull/1069

# Test Plan

FBOSS platform tests should run the same on both platforms.

PR #1266, opened 2026-06-05
[Nexthop] WEDGE800BNHP will be an alias for WEDGE800BACT

andrew-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

# Summary

WEDGE800BNHP and WEDGE800BACT should use the same platform code. Maps the WEDGE800BNHP BIOS platform name to WEDGE800BACT in sanitizePlatformName(), mirroring the existing platform aliasing.

# Test Plan

FBOSS platform tests should run the same on both platforms.

PR #1312, opened 2026-06-22
[Nexthop] fixmyfboss alias for BNHP and CNHP

andrew-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary
https://github.com/facebook/fboss/pull/1266 and https://github.com/facebook/fboss/pull/1277 added the alias for the platform services but we also need to update this equivalent mapping

# Test Plan
have the same alias behavior in .cpp and .py versions of PlatformNameLib

### DIFF
--- a/fboss/util/fixmyfboss/platform_name_lib.py
+++ b/fboss/util/fixmyfboss/platform_name_lib.py
@@ -19,20 +19,17 @@ def sanitize_platform_name(platform_name_from_bios: str) -> str:
     """
     platform_name_upper = platform_name_from_bios.upper()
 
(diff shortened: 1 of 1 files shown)


srikrishnagopu:
Can we not use aliasing for Agent and Qsfp service ? 

I would suggest to define new platform type for both the platforms and add the relevant code changes to support it.

andrew-nexthop:
> Can we not use aliasing for Agent and Qsfp service ?
> 
> I would suggest to define new platform type for both the platforms and add the relevant code changes to support it.

i updated this PR to only include the change for the python platform alias. WEDGE800BNHP agent/qsfp service code will be in a separate PR

Scott8440 (approved):

PR #1317, opened 2026-06-23
[Celestica] Leh800bcls: Fix switched packet routing fallback in triggerPfcGeneration for multi-NPU

gang-tao:
**Pre-submission checklist**
- [ ✓ ] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [ ✓ ] `pre-commit run`
clang-format.............................................................Passed
shellcheck...........................................(no files to check)Skipped
shfmt................................................(no files to check)Skipped
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...........................................(no files to check)Skipped
check json...........................................(no files to check)Skipped
check for merge conflicts................................................Passed
ruff check...........................................(no files to check)Skipped
ruff format..........................................(no files to check)Skipped
Prevent sai_impl in fboss manifest.......................................Passed

# Summary
This PR addresses a deterministic traffic-routing bug inside the test utility triggerPfcGeneration() which consistently causes lossless  PFC watermark tests to fail on multi-NPU/Split-Agent platforms (e.g., Leh800bcls with two Tomahawk 6 NPUs). 

By explicitly targeting the injected switched packets to the correct SwitchID of the active port under test, we restore 100% deterministic  queue buildup and PFC frame generation on secondary NPUs (e.g. switch1), preventing flaky test failures.

# Root Cause
Switched Packet Routing Fallback (PfcTestUtils):
     * Root Cause: On multi-NPU environments, calling sendPacketAsync with std::nullopt causes the routing handler MultiHwSwitchHandler::sendPacketSwitchedAsync to fallback and route all packets to SwitchID(0).
     * Result: When running tests on switch1, congestion traffic is incorrectly sent to switch0, leaving switch1 with zero traffic, zero PFC, and zero watermarks.

# Solution
 Switch-Aware Packet Injection in triggerPfcGeneration:
    Retrieve the targeted port's active parent SwitchID via ensemble->scopeResolver().scope(port).switchId(), and explicitly route switched packets through SwSwitch::sendPacketSwitchedAsync to target the active NPU under test, guaranteeing correct traffic injection.

# Test Plan
The following 3 Agent T1 test cases passed successfully on both NPU 0 and NPU 1.
```
[root@localhost log]# grep -h " OK " AgentTxNpu0/multi_switch_*.log
[       OK ] AgentTrafficPfcGenTest.verifyBufferPoolWatermarks (60775 ms)
[       OK ] AgentTrafficPfcGenTest.verifyIngressPriorityGroupWatermarks (60216 ms)
[       OK ] AgentTrafficPfcWatchdogTest.PfcWatchdogDetection (66896 ms)
[root@localhost log]# grep -h " OK " AgentTxNpu1/multi_switch_*.log
[       OK ] AgentTrafficPfcGenTest.verifyBufferPoolWatermarks (62278 ms)
[       OK ] AgentTrafficPfcGenTest.verifyIngressPriorityGroupWatermarks (60163 ms)
[       OK ] AgentTrafficPfcWatchdogTest.PfcWatchdogDetection (68499 ms)
[root@localhost log]#

```

### DIFF
--- a/fboss/agent/test/utils/PfcTestUtils.cpp
+++ b/fboss/agent/test/utils/PfcTestUtils.cpp
@@ -467,8 +467,9 @@ void triggerPfcGeneration(
               std::vector<uint8_t>(2000, 0xff));
 
           if (vlanId.has_value()) {
-            ensemble->sendPacketAsync(
-                std::move(txPacket), std::nullopt, pfcPriority);
+            ensemble->getSw()->sendPacketSwitchedAsync(
+                std::move(txPacket),
+                {ensemble->scopeResolver().scope(port).switchId()});
           } else {
             ensemble->sendPacketAsync(
                 std::move(txPacket), PortDescriptor(port), pfcPriority);


jrallapalli (commented):
Could you split this into two diffs? The PfcTestUtils.cpp send-targeting change is a clean, independent multi-NPU fix — please land it separately. The AgentTrafficPfcTests.cpp validation changes need a rethink before detailed review: they currently bypass the assertions they're meant to make (the continue at L256 passes the per-PG check on a global-watermark stand-in) and mix two different counter sources across the watchdog before/after comparison. Before adding a getHwSwitchStatsExpensive path, can we confirm why the scoped fb303 counters are empty on the switch under test in split-agent/OSS — that looks like the real bug. Happy to do a full line-level review once it's split and the approach is settled.

gang-tao:
> Could you split this into two diffs? The PfcTestUtils.cpp send-targeting change is a clean, independent multi-NPU fix — please land it separately. The AgentTrafficPfcTests.cpp validation changes need a rethink before detailed review: they currently bypass the assertions they're meant to make (the continue at L256 passes the per-PG check on a global-watermark stand-in) and mix two different counter sources across the watchdog before/after comparison. Before adding a getHwSwitchStatsExpensive path, can we confirm why the scoped fb303 counters are empty on the switch under test in split-agent/OSS — that looks like the real bug. Happy to do a full line-level review once it's split and the approach is settled.

@jrallapalli I totally agree with your suggestion. Let's use this PR for the send-targeting change, and I'll keep digging into why the fb303 counters are empty. Thanks for your review and comments!

gang-tao:
> fb303 counters are empty

@jrallapalli 
Regarding the empty fb303 counters, I did some investigation:

In Split Agent mode, the test main thread runs in the software agent process (multi_switch_agent), while the actual hardware watermarks and watchdog callbacks are processed and published strictly inside the local facebook::fb303::fbData singleton of the isolated hardware agent process (fboss_hw_agent-sai_impl).

In non-OSS builds, AgentEnsemble::getFb303RegexCounters() uses facebook::thrift::Monitor clients to fetch counters remotely across process boundaries. However, in IS_OSS open-source builds, the Monitor client classes are completely excluded from compilation.

Due to the missing Monitor classes, the **#else** branch in AgentEnsemble.cpp was historically forced to fall back to a process-local database query:
```
#ifndef IS_OSS
...  
#else
  // TODO: This needs to be updated to support multi-switch.
  counters = facebook::fb303::fbData->getRegexCounters(regex);
#endif
```

All updates regarding fb303 data occur within the hw agent. The specific call chain is as follows:
```
facebook::fboss::SaiSwitch::updateStatsImpl()                          //SaiSwitch.cpp per 1 sec
---> facebook::fboss::SaiBufferManager::publishGlobalWatermarks(...)   //SaiBufferManager.cpp 
--->  STATS_buffer_watermark_global_shared.addValue(globalSharedBytes) //SaiBufferManager.cpp
```
the following process is to save fb303 data:
```
void ThreadCachedServiceData::publishStats() //github.com-facebook-fb303.git/fb303/ThreadCachedServiceData.cpp
---> QuantileStatMap::makeKey(...)           //github.com-facebook-fb303.git/fb303/detail/QuantileStatMap-inl.h
---> facebook::fb303::fbData->setCounter()   //github.com-facebook-fb303.git/fb303/ServiceData.cpp 
```

PR #1330, opened 2026-06-24
[Nexthop][PlatformManager][M4062NHP] Support AMD CPU i2c buses C and D

anna-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Map the AMD i2c controllers I2CC and I2CD to CPU_BUS@2 and CPU_BUS@3 in I2cExplorer (index-aligned to the AMD i2c[N] controllers; the legacy I2CA/I2CB mapping is unchanged), and widen ConfigValidator to accept CPU_BUS@0 through CPU_BUS@3 in i2cAdaptersFromCpu. This lets a platform_manager.json model devices on the third and fourth CPU-direct i2c buses.

# Test Plan

Running the config_validator_test which now verifies CPU_BUS@[0-3] are valid while anything past that index is invalid.

### DIFF
--- a/fboss/platform/platform_manager/ConfigValidator.cpp
+++ b/fboss/platform/platform_manager/ConfigValidator.cpp
@@ -367,10 +367,11 @@ bool ConfigValidator::isValidI2cAdaptersFromCpu(
   std::set<std::string> seen;
   for (const auto& name : i2cAdaptersFromCpu) {
     if (re2::RE2::FullMatch(name, kCpuBusNameRegex)) {
-      if (name != "CPU_BUS@0" && name != "CPU_BUS@1") {
+      static const re2::RE2 kSupportedCpuBusNameRegex{"CPU_BUS@[0-3]"};
+      if (!re2::RE2::FullMatch(name, kSupportedCpuBusNameRegex)) {
         XLOG(ERR) << fmt::format(
             "Invalid virtual bus name '{}'. "
-            "Only CPU_BUS@0 and CPU_BUS@1 are supported",
+            "Only CPU_BUS@0 through CPU_BUS@3 are supported",
             name);
         return false;
       }
--- a/fboss/platform/platform_manager/I2cExplorer.cpp
+++ b/fboss/platform/platform_manager/I2cExplorer.cpp
@@ -179,8 +179,13 @@ std::map<std::string, uint16_t> I2cExplorer::resolveAmdCpuBusNums(
 std::map<std::string, uint16_t> I2cExplorer::resolveIntelCpuBusNums(
     const std::vector<std::string>& i2cAdaptersFromCpu) {
   // One SMBus I801 adapter per unit — only CPU_BUS@0 is valid.
-  CHECK_EQ(i2cAdaptersFromCpu.size(), 1)
-      << "resolveIntelCpuBusNums only supports a single CPU_BUS@0 entry";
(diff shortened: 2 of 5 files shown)


fboss/platform/platform_manager/I2cExplorer.cpp
```diff
@@ -26,6 +26,8 @@ constexpr auto kCpuI2cBusNumsWaitSecs = 1;
 const std::map<std::string, int> kAmdAcpiPathToBusIndex = {
     {R"(\_SB_.I2CA)", 1},
     {R"(\_SB_.I2CB)", 0},
+    {R"(\_SB_.I2CC)", 2},
+    {R"(\_SB_.I2CD)", 3},
 };
```
somasun, line 31: Please update the thrift configuration to match the code behavior.

somasun, line 31: Curious, why did you have to use these two buses? I am asking since I haven't come across a usecase where these buses were leveraged for devices.

fboss/platform/platform_manager/ConfigValidator.cpp
```diff
@@ -368,10 +367,11 @@ bool ConfigValidator::isValidI2cAdaptersFromCpu(
...
     if (re2::RE2::FullMatch(name, kCpuBusNameRegex)) {
-      if (name != "CPU_BUS@0" && name != "CPU_BUS@1") {
+      static const re2::RE2 kSupportedCpuBusNameRegex{"CPU_BUS@[0-3]"};
+      if (!re2::RE2::FullMatch(name, kSupportedCpuBusNameRegex)) {
         XLOG(ERR) << fmt::format(
             "Invalid virtual bus name '{}'. "
-            "Only CPU_BUS@0 and CPU_BUS@1 are supported",
+            "Only CPU_BUS@0 through CPU_BUS@3 are supported",
             name);
         return false;
```
somasun, line 376: Add a run time check in platform_manager than the code fails gracefully with clear message when a config is specified with CPU_BUS other than [0,1] on a Intel system ? (since that is not a supported usecase and the config validation is not catching that now)

somasun (changes requested):

fboss/platform/platform_manager/ConfigValidator.cpp
anna-nexthop: @somasun The function in fboss/platform/platform_manager/I2cExplorer.cpp `resolveAmdCpuBusNums` asserts that only one cpu bus is defined and the message assumes it should be 0. Will test to that instead of the suggested [0,1]

fboss/platform/platform_manager/I2cExplorer.cpp
anna-nexthop: Updated the comment in the thrift file to align with the code changes.

anna-nexthop: We selected that topology because it was the best trade-off given the platform's bus architecture and integration constraints.

fboss/platform/platform_manager/ConfigValidator.cpp
anna-nexthop: Tightened up the check and added a unit test for Intel.

somasun: ack. thanks!

PR #1332, opened 2026-06-24
[Nexthop][m4062nhp] Add Platform Manager support

anna-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the
    files I modified in this PR. You can install the linters by running `pip
    install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

# Summary
Add platform_manager.json for the M4062nhp platform, modeling the SCM and SMB PmUnits (PCI/I2C/FPGA topology, chassis EEPROM, port LEDs, fans, PSUs) per the Platform Manager spec. Also add a ConfigLibTest case that deserializes and validates the committed config.

Requires https://github.com/facebook/fboss/pull/1330 for exposing two additional cpu busses off of the ASIC. The changes are encompassed in 80a83922df286145ff344684ce4654ced101c941

# Test Plan
Validated the JSON parses and conforms to platform_manager_config.thrift.
Built the BSP kmods RPM from nexthop-ai/nexthop.fboss.bsp.kmods and confirmed it provides every module in requiredKmodsToLoad.
Added a ConfigLibTest case that deserializes and runs ConfigValidator on the committed config.

### DIFF
--- a/fboss/platform/configs/m4062nhp/platform_manager.json
+++ b/fboss/platform/configs/m4062nhp/platform_manager.json
@@ -1,0 +1,723 @@
+{
+  "___COMMENT___": "Platform Manager Configuration for M4062nhp with a cf2",
+  "platformName": "M4062NHP",
+  "rootPmUnitName": "SCM",
+  "rootSlotType": "SCM_SLOT",
+  "slotTypeConfigs": {
+    "SCM_SLOT": {
+      "numOutgoingI2cBuses": 10,
+      "idpromConfig": {
+        "busName": "CPU_BUS@3",
+        "address": "0x50",
+        "kernelDeviceName": "24c64"
+      },
+      "pmUnitName": "SCM"
+    },
+    "SMB_SLOT": {
+      "numOutgoingI2cBuses": 0,
+      "pmUnitName": "SMB"
+    }
+  },
+  "pmUnitConfigs": {
+    "SCM": {
+      "pluggedInSlotType": "SCM_SLOT",
+      "pciDeviceConfigs": [
+        {
+          "pmUnitScopedName": "SCM_IOB",
+          "vendorId": "0x10ee",
+          "deviceId": "0x7016",
+          "subSystemVendorId": "0x20a4",
+          "subSystemDeviceId": "0x0001",
+          "i2cAdapterBlockConfigs": [
+            {
+              "pmUnitScopedNamePrefix": "SCM_IOB_I2C",
+              "deviceName": "i2c_master",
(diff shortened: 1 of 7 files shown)


fboss/platform/configs/m4062nhp/platform_manager.json
```diff
@@ -0,0 +1,723 @@
...
+  "rootSlotType": "SCM_SLOT",
+  "slotTypeConfigs": {
+    "SCM_SLOT": {
+      "numOutgoingI2cBuses": 10,
+      "idpromConfig": {
+        "busName": "CPU_BUS@3",
+        "address": "0x50",
+        "kernelDeviceName": "24c64"
+      },
+      "pmUnitName": "SCM"
```
somasun: You shouldn't need this field, if the IDPROM in SCM_SLOT has the right ProductName.

fboss/platform/configs/m4062nhp/platform_manager.json
```diff
@@ -0,0 +1,723 @@
...
+    },
+    "SMB_SLOT": {
+      "numOutgoingI2cBuses": 0,
```
somasun, line 16: is this really 0 ?

somasun (changes requested):

fboss/platform/configs/m4062nhp/platform_manager.json
anna-nexthop: This is consistent with a sampling of the other platform_manager.json files on other systems and what is the downside of being explicit?

fboss/platform/configs/m4062nhp/platform_manager.json
anna-nexthop: yes

fboss/platform/configs/m4062nhp/platform_manager.json
somasun: > This is consistent with a sampling of the other platform_manager.json files 

We need to fix them.

> what is the downside of being explicit?

The code uses the ProductName in the IDPROM as the primary means to figure out which PmUnitConfig to load. The pmUnitName field was invented as a fallback for cases where the slots/PmUnits dont have an IDPROM.  The benefit of leaving the code to figure out based on the ProductName of the IDPROM is that it can resolve the right PmUnitConfig at Runtime. Say there are two types of SCMs - SCM1 and SCM2 (hypothetically), the resolution will happen at run time based on what is read from the IDPROM.

anna-nexthop: Removed (though I mistyped the commit message, should be removed from the SCM_SLOT)

PR #1334, opened 2026-06-25
[Accton][wedge800cact] Fix warmboot VerifyHostToQueueMappingClassID failure

BrandonCheng0121:
Regarding the T1 warmboot.AgentQueuePerHostL2Test.VerifyHostToQueueMappingClassID failure: the issue is a warmboot stats race condition.
The test gets stale cached ACL stats within milliseconds of warmboot completing because the background stats thread updates them only every 1 second.

Solution:
Explicitly call updateStats() in SwSwitch::initialConfigApplied() after warmboot completes to immediately sync hardware stats to the software cache.

**Pre-submission checklist**
- [x] `pre-commit run --files fboss/agent/SwSwitch.cpp fboss/agent/test/agent_hw_tests/AgentQueuePerHostL2Tests.cpp`

# Summary
After warmboot, tests immediately reading stats would get stale cached values because the background thread updates only every 1s.

# Test Plan
Test command:
`for i in {1..10}; do echo "=== run $i attempts ==="; time ./bin/run_test.py sai_agent --agent-run-mode mono --filter=AgentQueuePerHostL2Test.VerifyHostToQueueMappingClassID --skip-known-bad-tests "leaba/25.11.4210/25.11.4210/graphene202x" --enable-production-features g202x --config  /opt/fboss/share/hw_test_configs/wedge800cact.agent.materialized_JSON --fruid-path /home/Go_FBOSS_Test/W800CA-Fix/./fboss-configs/fboss/oss/scripts/run_configs/fruid.json --mgmt-if eth0 --platform_mapping_override_path /home/Go_FBOSS_Test/W800CA-Fix/./fboss-configs/fboss/lib/platform_mapping_v2/generated_platform_mappings/wedge800cact_platform_mapping-2026-0418-v0.7-honglim_20260409-del_pie.json  2>&1 | tee 463e_W800CACT_VerifyHostToQueueMappingClassID_DUT35_DVT1_v5_$(date '+%Y-%m-%d-%H:%M')_$i.log; done`

The test consistently passes for 10 consecutive runs.

### DIFF
--- a/fboss/agent/SwSwitch.cpp
+++ b/fboss/agent/SwSwitch.cpp
@@ -1649,6 +1649,20 @@ void SwSwitch::initialConfigApplied(
   sendNeighborSolicitationForConfiguredInterfaces("warm boot");
   sendArpRequestForConfiguredInterfaces("warm boot");
 
+  // After warmboot completes, immediately synchronize hardware stats to
+  // software cache (hwSwitchStats_). This ensures that tests reading stats
+  // immediately after warmboot get correct values instead of stale cached
+  // data. The background stats thread normally updates cache every 1 second,
+  // but tests may read stats within milliseconds of warmboot completing.
+
+  try {
+    XLOG(DBG2) << "Warm boot: synchronizing hardware stats to software cache";
+    updateStats();
+    XLOG(DBG2) << "Warm boot: stats synchronization completed";
+  } catch (const std::exception& ex) {
+    XLOG(ERR) << "Warm boot: stats synchronization failed: " << ex.what();
+  }
+
   if (flags_ & SwitchFlags::PUBLISH_STATS) {
     stats()->switchConfiguredMs(
         duration_cast<std::chrono::milliseconds>(
--- a/fboss/agent/test/agent_hw_tests/AgentQueuePerHostL2Tests.cpp
+++ b/fboss/agent/test/agent_hw_tests/AgentQueuePerHostL2Tests.cpp
@@ -47,6 +47,8 @@ class AgentQueuePerHostL2Test : public AgentHwTest {
     auto ttlCounterName = utility::getQueuePerHostTtlCounterName();
 
     auto statBefore = utility::getAclInOutPackets(getSw(), ttlCounterName);
+    XLOG(DBG2) << "ACL counter '" << ttlCounterName
+               << "' statBefore = " << statBefore;
 
     std::map<int, int64_t> beforeQueueOutPkts;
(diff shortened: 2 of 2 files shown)


BrandonCheng0121:
@Tianyu-Meta
Please take a look at this PR, thanks.

fboss/agent/SwSwitch.cpp
```diff
@@ -1649,6 +1649,20 @@ void SwSwitch::initialConfigApplied(
...
 
+  // After warmboot completes, immediately synchronize hardware stats to
+  // software cache (hwSwitchStats_). This ensures that tests reading stats
+  // immediately after warmboot get correct values instead of stale cached
+  // data. The background stats thread normally updates cache every 1 second,
+  // but tests may read stats within milliseconds of warmboot completing.
+
+  try {
+    XLOG(DBG2) << "Warm boot: synchronizing hardware stats to software cache";
+    updateStats();
```
vvasavada-nexthop, line 1660: Would this work in multi switch mode?
It seems `SwSwitch::updateSwitch` has this check and `updateHwSwitchStats` belongs under that check (mono mode).


```
  if (!isRunModeMultiSwitch()) {
    ...
    monoHwSwitchHandler->getHwStats(hwStats);
    hwStats.teFlowStats() = getTeFlowStats();
    updateHwSwitchStats(0 /*switchIndex*/, std::move(hwStats)); <-------
  }
```

vvasavada-nexthop (commented):
https://github.com/facebook/fboss/pull/1508 also tries to fix the same issue in different way.

fboss/agent/SwSwitch.cpp
BrandonCheng0121: The full pipeline is: monoHwSwitchHandler->updateStats() (read from HW) → getHwStats() (extract) → updateHwSwitchStats() (store in cache). 
This entire pipeline only runs inside SwSwitch::updateStats(), which the background UpdateStatsThread calls every 1 second. 
**After warmboot completes, the cache still holds stale pre-warmboot data until the background thread fires its next cycle.** The patch forces one immediate full pipeline execution so tests don't race against the 1-second interval.

We've only tested this in mono mode. The fix is effectively mono-mode only. 

Have you been able to test this in both mono and multi-switch mode on your side with #1508?

vvasavada-nexthop: Hi Brandon, I have verified that the test passes in mono mode as well with #1508 .

PR #1338, opened 2026-06-26
[Celestica] Tahansb: fan_service: Fixing OTP issues by optimizing the OTP algorithm

QiuyunXie:
**Pre-submission checklist**
- [ ] `pre-commit run`
image

# Summary

Meta reported a TH6 PCIe link issue, which, after debugging, was found to be caused by an SW OTP. It appears that during the TH6 SDK/agent initialization process, the TH6_TEMP sensor reads an abnormal, untrustworthy temperature jump. This anomalous reading value should not be used to trigger OTP. 
The OTP algorithm has been optimized. The previous algorithm, which calculated the average of _windowsize_ TH6 temperature samples to determine OTP, was a bit unreasonable. When an abnormally high temperature value is read within the window size, the calculated average will still remain high enough to exceed the overtemp threshold. It has been replaced with an algorithm that requires the TH6 temperature to exceed the overtempThreshold for _windowsize_ consecutive samples before triggering OTP.

# Test Plan

Continuously execute the agent initialization stress test while simultaneously running platform_manager, sensor_service, and fan_service.
Our SVT team has been running the stress test continuously for a week, and the issue has not been reproduced.

### DIFF
--- a/fboss/platform/configs/tahansb800bc/fan_service.json
+++ b/fboss/platform/configs/tahansb800bc/fan_service.json
@@ -20,8 +20,8 @@
     "conditions": [
       {
         "sensorName": "SMB_U2_TH6_TEMP",
-        "overtempThreshold": 120,
-        "slidingWindowSize": 10
+        "overtempThreshold": 103,
+        "slidingWindowSize": 2
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


QiuyunXie:
Close this PR, will raise new PRs to fix this issue.

PR #1297, opened 2026-06-15
[Nexthop][fboss2-dev] Add fboss2-dev config arp proactive subcommand

vybhav-nexthop:
**Pre-submission checklist**
- [x] I've ran the linters locally and fixed lint errors related to the files I modified in this PR. You can install the linters by running `pip install -r requirements-dev.txt && pre-commit install`
- [x] `pre-commit run`

## Summary

- Adds `fboss2-dev config arp proactive [enabled|disabled]` to the existing `config arp` command family
- Mutates `SwitchConfig.proactiveArp` (declared in `switch_config.thrift`) and persists via `HITLESS` config save — no agent restart required
- `proactiveArp` is not yet consumed by the agent (`ApplyThriftConfig.cpp` has no read, `switch_state.thrift` has no corresponding field); the CLI is wired ahead of agent support. A code comment flags this explicitly.

## Changed files

| File | Change |
|------|--------|
| `CmdConfigArp.cpp` | Add `proactive` to valid attrs; parse `enabled`/`disabled`; dispatch to `swConfig.proactiveArp()` |
| `CmdConfigArp.h` | Add `getBoolValue()` / `boolValue_`; update `addCliArg()` help text to mention `proactive` |
| `test/config/CmdConfigArpTest.cpp` | Unit tests for arg validation + `queryClient` round-trip for `proactive` |
| `test/integration_test/ConfigArpTest.cpp` | Integration test: set/commit/verify round-trip via running agent config |
| `test/integration_test/Fboss2IntegrationTest.h` | Add two-arg `getSwConfigField(field, default)` overload + private `getSwConfigFieldOpt<T>` helper for optional/defaulted config fields |

## `Fboss2IntegrationTest.h` refactor

`icmpV4UnavailableSrcAddress` is optional in the agent config — devices that have never configured it have no entry in the JSON. The integration test needs to read the field with a fallback to the agent's implicit default (`192.0.0.8`) when absent.

To support this without duplicating the type-dispatch logic (`bool`/`int`/`string`/`double`), the existing `getSwConfigField` is refactored:
- A private `getSwConfigFieldOpt<T>` helper does the actual field lookup and type conversion, returning `std::optional<T>`.
- The original single-arg `getSwConfigField` delegates to it and throws when `nullopt` (required fields — existing callers unaffected).
- A new two-arg overload delegates to it and returns the caller-supplied default when `nullopt` (optional fields).

## Integration Test Results — NH-4010-F

[...]

PR #1343, opened 2026-06-29
[Nexthop][fboss2-dev] Add fboss2-dev config hostname CLI

vybhav-nexthop:
**Pre-submission checklist**
- [X] `pre-commit run`

## Summary

Adds `fboss2-dev config switch hostname <name>`, which sets the optional
`sw.hostname` field in the agent switch config.

The command is nested under the `config switch` command group (introduced by
the admin-distance command) rather than at the top level, grouping
switch-level settings that do not belong to a specific network object.

- **Action level**: HITLESS — applied via
  `ThriftConfigApplier::updateSwitchSettings` (`newSwitchSettings->setHostname`);
  no SAI `ChangeProhibited` guard.
- **Upsert**: if `hostname` is already set to the requested value, the
  command is a no-op.
- **Validation**: the hostname argument is validated against RFC 952/1123
  using RE2 (alphanumeric plus hyphens, max 63 chars per label, total max
  253 chars).
- **Field**: `switch_config.thrift:SwitchConfig.hostname` (optional string) —
  "Overrides the system hostname, useful in ICMP responses." When unset, the
  agent falls back to the OS hostname at config-apply time
  (`getLocalHostname()`), so the value is always populated in switch state.

### Unit tests (`CmdConfigHostname{,Absent}TestFixture`)

9 tests, all passing in CI: argument validation (valid, FQDN, empty args,
empty string, too-many args, invalid format), set-new hostname, already-set
no-op, and set-when-field-absent.

### Integration test (`ConfigHostnameTest`) — on DUT gold407

The integration test reads the current hostname from **switch state**
(`getSwitchSettingsField`, always populated) inside a try/catch that returns
`""` if absent, sets a test value, verifies it via the switch-state thrift
path, and restores the original hostname in `TearDown`. It does not read the
optional config field directly, so it does not need any `getSwConfigField`
default-value helper — `Fboss2IntegrationTest.h` is unchanged from upstream.

Results from the internal CLI Integration Test (DUT) run on 2026-07-03
(device gold407):

```
[----------] 2 tests from ConfigHostnameTest
[ RUN      ] ConfigHostnameTest.SetAndVerifyHostname
  Current hostname: 'gold407'
  [Step 2] Setting hostname to 'cli-e2e-test-hostname'...
  Running CLI command: config switch hostname cli-e2e-test-hostname
  Verified: hostname is 'cli-e2e-test-hostname'
  Running CLI command: config switch hostname gold407      (restore)
[       OK ] ConfigHostnameTest.SetAndVerifyHostname (173 ms)
[ RUN      ] ConfigHostnameTest.IdempotentSet
  Running CLI command: config switch hostname cli-e2e-idempotent-hostname
  Running CLI command: config switch hostname cli-e2e-idempotent-hostname
  Running CLI command: config switch hostname gold407      (restore)
[       OK ] ConfigHostnameTest.IdempotentSet (120 ms)
[----------] 2 tests from ConfigHostnameTest (294 ms total)
```

Ran as part of the 25-test fboss2 CLI integration suite; all 25 passed
(total 50.5 s).

## Review Findings

Pre-publication review (single-pass automated reviewer). One item raised —
unchecked `*config.sw()` dereference in `CmdConfigHostname.cpp` — reviewed
and dismissed: it is the established pattern across all `fboss2 config`
commands (e.g. `CmdConfigArp.cpp`, `CmdConfigCopp.cpp`); `sw` is reliably
present for a loaded agent config. No issues above threshold.

## Presubmission Checks

- **clang-format** (`--style=file --dry-run --Werror`): pass on all added/modified C++ sources.
- **pre-commit**: pass (no violations on changed files).


fboss/cli/fboss2/test/integration_test/ConfigHostnameTest.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+using namespace facebook::fboss;
+
+class ConfigHostnameTest : public Fboss2IntegrationTest {
+ protected:
+  std::string getHostname() const {
+    // hostname is an optional thrift field; on platforms (e.g. sim) where it is
+    // unset, it is omitted from the running config. Default to "" so SetUp's
+    // initial read does not throw before any hostname has been set.
+    return getSwConfigField<std::string>("hostname", "");
```
joseph5wu: Please consider to use switch_state to verify the config result:
https://github.com/facebook/fboss/blob/main/fboss/agent/switch_state.thrift#L478

fboss2_integration_test already has the following utility function:
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.h#L131

fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.h
```diff
@@ -356,6 +348,30 @@ class Fboss2IntegrationTest : public ::testing::Test {
   std::string findIpv6OnIntf(int intfId) const;
 
  private:
+  template <typename T>
```
joseph5wu: Please use https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.h#L131 to verify rather than introducing the new getSwConfigFieldOpt() function

joseph5wu (changes requested):
Consider to use CmdConfigSwitch to manage individual Switch Settings

fboss/cli/fboss2/test/integration_test/ConfigHostnameTest.cpp
vybhav-nexthop: the issue with these fields is that they are optional, therefore we pass a default as the second argument , so if the config field is not present in the agent.conf, we return the default in our tests . Similar modifications were done in these PRs -> (https://github.com/facebook/fboss/pull/1297,
https://github.com/facebook/fboss/pull/1268).

vybhav-nexthop (commented):
Thanks for the review. Addressed all three points in the latest commit:

1. **CmdConfigSwitch**: Introduced `CmdConfigSwitch` as a new parent command and moved `hostname` under it as `config switch hostname`.
2. **switch_state verification**: Integration test now uses `getSwitchSettingsField`/`verifySwitchSettingsField` (reading from `switchSettingsMap` in switch_state) instead of reading from switch_config.
3. **Removed `getSwConfigFieldOpt`**: Dropped the `getSwConfigField{Opt}()` helpers from `Fboss2IntegrationTest.h` entirely.

fboss/cli/fboss2/test/integration_test/ConfigHostnameTest.cpp
vybhav-nexthop: wanted to build this PR on top of those, therefore used that field, but in this case , its always set to some value

PR #1364, opened 2026-07-06
[Nexthop] [fboss2] show route --filter (addressFamily) and --clientID

hillol-nexthop:
# Summary

Add generic `--filter` support to `show route` together with a per-command `--clientID` option, in place of the dedicated `show route static [ip | ipv6 | mpls]` subcommands this PR originally introduced. Reworked per review: the CLI no longer derives a "protocol"/winning client itself; the per-client view is served by the agent's `getRouteTableByClient` thrift API.

```
fboss2 show route --clientID static_route
fboss2 show route --clientID INTERFACE_ROUTE
fboss2 --filter "addressFamily == ipv6" show route --clientID bgpd
fboss2 --filter "addressFamily == ipv4" show route
```

`--filter` is a global option and must precede the subcommand (the CLI has no CLI11 fallthrough); `--clientID` is a `show route` local option and follows it.

**How it works**

- `--clientID <name|number>`: when given, `show route` calls `getRouteTableByClient(i16)` — the agent resolves the per-client entry (`Route::getEntryForClient`) — otherwise it keeps calling `getRouteTable`. Both return `list<UnicastRoute>`, so `createModel` is shared. Values are any `ClientID` enum name (case-insensitive, via thrift `EnumUtils`) or a numeric client id; anything else is rejected with the list of accepted names (exit 1).
- `cli::RouteEntry` gains an `addressFamily` string field (`ipv4|ipv6`) so the thrift-reflection filter engine can act on it (`ALLOW_FILTERING = true`). Accepted values are restricted, so typos fail with `VALUE_ERROR` and the accepted list. Filterable keys: `networkAddress`, `overridenEcmpMode`, `addressFamily`.
- No `protocol` field: ClientIDs such as `STATIC_INTERNAL` are not routing protocols, and the previous default-admin-distance "winning client" heuristic was not accurate under reconfigured admin distances.
- Bonus fix: `Fboss2IntegrationTest` resets `CmdGlobalOptions` filter/aggregate state and clears `CmdLocalOptions` (new `clear()`) before each in-process CLI invocation — CLI11 leaves bound singleton members untouched when an option isn't passed, so a `--filter` or `--clientID` from one `runCli()` silently applied to every later CLI call in the test binary.

# Test Plan

**Unit tests** — `fboss2_cmd_test` (full suite passes). `CmdShowRouteTest` covers `addressFamily` population, valid filter keys, filter application (single term, `&&`, `||`), `parseClientId` (name / lowercase / numeric / invalid), and mocked `queryClient` dispatch: no option → `getRouteTable`; `--clientID static_route` → `getRouteTableByClient(STATIC_ROUTE)`; invalid → throws without any RPC. The pre-existing `AnnotatesBackupNextHops` and `wikiDocHooks` cases are kept.

**Integration test** — `ShowRouteFilterTest.FilterRoutes` injects static routes (v4 nexthop, null0, cpu, v6 nexthop) plus a connected subnet into the agent config, restarts the agent, and verifies `--clientID static_route` / `--clientID INTERFACE_ROUTE` / `--clientID 2` / `addressFamily` filters / their intersection / invalid `--clientID` rejection end to end. Passes against a live agent (fake-SAI simulator and a lab device).

**Manual verification**:

```
$ fboss2 show route --help | grep clientID
  --clientID TEXT   Only show routes from this client (ClientID name, e.g. BGPD, STATIC_ROUTE, INTERFACE_ROUTE, or its numeric value)

$ fboss2 show route --clientID static
Invalid argument: Invalid --clientID 'static'. Accepted values are: BGPD, STATIC_ROUTE, INTERFACE_ROUTE, LINKLOCAL_ROUTE, REMOTE_INTERFACE_ROUTE, STATIC_INTERNAL, OPENR, TE_AGENT or a numeric ClientID
```


joseph5wu:
@hillol-nexthop : 
- For show command, we usually recommend to --filter to "filter" a show result like:
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/show/port/CmdShowPort.h#L37

```
$ fboss2 --filter "id <= 10 || tcvrPresent == Absent" show port
 ID  Name      AdminState  LinkState  ActiveState  Transceiver  TcvrID  Speed  ProfileID                             HwLogicalPortId  Drained  PeerSwitchDrained  PeerPortDrainedOrDown  Errors  Core Id  Virtual device Id  Cable Len meters
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 1   eth1/1/1  Enabled     Up         --           Present      0       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  28               No       --                 --                     --      --       --                 --
 2   eth1/1/5  Enabled     Up         --           Present      0       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  29               No       --                 --                     --      --       --                 --
 3   eth1/2/1  Enabled     Up         --           Present      1       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  30               No       --                 --                     --      --       --                 --
 4   eth1/2/5  Enabled     Up         --           Present      1       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  31               No       --                 --                     --      --       --                 --
 5   eth1/3/1  Enabled     Up         --           Present      2       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  32               No       --                 --                     --      --       --                 --
 6   eth1/3/5  Enabled     Up         --           Present      2       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  33               No       --                 --                     --      --       --                 --
 7   eth1/4/1  Enabled     Up         --           Present      3       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  34               No       --                 --                     --      --       --                 --
 8   eth1/4/5  Enabled     Up         --           Present      3       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  35               No       --                 --                     --      --       --                 --
 9   eth1/5/1  Enabled     Up         --           Present      4       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  36               No       --                 --                     --      --       --                 --
 10  eth1/5/5  Enabled     Up         --           Present      4       400G   PROFILE_400G_4_PAM4_RS544X2N_OPTICAL  37               No       --                 --                     --      --       --                 --
```

Can you check whether you can implement the show static routes using --filter on the existing `fboss2 show route` command?

- Btw we also have specific mpls command for all the mpls routes since they're critical
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/show/mpls/CmdShowMplsRoute.h

Can you check whether you can do the filtering static in the existing CmdShowMplsRoute too?

fboss/cli/fboss2/commands/show/route/utils.cpp
```diff
@@ -15,6 +19,72 @@ bool isFpfEncoding(
...
+    case ClientID::BGPD:
+      return 20; // EBGP
+    case ClientID::STATIC_INTERNAL:
+      return 255; // MAX_ADMIN_DISTANCE
+  }
+  return 255;
+}
+} // namespace
+
+ClientID getBestClientId(const facebook::fboss::RouteDetails& entry) {
```
joseph5wu: I don't think this function is 100% accurate and I would prefer to use 
https://github.com/facebook/fboss/blob/main/fboss/agent/if/ctrl.thrift#L1115
which has a more accurate implementation of fetching route based on ClientID using this logic:
https://github.com/facebook/fboss/blob/main/fboss/agent/ThriftHandler.cpp#L2129

Therefore, if we can't just simply filter from `getRouteTableDetails` result, I prefer to use other api,
rather than introduce some services related logic in the cli.

IMO, we need to compromise by supporting:
```
// All ClientIDs
fboss2 --filter "XXX==YYY" show route

// Specific ClientID
fboss2 --filter "XXX==YYY" show route --clientID interface_route
```
1) Keep what you did to support filter but remove the `protocol` filed from model.thrift. Please don't use protocol any way, as you can see these ClientID is not necessary just routing protocol. It looks confusing when you use STATIC_INTERNAL as protocol

2) If user specify --clientID, then we call `getRouteTableByClient` otherwise uses `getRouteTable` thrift api. Both APIs return the same list<UnicastRoute> so the createModel should be reusable.

fboss/cli/fboss2/commands/show/route/utils.cpp
```diff
@@ -15,6 +19,72 @@ bool isFpfEncoding(
       facebook::bgp::nsf_policy::NsfTeWeightEncoding::Type::fpf_l2_encoding;
 }
 
+std::string getProtocolStr(ClientID clientId) {
```
joseph5wu: You can use the following thrift library:
```
  #include <thrift/lib/cpp/util/EnumUtils.h>

  auto name =
      apache::thrift::util::enumNameSafe(ClientID::BGPD);
  // "BGPD"

  For an integer client ID:

  auto name = apache::thrift::util::enumNameSafe(
      static_cast<ClientID>(clientId));
```

joseph5wu (changes requested):
I wouldn't recommend your `getBestClientId()` considering current agent already support thrift api: getRouteTableByClient()

fboss/cli/fboss2/commands/show/route/utils.cpp
hillol-nexthop: Done in the latest push — reworked as suggested: the `protocol` field and the admin-distance heuristic are gone; `show route` now takes a `--clientID` local option (any `ClientID` enum name via `EnumUtils`, case-insensitive, or a numeric id) and calls `getRouteTableByClient` when it is given, `getRouteTable` otherwise, so `createModel` is shared and only `addressFamily` remains as a new filterable field. Unit tests cover the dispatch and `parseClientId`; the integration test now exercises `--clientID static_route` / `INTERFACE_ROUTE` / `2` alongside the `addressFamily` filter.

fboss/cli/fboss2/commands/show/route/utils.cpp
hillol-nexthop: Done — `parseClientId` uses `apache::thrift::util::tryParseEnum` / `TEnumTraits<ClientID>::names` from `thrift/lib/cpp/util/EnumUtils.h`; the hand-written name table is gone.

hillol-nexthop:
Note: the `fboss-sim` check on 42c980f failed before building anything — getdeps could not download `binutils-2.43.tar.xz` from ftpmirror.gnu.org (HTTP 502, transient). I don't have permission to re-run the workflow; a re-run should clear it.

hillol-nexthop:
CI note: the earlier red checks (42c980f, 34abad4) died in getdeps before compiling — `binutils-2.43.tar.xz` was unreachable via ftpmirror.gnu.org (HTTP 502, ~08:58–12:28 UTC). I can't re-run workflows here, so 05e50b6 is the same change re-pushed after the mirror recovered.

PR #1370, opened 2026-07-08
[Nexthop][fboss2-dev] Add fboss2-dev config acl CLI subcommands

benoit-nexthop:
## Summary

Adds two new subcommands under `config acl` for configuring ACL table groups and tables:

- **`config acl table-group <group-name> stage <stage-value>`** — sets `AclTableGroup.stage` (enum: `ingress`, `ingress-macsec`, `egress-macsec`, `ingress-post-lookup`, or numeric 0–3)
- **`config acl table <group-name> <table-name> priority <value>`** — sets `AclTable.priority` (non-negative i16)

The Thrift model has a two-level key structure (AclTableGroup → AclTable). Both commands target `sw.aclTableGroups` (field 56, the preferred form); the deprecated `aclTableGroup` (field 45) is not supported.

When `FLAGS_enable_acl_table_group` is set, both changes are applied at runtime via `processAclTableGroupDelta` in SaiSwitch with no warmboot-prohibited guard, so the CLI marks them **HITLESS**. When that flag is off (the default on some platforms today), `aclTableGroups` changes take a different code path in SaiSwitch; this PR only adds the CLI plumbing and was validated against a DUT with the flag off, so the flag-on runtime path is not covered by the included tests (see Test Plan).

Group/table lookup logic shared between the two subcommand handlers lives in `AclConfigUtils.{h,cpp}` rather than being duplicated between the `table` and `table-group` handlers.

### Unit tests (19/19 passing)

The seed config includes two table groups (the first with two tables), so tests can assert that mutating the targeted group/table leaves its siblings untouched, and dedicated tests target the non-first group/table by name — with only one group/table, or only ever mutating index 0, a `find_if` bug that matched "first element" instead of "match by name" would pass undetected. A separate fixture with no `aclTableGroups` at all covers the "no aclTableGroups configured" error branch shared by both handlers.

```
./fboss/oss/scripts/bazel.sh test //fboss/cli/fboss2/test/config:cmd_config_test \
    --test_filter='CmdConfigAcl*' --test_output=all

[ RUN      ] CmdConfigAclTableGroupTestFixture.argValidation_valid                    OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.argValidation_badArity                 OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.argValidation_unknownAttr              OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.argValidation_invalidStage             OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.setStageIngressPostLookup              OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.setStageNoopSameValue                  OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.setStageOnNonFirstGroup                OK
[ RUN      ] CmdConfigAclTableGroupTestFixture.groupNotFound                          OK
[ RUN      ] CmdConfigAclNoGroupsTestFixture.tableGroupNoAclTableGroupsConfigured     OK
[ RUN      ] CmdConfigAclTableTestFixture.argValidation_valid                         OK
[ RUN      ] CmdConfigAclTableTestFixture.argValidation_badArity                      OK
[ RUN      ] CmdConfigAclTableTestFixture.argValidation_unknownAttr                   OK
[ RUN      ] CmdConfigAclTableTestFixture.argValidation_nonInteger                    OK
[ RUN      ] CmdConfigAclTableTestFixture.argValidation_negative                      OK
[ RUN      ] CmdConfigAclTableTestFixture.setPriority                                 OK
[ RUN      ] CmdConfigAclTableTestFixture.setPriorityOnNonFirstTable                  OK
[ RUN      ] CmdConfigAclTableTestFixture.tableNotFound                              OK
[ RUN      ] CmdConfigAclTableTestFixture.groupNotFound                               OK
[ RUN      ] CmdConfigAclNoGroupsTestFixture.tableNoAclTableGroupsConfigured          OK
[  PASSED  ] 19 tests.
```

### Integration tests (NH-4010-F — skipped: DUT has no sw.aclTableGroups)

The NH-4010-F DUT used for testing uses a single flat ACL map without ACL table groups (`FLAGS_enable_acl_table_group` is off / field 56 absent). The integration tests detect this and skip gracefully:

```
[ SKIPPED ] ConfigAclTest.SetTableGroupStage — DUT has no sw.aclTableGroups (field 56)
[ SKIPPED ] ConfigAclTest.SetTablePriority   — DUT has no sw.aclTableGroups (field 56)
```

### Sample CLI transcript (NH-4010-F)

```
$ fboss2-dev config acl --help
Configure ACL table group and table settings
Subcommands:
  table                 Configure an ACL table within a group (e.g. set priority)
  table-group           Configure an ACL table group (e.g. set stage)

$ fboss2-dev config acl table-group --help
Positionals:
  acl_table_group_config TEXT x 3 REQUIRED
    <group-name> stage <stage-value> where <stage-value> is one of:
    ingress, ingress-macsec, egress-macsec, ingress-post-lookup (or numeric 0-3)

$ fboss2-dev config acl table --help
Positionals:
  acl_table_config TEXT x 4 REQUIRED
    <group-name> <table-name> priority <value>

$ fboss2-dev config acl
localhost: Thrift call failed: 'Incomplete command, please use 'table-group' or 'table' subcommand'
```


fboss/cli/fboss2/commands/config/acl/AclConfigUtils.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+namespace facebook::fboss::acl_utils {
+
+/*
+ * Returns a mutable reference to the AclTableGroup named `groupName` in
+ * swConfig.aclTableGroups() (field 56; the deprecated aclTableGroup, field
+ * 45, is not supported). Throws std::runtime_error if aclTableGroups is
+ * unset/empty, or if no group with that name exists.
+ */
+cfg::AclTableGroup& findAclTableGroupOrThrow(
```
joseph5wu, line 30: Can we consider to create a new acl table group if it doesn't exist?

fboss/cli/fboss2/commands/config/acl/table_group/CmdConfigAclTableGroup.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+cfg::AclStage parseAclStage(const std::string& s) {
+  // Try named string first (case-insensitive via tolower).
+  std::string lower = s;
+  std::transform(lower.begin(), lower.end(), lower.begin(), ::tolower);
+  auto it = kAclStageByName.find(lower);
+  if (it != kAclStageByName.end()) {
+    return it->second;
+  }
+  // Numeric fallback.
```
joseph5wu: I wouldn't recommend to use Numeric fallback cause they're not straightforward.
Let's just enforce the string values as the only options

fboss/cli/fboss2/commands/config/acl/table/CmdConfigAclTable.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+  auto& group =
+      acl_utils::findAclTableGroupOrThrow(swConfig, args.getGroupName());
+
+  auto& tables = *group.aclTables();
+  auto tableIt =
+      std::find_if(tables.begin(), tables.end(), [&](const cfg::AclTable& t) {
+        return *t.name() == args.getTableName();
+      });
+  if (tableIt == tables.end()) {
```
joseph5wu: Let's also support create an acl table if it doesn't exist

joseph5wu (changes requested):
Let's support create if not exist for both acl table group and acl table

vybhav-nexthop:
The branch has been reworked per review; this comment supersedes the PR description above.

## Summary

Four commands for the ACL table-group object model, all on `sw.aclTableGroups`
(field 56; the deprecated field-45 `aclTableGroup` is not supported):

```
config acl table-group <group-name> stage <stage-value>
config acl table <table-name> group <group-name> priority <value>
delete acl table <table-name>
delete acl rule <table-name> <rule-name>        (existing; cascade fixed)
```

Both config commands create the object when absent — including the very first
table-group on a config that has none, which is how a fresh box bootstraps ACL
config through the CLI. `<stage-value>` takes names only (`ingress`,
`ingress-macsec`, `egress-macsec`, `ingress-post-lookup`); per review, the
numeric fallback was dropped.

These commands write config the agent only reads when `enable_acl_table_group`
is on, so each one first checks that the agent config enables it and refuses
otherwise. 
```
"defaultCommandLineArgs": {
    "check_wb_handles": "true",
    "cleanup_probed_kernel_data": "true",
    "counter_refresh_interval": "0",
    "disable_neighbor_updates": "false",
    "ecmp_width": "320",
    "enable_1to1_intf_route_table_mapping": "true",
    "enable_acl_table_group": "true",               <----------- this field
    "enable_replayer": "true",
    "log_variable_name": "true",
    }
    "sw": {
    .....
 ````

Deleting a table removes its rules, so every traffic-policy matcher naming them
is stripped from both `dataPlaneTrafficPolicy` and
`cpuTrafficPolicy.trafficPolicy` — a leftover matcher fails in the next config and could silently re-attach to a later rule of the same name. Deleting the last ACL table in the config is refused (it would
silently drop every ACL and CoPP classification rule shipped by default). 
`delete acl rule` now ensures that the rule is removed from both dataplane and copp traffic policy

Refusals, each with a unit test: moving a group between stages (the agent keys
groups by stage — observed on a TH4, the commit restarts the agent and the
stage comes back unchanged), a second group on an occupied stage, a table name
already used in another group, and deleting the last table.

Two limitations worth knowing:

- `AclTable.priority` is stored in config, but today's parallel ACL table
  groups ignore it for hardware ordering
  (`SaiAclTableGroupManager` hardcodes the member priority), and changing it
  on an existing table recreates the table. Documented in the command help.
- Whether the ASIC supports a stage is not checked by the CLI; an unsupported
  one (e.g. `ingress-post-lookup` on TH3) fails at the agent when the config
  is applied.

### Why there is no `delete acl table-group`

`SaiAclTableGroupManager::removeAclTableGroup` is an unimplemented stub
(`// TODO(skhare) add support`), and `processAclTableGroupDelta` walks only the
new state's groups — so a deleted group's SAI objects would survive until an
agent restart, and recreating a group at the same stage then throws
`"attempted to add a duplicate aclTableGroup"`. The CLI side can follow once
the agent can tear a group down.

### The simulator now runs in table-group mode

`mono.conf` sets `enable_acl_table_group` and drops the deprecated field-45
block(sw.acls) (the agent prefers field 45 when both are present). With the flag on,
`ApplyThriftConfig` programs what `ConfigAclRuleTest` writes for the first
time — previously the agent read `sw.acls` (empty), so those tests only ever
asserted the CLI's config readback.

Two cases are removed: `SetEtherType` and `SetVlan`. The fake ASIC can't match
on those two fields, so the agent crashes when a test programs them. What they
verified — the CLI writing the field into config — is already covered by unit
tests. They can return once the agent gates the ether-type qualifier on
`HwAsic::Feature::ACL_ENTRY_ETHER_TYPE`; that's an agent-side change, not made
here. These tests started failing after the addition of the `enable_acl_table_group` flag in the mono.conf in the fake sai. 

## How to read this

Four commands ship here (grammar above). Where the 26 files go:

|                                                                 | files | lines     |
|-----------------------------------------------------------------|-------|-----------|
| shared helpers — `AclConfigUtils.{h,cpp}`                       | 2     | ~300      |
| the three commands — one `.h`/`.cpp` pair each                  | 6     | ~590      |
| tests                                                           | 4     | ~1000     |
| registration — `CmdListConfig.cpp`, 3 cmake lists, 3 BUCK lists | 7     | ~60       |
| refactor — `findAclTable` moved out of `AclRuleAttrs`, callers repointed | 5 | small |
| simulator config — `mono.conf` + 2 removed integration tests    | 2     | small     |

Three worth real review:

1. **`commands/config/acl/AclConfigUtils.{h,cpp}`** — every command above runs
   through these helpers.
2. **`commands/delete/acl/table/CmdDeleteAclTable.cpp`** — the delete cascade
   into both traffic policies, and the last-table refusal.
3. **`commands/config/acl/table_group/CmdConfigAclTableGroup.cpp`** — stage
   parsing and the stage-move refusal.

Then: `table` and `table_group` share their skeleton, so skim whichever you
read second. The refusals only look excessive until "the simulator now runs in
table-group mode" above — the flag is off by default, so much of the code is
guarding that boundary.

## Test Plan

Unit (cmake `cmd_config_test` target): **904/904 pass** — the flag-off refusal,
first-group bootstrap, stage-names-only validation, the stage-move refusal,
both duplicate-name guards, delete cascades into both policies, and the
last-table refusal.

Integration: CI's `fboss2-Fake-SAI-Integration-Test` runs the full fboss2 suite
against the simulator in table-group mode; `ConfigAclTest` drives
create → commit → delete → commit against the live agent.

PR #1373, opened 2026-07-09
[Nexthop][m4062nhp] BSP platform mapping

anna-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Add the m4062nhp BSP mapping CSV input and the generated M4062nhpBspPlatformMapping class, and wire the platform into the BSP mapping generator, its unit test, the singleton registry, and the Buck/CMake builds.

# Test Plan

Builds and the associated unit test passes.


fboss/lib/bsp/bspmapping/input/M4062nhp_BspMapping.csv
```diff
@@ -0,0 +1,129 @@
+TcvrId,TcvrLaneIdList,PimId,AccessControllerId,AccessControlType,ResetPath,ResetMask,ResetHoldHi,PresentPath,PresentMask,PresentHoldHi,IoControllerId,IoControlType,IoPath,LedId,LedBluePath,LedYellowPath
```
harshitgulati18, line 1: @lohithchittineni - Should we use XcvrLib from the beginning itself instead of the csv?

lohithchittineni: yes @anna-nexthop please refer to the documentation here. https://github.com/facebook/fboss/blob/main/docs/docs/developing/bsp_mapping.md
we will be deriving the bsp mappings from platform_manager.json moving forward in order to standardize the information in 1 location

anna-nexthop: @harshitgulati18 @lohithchittineni We will rework to the XcvrLib pattern in a follow-up PR as discussed with @Scott8440 live. Please take a look at the current changes as is.

PR #1379, opened 2026-07-13
[Nexthop][fboss2-dev] Add config/delete copp cpu-traffic-policy match action CLI

vybhav-nexthop:
# Summary

Adds a config/delete command pair for CPU-plane action fields in
`sw.cpuTrafficPolicy.trafficPolicy.matchToAction`:

```
fboss2-dev config copp cpu-traffic-policy match <matcher-name> action <action-type> <value>
fboss2-dev delete copp cpu-traffic-policy match <matcher-name> action <action-type>
```

Supported action types: `send-to-queue <queue-id>`, `counter <name>`, `set-tc <tc-value>`, `user-defined-trap <queue-id>`.

Each action type sets one field of `cfg::MatchAction` and applies to the packets the referenced ACL matches that are punted to the CPU:

- **`send-to-queue <queue-id>`** — steer the matched CPU-bound packets to a specific CPU queue. CPU queues are individually rate-limited/prioritized, so this is meant for isolating control traffic (e.g. BGP) from best-effort punts. (`QueueMatchAction.queueId`)
- **`counter <name>`** — attach a named packet/byte counter to the matched traffic. Observability only; no forwarding effect. (`MatchAction.counter`)
- **`set-tc <tc-value>`** — set the internal traffic class on the matched packets; the ASIC's TC→queue map then resolves the destination CPU queue. Today this shares the same hardware primitive as `send-to-queue` (the queue-id is programmed as the TC value in `SaiAclTableManager`). `tcValue` is a thrift `byte`, so the accepted range is `[0, 127]`. (`SetTcAction.tcValue`)
- **`user-defined-trap <queue-id>`** — bind the match to a dedicated SAI hostif user-defined trap on the given CPU queue, so the ACL rule takes precedence over the default hostif trap rules. Platform-gated on `SAI_USER_DEFINED_TRAP`. (`UserDefinedTrapAction.queueId`)

Both are applied at **HITLESS** level: matchToAction is processed through the ACL update path in `ApplyThriftConfig.cpp`, and no `ChangeProhibited` guards exist for these fields.

# Test Plan

## Unit tests

`//fboss/cli/fboss2/test/config:cmd_config_test` — full suite passes, including:

- 8 new `cpuTrafficPolicy*` tests in `CmdConfigCoppTest.cpp`: arg validation (arity, literal tokens, action types, value ranges) and queryClient behavior (add to existing matcher preserving other action fields, create new matcher entry, overwrite same action type, user-defined-trap).
- 8 delete tests in `CmdDeleteCoppCpuTrafficPolicyMatchActionTest.cpp`: arg validation plus delete of each action type, absent-field and absent-matcher handling.

## Integration test

`ConfigCoppCpuTrafficPolicyTest.addAndDeleteAction` on an NH-4010-F device exercises the full cycle: picks an ACL from the running config as matcher (one without a pre-existing action of the tested type); if the config has no usable ACL — the default on freshly bootstrapped devices (empty `AclTable1`) and the simulator — it injects a test-owned ACL (`fboss2-copp-test-acl`) through the config session and uses that. Then add → commit → verify present → delete → commit → verify gone.

Cleanup is snapshot-based: SetUp captures the running config before any mutation, and TearDown writes it back through the session (hitless commit) only if the committed config drifted — so the test restores the switch on pass or fail, on both the pre-existing-matcher and injected-ACL paths.

Post-run checks: no `fboss2-copp-test-acl` leftovers in the committed agent config, `fboss_sw_agent` NRestarts=0 across all commits.

# Review Findings

A multi-angle pre-publication review (8 finder passes + adversarial verification) ran on this diff. Two confirmed issues were fixed in this PR:

- **Matcher name `action` collision**: the delete command models the grammar as subcommand tree nodes, so CLI11 classifies a bare `action` token as the subcommand before filling the matcher-name positional — a matcher literally named `action` would be configurable but undeletable. The config command now rejects that name up front (with a unit test).
- **Empty-entry pruning on the already-absent path**: deleting an action field that was already absent returned early before the empty-`MatchAction` prune, so a pre-existing entry with no action fields (hand-edited config) survived. The prune now runs regardless, and the entry is removed (with a unit test).

Remaining low-severity notes, documented for reviewer awareness: the four action-type dispatch chains (validate/apply/delete/isValid) are maintained separately and each ends in a bare `else` for `user-defined-trap`; the integration test's injected ACL uses a dstIp qualifier that could be rejected on a hypothetical ACL table configured with a restricted qualifier set (current platforms use the full-set fallback).


fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
```diff
@@ -278,6 +370,52 @@ std::string applyReasonConfig(
...
+    swConfig.cpuTrafficPolicy() = cfg::CPUTrafficPolicyConfig{};
+  }
+  auto& policy = *swConfig.cpuTrafficPolicy();
+  if (!policy.trafficPolicy().has_value()) {
+    policy.trafficPolicy() = cfg::TrafficPolicyConfig{};
+  }
+  auto& matchToActions = *policy.trafficPolicy()->matchToAction();
+
+  const auto& name = args.getMatcherName();
+  auto it = copp_cpu_traffic_policy::findMatchToAction(matchToActions, name);
```
joseph5wu, line 386: Based on the current MatchToAction design:
https://github.com/facebook/fboss/blob/main/fboss/agent/switch_config.thrift#L852

You need to make sure the matcher name is one of the created acls.

Therefore, you need to search the acl entry with the mater name in current config to see whether such acl exists first before you can see the match to action for copp.

And thanks to https://github.com/facebook/fboss/pull/1183, this PR allows us to create a new acl and set it to dataplane traffic policy if it's matchAction acl.
Since you're working on control plane, can you check whether your PR here can work with PR#1183

I think if we need to set a new copp matchAction acl, it should be:
```
// Create an acl table group if not exist
fboss2-dev config acl table-group <group-name> stage <stage-value>
// Create an acl table if not exist
fboss2-dev config acl table <group-name> <table-name> priority <value>
// Create an acl (regardless of data plane or control plane, they can be in the same table)
fboss2-dev config acl rule <table> <rule> <attr> <value> [<ttl-mask>]

// Then the MatchAction(We should expect acl name to be unique regardless of table or group)
fboss2-dev config acl rule <rule> action <subattr> [<value>] -> Go to dataPlaneTrafficPolicy
fboss2-dev config acl rule <rule> copp action <subattr> [<value>] -> Go to CPUTrafficPolicy
```
Let me know what you think? I'm also okay if we want to move copp to the front just like existing `config copp cpu-queue` cli:
```
fboss2-dev config coop acl rule <rule> action <subattr> [<value>] -> Go to CPUTrafficPolicy
```

What's more, I think PR 1183 and PR 1379 have lots of redundant code about how to manage the MatchAction, maybe we can consider reuse lots of similar logic

joseph5wu (changes requested):
As I mentioned in our group chat, I think we need Nexthop to discuss internally to come up with a better design for managing data plane and control plane ACLs.
https://chat.google.com/room/AAQATNypAsc/aUbuxxKNEOs/aUbuxxKNEOs?cls=10

I think there're some overlap b.w this PR and https://github.com/facebook/fboss/pull/1183, please let me know what's your thought

vybhav-nexthop:
will close this on an account of it being handled in #1550

PR #1386, opened 2026-07-15
[Nexthop][fboss2-dev] Add lookup-class CLI to config/delete interface commands

vybhav-nexthop:
# Summary

This PR adds a new attribute, `lookup-class`, to the existing `fboss2-dev config interface` and `delete interface` command families:

```
fboss2-dev config interface <interface-name> lookup-class <id>[,<id>...]   # replace the port's lookup-class list
fboss2-dev delete interface <interface-name> lookup-class                  # clear the port's lookup classes
```

**What is a lookup class?** Each front-panel port in the agent config carries a list field, `Port.lookupClasses` (a `list<AclLookupClass>` in `switch_config.thrift`). When this list is non-empty, the agent tags every neighbor (host) learned on that port with one of the class IDs from the list, and ACLs then match on those class IDs to steer each host's traffic into a specific egress queue. This is the mechanism behind the "queue-per-host" feature used on multi-host NICs (MH-NIC): several hosts share one physical switch port, and lookup classes let the switch give each host its own queue so one host cannot starve the others.

**How the commands behave:**

- **Set** replaces whatever list the port currently has with the given list. Real queue-per-host deployments configure the full class pool (e.g. `10,11,12,13,14`) so the agent can spread hosts across five queues; a single id is also accepted.
- **Delete** resets the list to empty (`[]`), the field's default ("no per-host classification on this port"). Deleting individual ids from the list is intentionally not supported — clear-all covers the use case; an id-list delete form can be added later if a real need appears.
- **Validation**: only the queue-per-host classes (`CLASS_QUEUE_PER_HOST_QUEUE_0..9`, ids 10-19) are accepted — the other `AclLookupClass` members (`CLASS_DROP`, `DST_CLASS_L3_LOCAL_*`, ...) are assigned by the agent itself and are rejected as reserved. Ids and enum names (case-insensitive) are both accepted. Unknown values, duplicate ids, and empty list slots (`10,,11`) are all rejected; the error lists every valid id and its name.
- **No traffic impact when applied (HITLESS)**: the change is staged in a config session and, on commit, flows through the agent's normal port-settings update path. Verified on hardware: committing reloads the config for `fboss_sw_agent` without restarting the agents or disrupting the port.

Changed files:

| File | Change |
|------|--------|
| `commands/config/interface/CmdConfigInterface.cpp` | `lookup-class` attribute; id-list parse/validate (enum check, duplicates, empty slots) + apply |
| `commands/delete/interface/CmdDeleteInterface.{h,cpp}` | `lookup-class` valueless delete attribute (clears `lookupClasses`) |
| `test/config/CmdConfigInterfaceTest.cpp` | 13 unit tests (set path: single id, lists, names, error cases) |
| `test/config/CmdDeleteInterfaceTest.cpp` | 3 unit tests (delete path: clear, idempotent, multi-port) |
| `test/integration_test/ConfigInterfaceLookupClassTest.cpp` | 7 end-to-end tests, all verifying the running config (4 rejection + 3 commit-path) |
| `cmake/CliFboss2TestIntegrationTest.cmake`, `test/integration_test/BUCK` | wire the new test file into both build systems |

# Test Plan

- Unit tests: set path (parse, apply single id and comma-separated lists, replacing an existing list, multiple ports, rejecting non-numeric / out-of-range / duplicate / empty-slot input) and delete path (clear all, idempotent when already empty, multiple ports); full `cmd_config_test` suite run to check for regressions
- Integration tests against a real switch (an NH-4010-F lab device): every test verifies the port's `lookupClasses` in the agent running config — rejected inputs assert it is unchanged; happy-path tests pick an eth port with an empty lookupClasses baseline, commit the staged change and verify the port's `lookupClasses` in the agent running config, then delete + commit and verify it clears — the device is restored to its original state; `fboss_sw_agent` / `fboss_hw_agent@0` stayed active through all commits (hitless)

Sample usage, including the committed result in the live agent config:

```
$ fboss2-dev config interface eth1/1/1 lookup-class 10,11,12,13,14
Successfully configured interface(s) eth1/1/1: lookup-class=10,11,12,13,14
Config session committed successfully and config reloaded for fboss_sw_agent.

# /etc/coop/agent.conf now has eth1/1/1 lookupClasses [10, 11, 12, 13, 14]

$ fboss2-dev delete interface eth1/1/1 lookup-class
Successfully reset attribute 'lookup-class' for interface(s): eth1/1/1

# eth1/1/1 lookupClasses now [] (back to default)

$ fboss2-dev config interface eth1/1/1 lookup-class 999
Invalid lookup-class value '999'. Valid values: 10 (CLASS_QUEUE_PER_HOST_QUEUE_0),
11 (CLASS_QUEUE_PER_HOST_QUEUE_1), ... 19 (CLASS_QUEUE_PER_HOST_QUEUE_9)

$ fboss2-dev config interface eth1/1/1 lookup-class 9
Invalid lookup-class value '9': CLASS_DROP is reserved for agent use. Valid values: 10 (CLASS_QUEUE_PER_HOST_QUEUE_0), ...

$ fboss2-dev delete interface eth1/1/1 lookup-class 11,13
Unknown delete attribute '11,13'. Valid attributes are: loopback-mode, lookup-class, ...
```

### DIFF
--- a/fboss/cli/fboss2/test/integration_test/ConfigInterfaceLookupClassTest.cpp
+++ b/fboss/cli/fboss2/test/integration_test/ConfigInterfaceLookupClassTest.cpp
@@ -27,11 +27,10 @@ using namespace facebook::fboss;
 class ConfigInterfaceLookupClassTest : public Fboss2IntegrationTest {
  protected:
   void TearDown() override {
-    // Never leave a staged lookup-class change behind for the next test.
-    discardSession();
-    // A test that failed between its set-commit and clear-commit would leave
-    // the committed class list on the port; clear it (best-effort) so the
-    // device is restored even on failure.
+    // If the test failed after committing a lookup-class change but before
+    // clearing it, best-effort restore the port so a shared DUT is left
+    // clean. (A committed change survives discardSession(), which the base
+    // SetUp already runs to clear any staged session before the next test.)
     if (!committedPort_.empty()) {
       try {
         if (!portLookupClasses(getRunningConfig(), committedPort_).empty()) {
@@ -39,11 +38,8 @@ class ConfigInterfaceLookupClassTest : public Fboss2IntegrationTest {
           commitConfig();
         }
       } catch (const std::exception&) {
-        // Restoration is best-effort; do not mask the test result.
+        // Best-effort; do not mask the test result.
       }
-      // A failed restore may itself have staged a delete; never leave a
-      // half-staged session behind for whatever runs next.
-      discardSession();
     }
     Fboss2IntegrationTest::TearDown();
   }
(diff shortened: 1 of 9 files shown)


fboss/cli/fboss2/test/integration_test/ConfigInterfaceLookupClassTest.cpp
```diff
@@ -0,0 +1,234 @@
...
+    const auto& sw = config["sw"];
+    if (!sw.isObject() || !sw.count("ports")) {
+      return "";
+    }
+    for (const auto& port : sw["ports"]) {
+      if (!port.count("name")) {
+        continue;
+      }
+      auto name = port["name"].asString();
+      if (name.rfind("eth", 0) != 0) {
```
joseph5wu: What's the reason to limit port name to start with `eth`?

fboss/cli/fboss2/test/integration_test/ConfigInterfaceLookupClassTest.cpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+    return "";
+  }
+
+  // Port a test committed a lookup-class list on; cleared in TearDown if the
+  // test did not restore it itself.
+  std::string committedPort_;
+};
+
+// A non-numeric value is rejected before anything is staged; the running
```
joseph5wu: I think we need to reconsider how to design our test cases here.
fboss2_integration_tests runs on a real hardware environment as an integration test, hence it's usually heavier compared to unit test.
For most of the following tests(like the reject ones), the cli runs are usually rejected in the fboss2 config implementation. Therefore, I think these tests should be included in the unit test(tes/config); and the one you really want to verify program to the hardware should be included in the integration tests

joseph5wu (changes requested):
The implementation looks solid to me.
But please make sure `fboss2_integration_tests` are used to verify the logic that actually program to fboss agent; and you should consider to write "reject" test cases in unit test format.

fboss/cli/fboss2/test/integration_test/ConfigInterfaceLookupClassTest.cpp
vybhav-nexthop: Good point — I've removed the eth filter. The reasoning: only INTERFACE_PORT ports can have neighbors attached (which is what makes lookup classes meaningful, since they classify neighbors), and those are typically the eth*-named ports. But neither the CLI nor the agent mandates a port type for lookupClasses, so the test doesn't need to either — it now just picks the first port with an empty lookupClasses list as its baseline.

fboss/cli/fboss2/test/integration_test/ConfigInterfaceLookupClassTest.cpp
vybhav-nexthop: have covered all parts in the unit tests, maintain one test in the integration

PR #1345, opened 2026-06-29
[Nexthop][fboss2-dev] Add fboss2-dev config protocol bgp global command

hillol-nexthop:
> **Stack (bottom → top):** **#1345 (this PR)** → #1391 → #1395 → #1401 → #1475 → #1477 → #1487 → #1488 → #1489 → #1490 → #1491

## Summary

Add `fboss2-dev config protocol bgp global <attr> <value>` on top of the
BGP-aware `ConfigSession`. It edits the typed `bgp::thrift::BgpConfig` via
`ConfigSession::getBgpConfig()`/`saveBgpConfig()` — the whole-config,
scope-agnostic typed API (no global/peer/peer-group special-casing in
`ConfigSession`).

- Collapses the 10 per-attribute global command classes into one dispatcher;
  rejects `cluster-id` (no `BgpConfig` field) instead of writing dead config;
  bounds `switch-limit` / `max_golden_vips` so out-of-range values aren't
  truncated.
- Integration tests: `ConfigBgpGlobalTest` (each attribute set+commit, verified
  in the promoted `/etc/coop/bgpcpp/bgpcpp.conf`) and `ConfigBgpSessionTest`
  (clear/diff/commit/rollback + an unchanged-config no-op-restart regression),
  sharing `ConfigBgpTestBase`.

## Sample usage

```
$ fboss2-dev config protocol bgp global router-id 10.0.0.1
Successfully set BGP router-id to: 10.0.0.1
Config saved to: ~/.fboss2/bgp_config.json

$ fboss2-dev config protocol bgp global local-asn 65000
Successfully set BGP local-asn to: 65000
Config saved to: ~/.fboss2/bgp_config.json

$ fboss2-dev config protocol bgp global hold-time 90
Successfully set BGP hold-time to: 90 seconds
Config saved to: ~/.fboss2/bgp_config.json

$ fboss2-dev config protocol bgp global switch-limit-overload-protection-mode APPLY_GOLDEN_PREFIX_POLICY
Successfully set switch_limit_config overload_protection_mode to: 2
Config saved to: ~/.fboss2/bgp_config.json

# Rejected before anything is staged:
$ fboss2-dev config protocol bgp global router-id not-an-ip
Error: Invalid router-id address 'not-an-ip'

$ fboss2-dev config protocol bgp global router-id 2001:db8::1
Error: router-id requires an IPv4 address, got '2001:db8::1'

$ fboss2-dev config protocol bgp global switch-limit-overload-protection-mode 999
[...]

PR #1391, opened 2026-07-17
[Nexthop][fboss2-dev] fboss2 config/delete protocol bgp neighbor commands

hillol-nexthop:
> **Stacked on #1345** (`config protocol bgp global`). This PR's branch contains the whole stack below it, so the diff shows those commits too — please review only the `config/delete protocol bgp neighbor` commit(s) here. Merge the lower PRs first.
>
> **Stack (bottom → top):** #1345 → **#1391 (this PR)** → #1395 → #1401 → #1475 → #1477 → #1487 → #1488 → #1489 → #1490 → #1491

## Summary

Adds `fboss2-dev config protocol bgp neighbor <ip-address> [<attribute> <value> ...]` as a single dispatcher mirroring the `global` command: the neighbor address is the first positional token, one- or two-token attributes (`timers hold-time`, `add-path send`, ...) are matched longest-prefix-first, and handlers mutate the typed `bgp::thrift::BgpPeer` via `ConfigSession::getBgpConfig()` / `saveBgpConfig()`. Also adds `fboss2-dev delete protocol bgp neighbor <ip-address>`. The 26 old per-attribute `peer` command classes and the now-dead per-peer `folly::dynamic BgpConfigSession` API are deleted.

| Attribute | `BgpPeer` field |
|---|---|
| `remote-asn` / `local-asn` | `remote_as_4_byte` / `local_as_4_byte` (4-byte-bounded) |
| `description`, `peer-tag`, `peer-group`, `peer-id`, `type` | `description`, `peer_tag`, `peer_group_name`, `peer_id`, `type` |
| `ingress-policy` / `egress-policy` | `ingress_policy_name` / `egress_policy_name` |
| `rr-client`, `confed-peer`, `redistribute-peer`, `enhanced-route-refresh`, `route-refresh`, `remove-private-as`, `enforce-first-as`, `next-hop-self` | matching bools |
| `passive <true\|false>` | `is_passive` (true = PASSIVE_ONLY, false = PASSIVE_ACTIVE in bgpd) |
| `ttl-security-hops <1-255>` | `ttl_security_hops` (bgpd's `kMin/kMaxTtlSecurityHops`; it throws at load outside that range) |
| `afi disable-ipv4-afi` / `disable-ipv6-afi` / `ipv4-over-ipv6-nh` | `disable_ipv4_afi` / `disable_ipv6_afi` / `v4_over_v6_nexthop` |
| `add-path send\|receive <bool>` | merged into the `AddPath` enum (RECEIVE/SEND/BOTH) |
| `bind-addr address`, `next-hop4`, `next-hop6` | `local_addr`, `next_hop4`, `next_hop6` (IP/family-validated) |
| `graceful-restart restart-time` / `stateful-ha` | `bgp_peer_timers.graceful_restart_seconds` / `enable_stateful_ha` |
| `timers hold-time` / `keepalive` / `out-delay` / `withdraw-unprog-delay` | `bgp_peer_timers.*` |
| `max-route pre-filter` / `post-filter` / `{pre,post}-warning-threshold` / `{pre,post}-warning-only` | `pre_filter` / `post_filter` RouteLimit fields |
| `link-bandwidth`, `advertise-lbw`, `receive-lbw` | `link_bandwidth_bps`, `advertise_link_bandwidth`, `receive_link_bandwidth` (enum-membership-validated) |

That is every `BgpPeer` field bgpd's `Config.cpp` reads. Deliberately not exposed (documented in the dispatcher): `enabled`, `router_port_id` and `bgp_peer_timers.graceful_restart_end_of_rib_seconds` — nothing in bgpd reads them, so staging them would persist dead config.

Bare `neighbor <ip>` creates the peer; attribute sets find-or-create it, keyed by the `folly::IPAddress`-normalized `peer_addr`. Prefix notation for passive sessions is accepted and host bits are masked (`10.0.0.1/24` → `10.0.0.0/24`), matching how bgpd keys dynamic peers, so equivalent spellings update/delete the same neighbor. `delete protocol bgp neighbor <ip>` removes the matching peer via the same shared normalize/find helpers (`BgpPeerAddr.h`); deleting an unknown neighbor errors without persisting anything.

### Behavior notes

- **Inherited values are preserved.** bgpd resolves `bgp_peer_timers`, `pre_filter` and `post_filter` per *struct* against the peer group, and `hold_time_seconds` / `keep_alive_seconds` are non-optional. Before the first attribute of such a struct is set on a peer, the struct is seeded from what the peer inherits today — the group's copy, else the global hold time with hold/3 keepalive — and only the requested field changes. For route limits with nothing to inherit, the seed is `{max_routes 0, warning_only false, warning_limit 0}`: bgpd applies no limit to a peer without a struct and reads `max_routes == 0` as unlimited, whereas the thrift default (12000) would have introduced a hard cap with session teardown that the operator never asked for.
- **Handlers edit a copy** of the peer and commit on success, so a rejected value never leaves a half-edited peer or a seeded struct behind.
- **add-path** merges into the *effective* value (the peer's, else the group's). Disabling a direction the group still enables is rejected with a message naming the group: `AddPath` has no disabled member and an unset peer field inherits the group, so it would otherwise be a no-op reported as success. Narrowing a group's `BOTH` to one direction is allowed.
- **Validation matches bgpd**: hold time `0` or `3..65535`, graceful-restart `0..4095`, TTL security hops `1..255`; `link-bandwidth` accepts exactly what bgpd's `parseLinkBandwidthBps` accepts (`auto`, decimals, case-insensitive K/M/G). New peers are seeded with parseable any-addresses (`local_addr`/`next_hop4`/`next_hop6` = `0.0.0.0`/`::`) because bgpd constructs `folly::IPAddress` from these non-optional fields at load and aborts on empty strings.
- **Rejected instead of persisting dead config**: `peer-port` / `bind-addr port` (no per-peer port field; the daemon uses the global `listen_port`), `afi ipv4/ipv6-labeled-unicast` (no thrift field).
- **`--help`** lists every attribute with its value shape, generated from the dispatch table (`bgpNeighborAttributeHelp()` is the subcommand's CLI11 footer); tests assert every registered attribute appears and parses. Tab completion cannot see free-form attributes; registering them as CLI11 subcommands is a larger change left for a follow-up.
- **CLI11 token stealing fixed**: `positionals_at_end()` on the neighbor command stops parent-chain subcommand fallthrough from reclassifying attribute tokens that match sibling command names (`neighbor <ip> peer-group X` misparsed with `X` as the neighbor address).
- **`bgp_json_to_cli.py`** emits the neighbor grammar with every JSON-sourced value passed through `shlex.quote`; explicit zero timers/limits are kept; symbolic `add_path` enum names are handled; the deprecated i32 `remote_as`/`local_as` map to the 4-byte attribute; any JSON field with no CLI equivalent produces a `# WARNING: neighbor <addr>: field <f> has no CLI equivalent, not converted` comment line in the generated script (and on stderr) instead of being dropped silently. The shared bool/int/link-bandwidth/timer/route-limit tables are reused by the peer-group generator in #1395.
- Integration tests verify the committed neighbor config through the daemon's `TBgpService::getRunningConfig` RPC, proving bgpd parsed and adopted the promoted config rather than only checking the file the CLI wrote.

**Delete is idempotent.** Deleting an absent target is a success with a `Warning:` (nothing is saved), so a replayed script never fails on something already gone; the reference guards still return `Error:`.

## Sample usage

```
$ fboss2-dev config protocol bgp neighbor 2001:db8::99 remote-asn 65510
Successfully set remote-asn to: 65510 for neighbor 2001:db8::99
$ fboss2-dev config protocol bgp neighbor 2001:db8::99 timers hold-time 90
Successfully set timers hold-time to: 90 seconds for neighbor 2001:db8::99
$ fboss2-dev config protocol bgp neighbor 2001:db8::99 passive true
Successfully enabled passive for neighbor 2001:db8::99
$ fboss2-dev config protocol bgp neighbor 2001:db8::99 ttl-security-hops 300
Error: Invalid ttl-security-hops value '300'; expected 1-255
Config session committed successfully as d99c4cd and bgpd (restart) restarted.
$ fboss2-dev delete protocol bgp neighbor 2001:db8::99
Successfully deleted BGP neighbor 2001:db8::99
```

Idempotent delete, recorded on a lab device:

```
$ fboss2-dev delete protocol bgp neighbor 192.0.2.99
Warning: BGP neighbor 192.0.2.99 does not exist; nothing to delete
```

## Test plan

**Unit (delete)** — `fboss2_cmd_config_test`:

```
$ fboss2_cmd_config_test --gtest_filter=*DeleteBgpNeighbor*
[       OK ] CmdDeleteBgpNeighborTestFixture.argValidation (52 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteExistingNeighbor (165 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteByAnySpelling (123 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteUnknownNeighborWarns (68 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteTwiceIsIdempotent (588 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteUnknownLeavesOthersIntact (119 ms)
[==========] 6 tests from 1 test suites ran.
[  PASSED  ] 6 tests.
```

**Unit** — `fboss2_cmd_config_test`:

```
$ fboss2_cmd_config_test --gtest_filter=*BgpNeighbor*
[       OK ] CmdConfigBgpNeighborTestFixture.argValidation (95 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.bareCreate (110 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.setRemoteAsn (116 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.timersAccumulate (151 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.addPathMerge (230 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.passive (185 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.boolAndStringAttributes (1007 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.parityAttributes (873 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.valueValidation (133 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.routeLimits (140 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.normalizedAddressesShareOnePeer (114 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.prefixSpellingsShareOnePeer (90 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.timersSeededFromPeerGroup (111 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.timersSeededFromGlobalWithoutGroup (75 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.timersNotReseededOncePresent (140 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.routeLimitSeededFromPeerGroup (88 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.routeLimitSeededUnlimitedWithoutGroup (127 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.routeLimitNotReseededOncePresent (126 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.ttlSecurityHopsRange (171 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.addPathMergesWithPeerGroupValue (149 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.addPathCannotDisableGroupDirection (116 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.addPathNarrowsGroupBothToOneDirection (119 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.timerRanges (164 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.rejectedValueLeavesPeerUntouched (162 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.distinctAddressesCreateDistinctPeers (109 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.attributeHelpListsEveryAttribute (50 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.cliHelpFooterCarriesAttributeList (53 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.invalidValuesRejected (66 ms)
[       OK ] CmdConfigBgpNeighborTestFixture.unsupportedAttributesRejected (70 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.argValidation (53 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteExistingNeighbor (106 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteByAnySpelling (160 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteByEquivalentPrefixSpelling (115 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteUnknownNeighborRejected (57 ms)
[       OK ] CmdDeleteBgpNeighborTestFixture.deleteUnknownLeavesOthersIntact (180 ms)
[==========] 35 tests from 2 test suites ran. (5828 ms total)
[  PASSED  ] 35 tests.
```

**Integration** — `fboss2_integration_test` on a lab device with `bgp_pp` active:

```
$ fboss2_integration_test --gtest_filter='ConfigBgpNeighborTest.*'
[       OK ] ConfigBgpNeighborTest.SetRemoteAsnAndCommit (13722 ms)
[       OK ] ConfigBgpNeighborTest.SetTimersHoldTimeAndCommit (12148 ms)
[       OK ] ConfigBgpNeighborTest.GracefulRestartKeepsInheritedHoldTime (11413 ms)
[       OK ] ConfigBgpNeighborTest.SessionKnobsAndCommit (11168 ms)
[       OK ] ConfigBgpNeighborTest.RouteLimitWarningKeepsUnlimitedCap (12787 ms)
[       OK ] ConfigBgpNeighborTest.DeleteNeighborAndCommit (20256 ms)
[  PASSED  ] 6 tests.
```

Based-on: facebook/fboss#1345


fboss/cli/fboss2/tools/bgp_json_to_cli.py
```diff
@@ -10,95 +10,98 @@
...
 
 def escape_shell_arg(arg: str) -> str:
-    """Escape argument for shell usage."""
-    return shlex.quote(str(arg))
+    """Escape argument for shell usage.
+
+    Backslash-escapes the characters that stay special inside double quotes
+    (backslash, double quote, dollar, backtick) so free-form fields like
+    descriptions cannot terminate the quoted string or trigger expansion in
+    the generated script.
```
induvsuresh, line 31: But this will allow characters such as `;`, `&`, and `|`.
For eg, a description of `x;id` generates:
fboss2 ... description x;id

and this will run id separately again, could we stick to the shlex.quote method? otherwise its going to be a long list of characters to check in the if condition below and is not maintainable

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
```diff
@@ -0,0 +1,583 @@
...
+void setIpv4OverIpv6Nh(BgpPeer& p, bool v) {
+  p.v4_over_v6_nexthop() = v;
+}
+
+void setBindAddrAddress(BgpPeer& p, const std::string& v) {
+  p.local_addr() = v;
+}
+
+void setGracefulRestartTime(BgpPeer& p, int32_t v) {
+  p.bgp_peer_timers().ensure().graceful_restart_seconds() = v;
```
induvsuresh, line 192: With ensure() to bgp_peer_timers() will all the timers be cleared when graceful restart is reset?
we have to preserve the inherited values before clearing this

fboss/cli/fboss2/commands/config/protocol/bgp/BgpPeerAddr.h
```diff
@@ -0,0 +1,61 @@
...
+namespace facebook::fboss::bgpcli {
+
+// Normalize an address (or prefix, for passive listening sessions) to its
+// canonical string form. nullopt if the input parses as neither.
+inline std::optional<std::string> normalizeBgpPeerAddr(const std::string& s) {
+  if (auto addr = folly::IPAddress::tryFromString(s)) {
+    return addr->str();
+  }
+  if (auto network = folly::IPAddress::tryCreateNetwork(
+          s, /* defaultCidr */ -1, /* applyMask */ false)) {
```
induvsuresh, line 38: This helper keeps the host part of a prefix. For example, it keeps `10.0.0.1/24` instead of converting it to `10.0.0.0/24`.
BGP++ masks the prefix when it loads the config. It therefore treats `10.0.0.1/24` and `10.0.0.2/24` as the same dynamic neighbor. The CLI still treats these spellings as different. An update can create a duplicate, and a delete can fail to find the existing neighbor.
Please use `applyMask=true` and test these equivalent prefix spellings.

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
```diff
@@ -0,0 +1,583 @@
...
+  }
+  auto enable = parseBool(values[0]);
+  if (!enable) {
+    return err(
+        fmt::format(
+            "Error: Invalid {} value '{}'; expected true or false",
+            name,
+            values[0]));
+  }
+  int bits = peer.add_path() ? static_cast<int>(*peer.add_path()) : 0;
```
induvsuresh, line 332: Assume the peer group enables add_path=SEND. The peer does not have its own add_path value.
Running add-path send false clears the peer value. An empty peer value tells BGP++ to keep using the peer-group value, so SEND remains enabled even though the CLI reports success.

Please add an explicit disabled value, or start from the inherited value before applying the change. Add a test with SEND or BOTH enabled on the peer group.

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
```diff
@@ -0,0 +1,583 @@
...
+    return err(
+        fmt::format(
+            "Error: {} requires <{}|{}>",
+            kConnectMode,
+            kConnectModePassive,
+            kConnectModeActive));
+  }
+  if (values[0] == kConnectModePassive) {
+    peer.is_passive() = true;
+  } else if (values[0] == kConnectModeActive) {
```
induvsuresh, line 291: The ACTIVE option stores is_passive=false. BGP++ reads false as PASSIVE_ACTIVE, It does not mean active-only.
The CLI also rejects BOTH, even though BOTH describes the stored behavior.

Please expose PASSIVE_ONLY and PASSIVE_ACTIVE, or use passive true|false. The CLI should not call this mode ACTIVE.

fboss/cli/fboss2/commands/config/protocol/bgp/BgpCliAttrHandlers.h
```diff
@@ -0,0 +1,392 @@
...
+AttrHandler<T> bitRateAttr(
+    std::string_view name,
+    std::function<void(T&, const std::string&)> set) {
+  return [name, set = std::move(set)](
+             T& target, const Tokens& values) -> Result {
+    if (values.size() != 1) {
+      return err(
+          fmt::format("Error: {} requires <bits-per-second>[K|M|G]", name));
+    }
+    const std::string& v = values[0];
```
induvsuresh, line 292: BGP++ accepts link-bandwidth values such as auto, 1.5G, and 150g. This CLI accepts only whole numbers with uppercase K, M, or G.
As a result, an existing valid BGP config cannot always be recreated through this CLI. bgp_json_to_cli.py copies these values into the generated commands, which then fail.

Please use the same parser as BGP++. Add tests for auto, decimal values, and lowercase suffixes.

fboss/cli/fboss2/commands/config/protocol/bgp/BgpCliAttrHandlers.h
```diff
@@ -0,0 +1,392 @@
...
+template <typename T>
+AttrHandler<T> secondsAttr(
+    std::string_view name,
+    std::function<void(T&, int32_t)> set) {
+  return
+      [name, set = std::move(set)](T& target, const Tokens& values) -> Result {
+        if (values.size() != 1) {
+          return err(fmt::format("Error: {} requires <seconds>", name));
+        }
+        auto seconds = parseNonNegInt32(values[0]);
```
induvsuresh, line 231: A hold time must be 0 or between 3 and 65535 seconds. Values 1 and 2 prevent the peer session from starting. Values above 65535 change when written to the BGP message; for example, 65536 becomes 0.
Graceful-restart time can only hold values from 0 through 4095. For example, 4096 becomes 0.

Please reject values outside these ranges before saving the config.

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
```diff
@@ -0,0 +1,583 @@
...
+constexpr std::string_view kAddPathSend = "add-path send";
+constexpr std::string_view kAddPathReceive = "add-path receive";
+constexpr std::string_view kAfiDisableIpv4Afi = "afi disable-ipv4-afi";
+constexpr std::string_view kAfiDisableIpv6Afi = "afi disable-ipv6-afi";
+constexpr std::string_view kAfiIpv4OverIpv6Nh = "afi ipv4-over-ipv6-nh";
+constexpr std::string_view kAfiIpv4LabeledUnicast = "afi ipv4-labeled-unicast";
+constexpr std::string_view kAfiIpv6LabeledUnicast = "afi ipv6-labeled-unicast";
+constexpr std::string_view kBindAddrAddress = "bind-addr address";
+constexpr std::string_view kBindAddrPort = "bind-addr port";
+constexpr std::string_view kGracefulRestartTime =
```
induvsuresh, line 68: All neighbor options are parsed from one free-form args list. CLI11 does not know timers, afi, and add-path are actual commands.

Will this show up in help or completion checks ? 
The BGP Show CLIs register each command in `CmdList.cpp`. Please follow the same structure here and register these options as real subcommands.

induvsuresh:
Please update the BUCK dependencies. The build cannot find `bgp_config_types_fwd.h`, `bgp_config_types.h`, or `ConfigBgpTestBase.h` so CI internally will fail when I export

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
hillol-nexthop: A new wrapper is written which uses seedTimers to seed the values from peer groups or global

fboss/cli/fboss2/commands/config/protocol/bgp/BgpPeerAddr.h
hillol-nexthop: Used the `applyMask=true`

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
hillol-nexthop: @induvsuresh  With no bit set, it will inherit from peer-group by default. So, how do the operator use CLI to disable both send and receive even though the group has then enabled?
For now I am explicitly refusing the CLI when the group has some bit set and the CLI tries to clear all the bits.

fboss/cli/fboss2/commands/config/protocol/bgp/BgpCliAttrHandlers.h
hillol-nexthop: done

fboss/cli/fboss2/commands/config/protocol/bgp/BgpCliAttrHandlers.h
hillol-nexthop: done

hillol-nexthop:
> Please update the BUCK dependencies. The build cannot find `bgp_config_types_fwd.h`, `bgp_config_types.h`, or `ConfigBgpTestBase.h` so CI internally will fail when I export

done

fboss/cli/fboss2/commands/config/protocol/bgp/neighbor/CmdConfigProtocolBgpNeighbor.cpp
hillol-nexthop: Yes they will be shown in helptext. For now tab completion doesn't work for any subcommand as well. The reason we have moved to positional args based implementation rather than subcommands is because it reduces the code volume to a large extent. Every subcommand adds a large boilerplate code which adds to code volume.

fboss/cli/fboss2/tools/bgp_json_to_cli.py
```diff
@@ -23,11 +23,20 @@
...
-def escape_shell_arg(arg: str) -> str:
-    """Escape argument for shell usage."""
+def escape_shell_arg(arg: Any) -> str:
+    """Quote a JSON-sourced value for the generated shell script.
+
+    Applied to every value taken from the input JSON, numeric fields included:
+    the script does not validate types, and shlex.quote leaves plain digits
+    and identifiers unchanged, so valid input produces identical output.
+    """
     return shlex.quote(str(arg))
```
hjkim417: shell escaping misses characters such as ;, &, |, and newlines. A description like x;id injects a second command; restore shlex.quote() for every JSON-derived argument.

PR #1395, opened 2026-07-20
[Nexthop][fboss2-dev] fboss2 config/delete protocol bgp peer-group commands

hillol-nexthop:
> **Stacked on #1391** (`config/delete protocol bgp neighbor`). This PR's branch contains the whole stack below it, so the diff shows those commits too — please review only the `config/delete protocol bgp peer-group` commit(s) here. Merge the lower PRs first.
>
> **Stack (bottom → top):** #1345 → #1391 → **#1395 (this PR)** → #1401 → #1475 → #1477 → #1487 → #1488 → #1489 → #1490 → #1491

## Summary

Collapses the ~20 legacy per-attribute `config protocol bgp peer-group` command classes (built on the deprecated `folly::dynamic` `BgpConfigSession`) into a single typed dispatcher, mirroring `CmdConfigProtocolBgpNeighbor` (#1391). The group name is the first positional token; the next one or two tokens name the attribute (matched longest-prefix-first, so `timers hold-time` wins over any `timers` prefix); the rest are its value(s). Handlers mutate the typed `bgp::thrift::PeerGroup` directly via `ConfigSession::getBgpConfig()` / `saveBgpConfig()`, so adding a tunable is a one-entry change in the dispatch table rather than a new command class.

Also adds `delete protocol bgp peer-group <name>`. Deleting a group that a neighbor's `peer_group_name` still references is refused (the error names the referencing neighbors), so a commit can never carry a dangling group reference.

Covers the 36 dispatch keys that map to a `PeerGroup` field bgpd reads from a group: `remote-asn`/`local-asn` (4-byte-bounded), `description`, `peer-tag`, `ingress-policy`/`egress-policy`, `rr-client`, `confed-peer`, `redistribute-peer`, `enhanced-route-refresh`, `route-refresh`, `remove-private-as`, `enforce-first-as`, `passive`, `ttl-security-hops`, `add-path send|receive`, `afi disable-ipv4-afi|disable-ipv6-afi|ipv4-over-ipv6-nh`, `graceful-restart restart-time|stateful-ha`, `max-route pre-filter|post-filter` (+ the warning-threshold/warning-only knobs), `timers hold-time|keepalive|out-delay|withdraw-unprog-delay`, `next-hop-self`, `link-bandwidth`, `advertise-lbw`, `receive-lbw`. Spellings and validation are identical to the neighbor command, so the two share one attribute grammar.

Deliberately not exposed (documented in the dispatcher): `local_addr`, `next_hop4`, `next_hop6` (bgpd reads them per peer only, with no group fallback), `enabled`, `router_port_id` and `bgp_peer_timers.graceful_restart_end_of_rib_seconds` (read nowhere in bgpd).

### Behavior notes

- **`passive true|false`** exposes `is_passive` directly: bgpd reads true as PASSIVE_ONLY and false as PASSIVE_ACTIVE (listen and connect); the config model has no way to request active-only. `add-path send|receive` merges into the `AddPath` enum bitmask (RECEIVE/SEND/BOTH), and clearing the last direction unsets the field.
- **Inherited timers are preserved.** bgpd applies a group's `bgp_peer_timers` to its members as a whole struct, and hold/keepalive are non-optional i32 (default 0 = no keepalives). The first timer attribute set on a group seeds the struct from the global hold time (keepalive = hold/3, bgpd's own fallback) and then changes the one field.
- **Route limits are seeded unlimited, not 12000.** A group with no `pre_filter`/`post_filter` imposes no limit on its members (`AdjRib::capRoutesPerPeer` returns early on an absent struct), and bgpd reads `max_routes == 0` as unlimited. So the first `max-route ...-warning-threshold` / `...-warning-only` edit seeds `{max_routes 0, warning_only false, warning_limit 0}` and changes only the requested field, rather than materialising the thrift default `max_routes = 12000` — a hard cap with session teardown the operator never asked for.
- **Handlers edit a copy** of the group, so a rejected value leaves nothing seeded behind (and a rejected first attribute leaves no half-created group).
- **Validation matches bgpd**: hold time 0 or 3..65535 (bgpd refuses 1-2, the OPEN field is 16 bits), graceful-restart 0..4095 (RFC 4724), TTL security hops 1..255 (`kMin/kMaxTtlSecurityHops`; bgpd throws at load otherwise), `link-bandwidth` accepts what `parseLinkBandwidthBps` accepts (`auto`, decimals, case-insensitive K/M/G).
- **`bgp_json_to_cli.py`** generates the peer-group grammar above from the per-shape tables shared with the neighbor generator (only the scalar field list differs), so a fully populated `PeerGroup` converts with nothing dropped: `test_complete_peer_group_loses_no_field` asserts one command per leaf field (36). The deprecated i32 `remote_as`/`local_as` map to the 4-byte attribute. Fields bgpd never reads from a group are reported as `# WARNING: peer-group <name>: field <f> has no CLI equivalent, not converted` comment lines in the generated script (and on stderr) instead of being dropped silently.
- **Attributes with no per-peer-group thrift field** — `afi ipv4-labeled-unicast`, `afi ipv6-labeled-unicast`, `peer-port` — are not part of the dispatch table, so they are refused at parse time as unknown attributes; a rejected value never lands on disk.
- **CLI11 token stealing fixed** (same fix as the neighbor command): `positionals_at_end()` stops parent-chain subcommand fallthrough from reclassifying an attribute token that matches a sibling command name once the group name has been consumed.
- Integration tests verify the committed peer-group config through the daemon's `getRunningConfig` RPC — proving `bgpd` parsed and adopted the promoted `peer_groups` rather than only checking the file the CLI wrote.

**Delete is idempotent.** Deleting an absent target is a success with a `Warning:` (nothing is saved), so a replayed script never fails on something already gone; the reference guards still return `Error:`.

## Sample usage

```
$ fboss2-dev config protocol bgp peer-group SPINE remote-asn 65510
Successfully set remote-asn to: 65510 for peer-group SPINE
$ fboss2-dev config protocol bgp peer-group SPINE timers hold-time 90
Successfully set timers hold-time to: 90 seconds for peer-group SPINE
$ fboss2-dev config protocol bgp peer-group SPINE passive true
Successfully enabled passive for peer-group SPINE
$ fboss2-dev config protocol bgp peer-group SPINE max-route pre-warning-threshold 30000
Successfully set max-route pre-warning-threshold to: 30000 for peer-group SPINE
$ fboss2-dev config protocol bgp peer-group SPINE timers hold-time 2
Error: timers hold-time must be 0 or between 3 and 65535 seconds, got 2
$ fboss2-dev delete protocol bgp peer-group SPINE
Error: BGP peer-group SPINE is still referenced by neighbor(s): 192.0.2.1; delete or re-point those neighbors first
```

Idempotent delete, recorded on a lab device:

```
$ fboss2-dev delete protocol bgp peer-group NO-SUCH-GROUP
Warning: BGP peer-group NO-SUCH-GROUP does not exist; nothing to delete
```

## Test plan

**Unit (delete)** — `fboss2_cmd_config_test`:

```
$ fboss2_cmd_config_test --gtest_filter=*DeleteBgpPeerGroup*
[       OK ] CmdDeleteBgpPeerGroupTestFixture.argValidation (91 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteExistingGroup (195 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteUnknownGroupWarns (84 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteTwiceIsIdempotent (149 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteUnknownLeavesOthersIntact (102 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteReferencedGroupRejected (92 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteUnreferencedGroupSucceeds (124 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteThenRecreateStartsClean (174 ms)
[==========] 8 tests from 1 test suites ran.
[  PASSED  ] 8 tests.
```

**Unit** — `fboss2_cmd_config_test`:

```
$ fboss2_cmd_config_test --gtest_filter=*BgpPeerGroup*
[       OK ] CmdConfigBgpPeerGroupTestFixture.argValidation (94 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.bareCreate (106 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.setRemoteAsn (79 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.timersAccumulate (187 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.addPathMerge (152 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.passive (97 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.boolAndStringAttributes (765 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.routeLimits (369 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.linkBandwidthAttributes (211 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.ttlSecurityHopsRange (108 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.valueValidation (82 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.wrongValueCountRejected (76 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.overwriteSameAttributeLastWins (137 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.namedGroupsAreDistinct (128 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.timersSeededFromGlobal (102 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.timersNotReseededOncePresent (161 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.routeLimitSeededUnlimited (122 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.routeLimitNotReseededOncePresent (115 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.timerRanges (193 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.rejectedValueLeavesGroupUntouched (114 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.invalidValuesRejected (66 ms)
[       OK ] CmdConfigBgpPeerGroupTestFixture.unsupportedAttributesRejected (65 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.argValidation (69 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteExistingGroup (118 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteUnknownGroupRejected (59 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteUnknownLeavesOthersIntact (95 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteReferencedGroupRejected (103 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteUnreferencedGroupSucceeds (148 ms)
[       OK ] CmdDeleteBgpPeerGroupTestFixture.deleteThenRecreateStartsClean (128 ms)
[==========] 29 tests from 2 test suites ran. (4261 ms total)
[  PASSED  ] 29 tests.
```

**Integration** — `fboss2_integration_test` on a lab device with `bgp_pp` active:

```
$ fboss2_integration_test --gtest_filter='ConfigBgpPeerGroupTest.*'
[       OK ] ConfigBgpPeerGroupTest.SetRemoteAsnAndCommit (11580 ms)
[       OK ] ConfigBgpPeerGroupTest.SetTimersHoldTimeAndCommit (14385 ms)
[       OK ] ConfigBgpPeerGroupTest.SetPassiveAndCommit (11728 ms)
[       OK ] ConfigBgpPeerGroupTest.SessionKnobsAndCommit (12531 ms)
[       OK ] ConfigBgpPeerGroupTest.RouteLimitWarningKeepsUnlimitedCap (11697 ms)
[       OK ] ConfigBgpPeerGroupTest.DeleteGroupAndCommit (19856 ms)
[  PASSED  ] 6 tests.
```

Based-on: facebook/fboss#1391


fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
```diff
@@ -13,30 +13,480 @@
...
+void setDisableIpv6Afi(PeerGroup& g, bool v) {
+  g.disable_ipv6_afi() = v;
+}
+
+void setIpv4OverIpv6Nh(PeerGroup& g, bool v) {
+  g.v4_over_v6_nexthop() = v;
+}
+
+void setGracefulRestartTime(PeerGroup& g, int32_t v) {
+  g.bgp_peer_timers().ensure().graceful_restart_seconds() = v;
```
induvsuresh, line 161: Same comment as the previous stack, we'd have to preserve the inherited values for this and the route limits

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
```diff
@@ -13,30 +13,480 @@
...
+          {std::string(kConfedPeer),
+           boolAttr<PeerGroup>(kConfedPeer, setConfedPeer)},
+          {std::string(kRedistributePeer),
+           boolAttr<PeerGroup>(kRedistributePeer, setRedistributePeer)},
+          {std::string(kEnhancedRouteRefresh),
+           boolAttr<PeerGroup>(kEnhancedRouteRefresh, setEnhancedRouteRefresh)},
+          {std::string(kConnectMode), connectMode},
+          {std::string(kAddPathSend), addPathSend},
+          {std::string(kAddPathReceive), addPathReceive},
+          {std::string(kAfiDisableIpv4Afi),
```
induvsuresh, line 328: Please update bgp_json_to_cli.py to use the new peer-group commands.
For example, it generates disable-ipv4-afi, but this command now requires afi disable-ipv4-afi. It also generates max-routes, warning-limit, and warning-only, which this command no longer accepts.

 Please add a test that converts a complete peer group and verifies that no fields are lost.

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
```diff
@@ -13,30 +13,480 @@
...
+    return err(
+        fmt::format(
+            "Error: {} requires <{}|{}>",
+            kConnectMode,
+            kConnectModePassive,
+            kConnectModeActive));
+  }
+  if (values[0] == kConnectModePassive) {
+    group.is_passive() = true;
+  } else if (values[0] == kConnectModeActive) {
```
induvsuresh, line 233: Same comment as the previous stack, Please expose PASSIVE_ONLY and PASSIVE_ACTIVE, or use passive true|false.

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
hillol-nexthop: done

fboss/cli/fboss2/tools/bgp_json_to_cli.py
```diff
@@ -480,6 +365,52 @@ def generate_peer_commands(peer: dict[str, Any]) -> list[str]:
...
+        return []
+
+    prefix = f"config protocol bgp peer-group {escape_shell_arg(name)}"
+    commands = _unconverted_field_warnings(
+        peer_group, _PEER_GROUP_HANDLED_FIELDS, f"peer-group {_printable(name)}"
+    )
+    commands.extend(
+        _generate_scalar_commands(peer_group, prefix, _PEER_GROUP_SCALAR_FIELDS)
+    )
+    commands.extend(_generate_session_commands(peer_group, prefix))
```
hjkim417, line 410: A valid peer group containing only its required name produces no executable command here, so JSON-to-CLI replay silently drops the group. Any neighbor referring to it then references a group that was never created. Please emit the bare <prefix> command when no attribute command was generated, and add a regression test for {"name": "TEST"}.

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
```diff
@@ -13,30 +13,625 @@
...
+  auto field = pre ? group.pre_filter() : group.post_filter();
+  if (field) {
+    return;
+  }
+  bgp::thrift::RouteLimit unlimited;
+  unlimited.max_routes() = 0;
+  unlimited.warning_only() = false;
+  unlimited.warning_limit() = 0;
+  field = std::move(unlimited);
+}
```
hjkim417, line 314: seedRouteLimit writes the seeded RouteLimit to a local auto field variable, not to the group. group.pre_filter() / group.post_filter() return an optional_field_ref by value, so field = std::move(unlimited) mutates a detached copy and the group's field stays unset. Compare with seedTimers, which correctly writes group.bgp_peer_timers() = std::move(timers). Fix: replace the last line with if (pre) { group.pre_filter() = std::move(unlimited); } else { group.post_filter() = std::move(unlimited); }, or assign through the ref directly (e.g. (pre ? group.pre_filter() : group.post_filter()) = ... if the ref supports it).

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
```diff
@@ -13,30 +13,625 @@
...
-
-  return fmt::format(
-      "BGP peer-group {} created.\n"
-      "Use subcommands: remote-asn, description, hold-time, keepalive\n"
-      "Config saved to: {}",
-      groupName,
-      session.getSessionConfigPath());
+  Result result = args.attr().empty()
+      ? ok(fmt::format(
+            "Successfully created BGP peer-group {}", args.groupName()))
```
hjkim417, line 619: NIT: When args.attr() is empty and the group already exists, the message still says "Successfully created BGP peer-group …". Consider checking existing != groups.end() and emitting something like "BGP peer-group <name> already exists" to avoid misleading the operator into thinking a new group was inserted.

hjkim417 (commented):
@hillol-nexthop, left few comments. Please take a look.

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
hillol-nexthop: `apache::thrift::optional_field_ref<bgp::thrift::RouteLimit&>` is the return type of group.pre_filter() and group.post_filter() and this is a ref. In fact UT `CmdConfigBgpPeerGroupTest.cpp` covers this.

fboss/cli/fboss2/commands/config/protocol/bgp/peer-group/CmdConfigProtocolBgpPeerGroup.cpp
hillol-nexthop: Updated

fboss/cli/fboss2/tools/bgp_json_to_cli.py
hillol-nexthop: done

PR #1402, opened 2026-07-22
[Nexthop][distro] Raise root's nofile ulimit for hw tests

zackary-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

`qsfp_hw_test` opens one thread per transceiver slot. On m4062nhp (129 transceiver slots) this exceeds the default `ulimit -n` of 1024, so every test case fails in `SetUp()` with "Too many open files". Lower-port platforms never hit this ceiling.

Raises root's nofile default to 65536 via a limits.d entry in the image template. No new capability granted: the hard limit was already 524288 system-wide.

# Test Plan

Applied the same `limits.conf` change on a live m4062nhp DUT and confirmed `ulimit -n` reads 65536 and the previously-failing qsfp hw tests now pass.

### DIFF
--- a/fboss-image/image_builder/templates/centos-09.0/root_files/etc/security/limits.d/99-fboss-nofile.conf
+++ b/fboss-image/image_builder/templates/centos-09.0/root_files/etc/security/limits.d/99-fboss-nofile.conf
@@ -1,0 +1,4 @@
+# Raise root's default nofile cap for hw tests that open one fd set per
+# transceiver slot (e.g. qsfp_hw_test on high-port-count platforms).
+root soft nofile 65536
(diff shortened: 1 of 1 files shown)


srikrishnagopu:
Please remove the m4062 tag since this is related to Distor.

@kevin645 please review.

zackary-nexthop:
> @kevin645 has **imported** this pull request. If you are a Meta employee, you can view this in [D113475164](https://www.internalfb.com/diff/D113475164).

@kevin645 @srikrishnagopu I hadn't noticed that this had been imported already. Had to rebase so workflows pass, hope that's ok!

PR #1407, opened 2026-07-23
[Nexthop][fboss2-dev] Create VLANs on the fly in "config vlan" and "switchport access vlan"

benoit-nexthop:
# Summary

"config vlan 4005" with no subcommand used to fail with "Incomplete command, please use one of the subcommands", and "config interface <port> switchport access vlan <id>" refused to run when the VLAN didn't exist. Both now create the VLAN (and its barebone interface) via VlanManager::createVlan() in the config session:

- config vlan <id>: creates the VLAN, or reports "already exists". Implementation moved from the inline stub in CmdConfigVlan.h to CmdConfigVlan.cpp.
- config interface <port> switchport access vlan <id>: auto-creates a missing VLAN and notes "(VLAN <id> created)" in the output.

Add CmdConfigVlanTest.cpp covering creation (including persistence of the session config to disk) and idempotency, and rewrite the switchport access vlan "throws when VLAN missing" test to verify the new auto-create behavior.

# Test Plan

Add a new unit test / update existing unit test.

Tested on wedge800bnhp: "config vlan 4005" and "switchport access vlan 4021" both create the VLAN + fbossNNNN interface, visible in "config session diff"; re-running reports the VLAN already exists.

### DIFF
--- a/cmake/CliFboss2TestConfig.cmake
+++ b/cmake/CliFboss2TestConfig.cmake
@@ -33,6 +33,7 @@ add_executable(fboss2_cmd_config_test
   fboss/cli/fboss2/test/config/CmdConfigVlanManagerTest.cpp
   fboss/cli/fboss2/test/config/CmdConfigVlanPortTaggingModeTest.cpp
   fboss/cli/fboss2/test/config/CmdConfigVlanStaticMacTest.cpp
+  fboss/cli/fboss2/test/config/CmdConfigVlanTest.cpp
   fboss/cli/fboss2/test/config/CmdDeleteAclRuleTest.cpp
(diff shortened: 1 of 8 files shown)


joseph5wu:
@benoit-nexthop this PR failed the GHA build fboss2 cli job, can you fix it?

hillol-nexthop:
Rebased onto latest main. The three integration tests this PR adds to `ConfigVlanCreateTest` were adapted to 5e013971e6 ("Pick supported VLAN IDs in integration tests"), which landed since the last rebase and reworked the same suite: they now use the dynamically picked `s_newVlanId_` and `deleteVlanIfPresent()` instead of hardcoded VLAN IDs and `cleanupVlanFromConfig()` (which 5e013971e6 removed). The command-side changes are untouched.

PR #1422, opened 2026-07-27
[Nexthop][fboss2-dev] sort show interface status and transceiver by numeric interface ID

benoit-nexthop:
# Summary

`show interface status` and `show transceiver` ordered interfaces by a plain lexicographic comparison of the interface name, so ports came out as:

```
eth1/1/1
eth1/10/1
eth1/100/1
eth1/101/1
eth1/2/1
```

`show port` doesn't have this problem because it sorts with `utils::comparePortName`, which parses the `moduleNum/port/subport` components and compares them numerically. This PR points the other two commands at the same comparator so all three sort consistently:

```
eth1/1/1
eth1/2/1
eth1/10/1
eth1/100/1
eth1/101/1
```

- Adds `utils::parsePortName()` — a non-throwing parse returning `std::optional<PortNameParts>`.
- Adds `utils::comparePortNameOrId(nameA, idA, nameB, idB)`: orders by numeric module/port/subport when the name encodes a front panel position, falls back to the numeric ID when it doesn't, and keeps physical ports ahead of unparseable names. Never throws.
- Points `show port`, `show interface status` and `show transceiver` at it, plumbing the numeric IDs through their models (`InterfaceStatus.portId`; `TransceiverDetail.transceiverID` + optional `portID`, since bypass modules have no agent port).

# Test Plan

Added unit tests:

- `CmdShowInterfaceStatusTest.createModelNaturalSort` — feeds `eth1/2/1, eth1/10/1, eth1/1/1, eth1/100/1, eth2/1/1` and asserts the model comes out in numeric order.
- `CmdShowInterfaceStatusTest.createModelFreeFormNames` — `downlinks` and an unset name alongside a normal port; `ASSERT_NO_THROW`, and the unparseable names order by ID after the physical port.
- `CmdShowTransceiverTest.printOutputNaturalSort` — injects `eth1/10/1` (which sorts between `eth1/1/1` and `eth1/2/1` as a string) and asserts it is printed after `eth1/8/1`, so same coverage for the transceiver path, including a bypass module with no port ID.

Existing `CmdShowPortTest` / `CmdShowTransceiverTest` expectations are unchanged; the transceiver fixture now populates the new ID fields.

### DIFF
--- a/fboss/cli/fboss2/commands/show/interface/status/CmdShowInterfaceStatus.cpp
+++ b/fboss/cli/fboss2/commands/show/interface/status/CmdShowInterfaceStatus.cpp
@@ -77,6 +77,7 @@ CmdShowInterfaceStatus::RetType CmdShowInterfaceStatus::createModel(
       const auto operState = folly::copy(portInfo.operState().value());
 
       ifStatus.name() = portInfo.name().value();
+      ifStatus.portId() = folly::copy(portInfo.portId().value());
       ifStatus.description() = portInfo.description().value();
       ifStatus.status() =
           (operState == facebook::fboss::PortOperState::UP) ? "up" : "down";
@@ -111,8 +112,12 @@ CmdShowInterfaceStatus::RetType CmdShowInterfaceStatus::createModel(
   std::sort(
       model.interfaces()->begin(),
       model.interfaces()->end(),
-      [](cli::InterfaceStatus& a, cli::InterfaceStatus b) {
-        return a.name().value() < b.name().value();
+      [](const cli::InterfaceStatus& a, const cli::InterfaceStatus& b) {
+        return utils::comparePortNameOrId(
+            a.name().value(),
+            folly::copy(a.portId().value()),
+            b.name().value(),
+            folly::copy(b.portId().value()));
       });
   return model;
 }
--- a/fboss/cli/fboss2/commands/show/interface/status/model.thrift
+++ b/fboss/cli/fboss2/commands/show/interface/status/model.thrift
@@ -14,4 +14,5 @@ struct InterfaceStatus {
   5: string speed;
   6: string vendor;
   7: string mpn;
+  8: i32 portId;
 }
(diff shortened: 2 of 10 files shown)


joseph5wu:
@benoit-nexthop Now I'm looking into the following two commands implementation:
- https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/show/interface/CmdShowInterface.cpp#L108
- https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/show/transceiver/CmdShowTransceiver.cpp#L257

I think both show interface and show transceiver are actually using `portName` as the `Interface` name.

I am not a big fan of introducing portID in these transceiver/interface modes, so maybe we just need to have these two command to use the this function.
```
bool compareInterfaceName(
    const std::basic_string<char>& nameA,
    const std::basic_string<char>& nameB) {
  try {
    return comparePortName(nameA, nameB);
  } catch (const std::invalid_argument& ex) {
     // fall back to just compare the name 
     return nameA.value() < nameB.value();
  }
}
```

benoit-nexthop:
This was actually my initial implementation, however sorting on port ID generally leads to correct sort output when dealing with "non standard" names. What's the downside of introducing portID in these transceiver/interface models?

joseph5wu:
> This was actually my initial implementation, however sorting on port ID generally leads to correct sort output when dealing with "non standard" names. What's the downside of introducing portID in these transceiver/interface models?

The main downside is the misleading 1 interface/transceiver mapping to 1 port idea, which I am trying to avoid.
The portId is never assigned based on the name or front panel port order, as it's SDK logical ID and we never guarantee that id 1 is the first port. Hence, I will stick with using the string `name` to order the show command output table.

benoit-nexthop:
> > This was actually my initial implementation, however sorting on port ID generally leads to correct sort output when dealing with "non standard" names. What's the downside of introducing portID in these transceiver/interface models?
> 
> The main downside is the misleading 1 interface/transceiver mapping to 1 port idea, which I am trying to avoid. The portId is never assigned based on the name or front panel port order, as it's SDK logical ID and we never guarantee that id 1 is the first port. Hence, I will stick with using the string `name` to order the show command output table.

The goal isn't so much to guarantee that "ID 1 is the first port", but _in practice_ I've seen that sorting on IDs _tends_ to give us a natural sorting order. Sorting based on random strings give us an order that is entirely dependent on the string. The ID won't change.

PR #1432, opened 2026-07-28
[Nexthop] Enable "led_service" in FBOSS distro image

marif-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

`led_service` was missing from the FBOSS OSS distro image.

In this PR:
Build the `led_service` binary as part of the FBOSS platform stack and set up the runtime config/settings so it starts on boot:

- Add `led_service` to the `fboss_platform_services` CMake target so it ships in `platform-stack.tar`.
- Add a `systemd` unit (`led_service.service`) ordered after `platform_manager` and `fboss_init`.
- Enable the unit in the FBOSS distro image (`config.sh`).
- Add a minimal `led.conf` at `default_configs/` and symlink it from other platforms (`minipack3`, `wedge800bact`, `wedge800cact` and `wedge800bnhp` etc.).
- Have `fboss_init.sh` copy `/etc/coop/led.conf` from the platform's `default_configs` at boot.

- FSDB subscription paths - OSS stubs in `FsdbSwitchStateSubscriber.cpp` returned empty path vectors for both switch-state and transceiver-state, so the second `addStatePathSubscription({})` collided with the first on the root path `::1:/Path:/state:/` and threw `already exists` error. Filled the stubs in with the real paths, `agent().switchState()` and`qsfp_service().state().tcvrStates()` matching what each callback deserializes which finally fixed the error and `led_service` could run successfully.

# Test Plan

Verified `led_service` is healthy on a device after loading an image built from this branch.

```
root        1529  0.0  0.1 7082576 54460 ?       Ssl  18:16   0:01 /opt/fboss/bin/led_service
```

```
● led_service.service - FBOSS LED Service
     Loaded: loaded (/usr/lib/systemd/system/led_service.service; enabled; preset: disabled)
     Active: active (running) since Thu 2026-05-28 18:16:33 UTC; 1h 15min ago
```


srikrishnagopu:
@marif-nexthop I am not sure if we were able to get led service up and running with this config but internally we needed this for led service to work for montblanc.

"defaultCommandLineArgs": {
  "mode": "montblanc"
},
"ledControlledThroughService": true

Can you update your configs to reflect the platform name ?

marif-nexthop:
> @marif-nexthop I am not sure if we were able to get led service up and running with this config but internally we needed this for led service to work for montblanc.
> 
> "defaultCommandLineArgs": { "mode": "montblanc" }, "ledControlledThroughService": true
> 
> Can you update your configs to reflect the platform name ?

We had tested with and without the platform name and it worked the same. Not adding the name allowed to create only one file that can be symlinked into all the platforms and reused.
If a platform needs to add more info in this file, they'll have to delete the symlink and create their own configuration file but none of the existing platforms in the distro image needed it currently.

If you still want the platform name in this file we can make that change but AFAIK it is not required.

xanabani:
I dont know if there was a recent change that changed the led.conf, but this is what it looks like in a production device:
{
"defaultCommandLineArgs": {
  "mode": "montblanc"
},
"ledControlledThroughService": true
}

I belive the mode is required so that the correct led_manager for the device is created
[fbcode/fboss/led_service/LedManagerInit.cpp]

fbcode/fboss/led_service/LedManagerInit.cpp lines=40
  auto mode = productInfo->getType();

So the per-platform (e.g. wedge800cact/led.conf) default is wrong.

xanabani (approved):

marif-nexthop:
Thanks for the review @srikrishnagopu and @xanabani.
I have created per-platform `led.conf` files with `mode` field and deleted the common file that was symlinked earlier.

PR #1441, opened 2026-07-30
[Cisco][SRv6] Sync configured MySIDs from RIB to switch state

selekkala:
## Summary

Applying an unchanged configuration can produce no `ApplyThriftConfig` state
delta even when static MySIDs retained by the RIB need to be republished into
`SwitchState`. This change:

- adds a focused RIB API to republish the current MySID table;
- resynchronizes static MySIDs after config operations that add, update, or
  remove `mySidConfig`;
- restores MySIDs when the configuration is unchanged but `SwitchState` is
  stale;
- adds regression coverage for reapplying the same configuration.

## Test Plan

- `pre-commit` passed for all changed files.
- Installed the combined SRv6 image on an FBOSS switch and:
  - committed `mySidConfig` through the config-session workflow;
  - verified all configured entries were published through `show mysid`;
  - verified the generated agent config retained the expected MySID entries
    after config reload.

Validation artifacts:

- [SRv6 console log](https://gist.githubusercontent.com/selekkala/d104d72e5dc2e2969d894da12636199b/raw/76245e76e417e45b4de17dd9288386654fe4a4b7/srv6-console.log)
- [Generated agent.conf](https://gist.githubusercontent.com/selekkala/d104d72e5dc2e2969d894da12636199b/raw/666d04c69086cc7dd6a6037cb18684d53534d1a9/agent.conf)

### DIFF
--- a/fboss/agent/SwSwitch.cpp
+++ b/fboss/agent/SwSwitch.cpp
@@ -72,6 +72,7 @@
 #include "fboss/agent/RouteUpdateLogger.h"
 #include "fboss/agent/RxPacket.h"
 #include "fboss/agent/StaticL2ForNeighborObserver.h"
+#include "fboss/agent/SwSwitchMySidUpdater.h"
 #include "fboss/agent/SwSwitchRouteUpdateWrapper.h"
 #include "fboss/agent/SwSwitchWarmBootHelper.h"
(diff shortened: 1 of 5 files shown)


srikrishnagopu:
Closing PR based on discussion.

cc: @jasmeetbagga @vasantspatil

PR #1401, opened 2026-07-21
[Nexthop][fboss2-dev] fboss2 config/delete protocol bgp policy as-path-list commands

hillol-nexthop:
> **Stacked on #1395** (`config/delete protocol bgp peer-group`). This PR's branch contains the whole stack below it, so the diff shows those commits too — please review only the `config/delete protocol bgp policy as-path-list` commit(s) here. Merge the lower PRs first.
>
> **Stack (bottom → top):** #1345 → #1391 → #1395 → **#1401 (this PR)** → #1475 → #1477 → #1487 → #1488 → #1489 → #1490 → #1491

## Summary

Adds the first BGP **policy** object-type command family to `fboss2-dev`, under a new `policy` grouping node beneath `protocol bgp` (sibling to `global`/`neighbor`/`peer-group`):

- `config protocol bgp policy as-path-list <name> [description <string>]`
- `delete protocol bgp policy as-path-list <name>`

Patterns are flat `regex <regex>` attributes (one per invocation): bgpd matches on the flat `as_paths` string list and never reads the structured `as_path_list[]` entries, so the list has no nested entry level.

### Commands

| Command | Thrift target |
|---|---|
| `as-path-list <name>` | `aspath_lists[]` (create/select) |
| `as-path-list <name> description <string>` | `AsPathList.description` |
| `delete ... as-path-list <name>` | removes the `aspath_lists[]` entry |

### Design

- Follows the dispatcher shape the neighbor/peer-group families use: one factory per value shape from the shared `BgpCliAttrHandlers.h` (added in #1391), named setters that do nothing but assign the thrift field, and a registry that is one line per attribute. Writes `bgp_policy.BgpPolicies.aspath_lists[]` through the typed `ConfigSession` — no new session plumbing, `BgpConfig.policies` already exists.
- Lookup/create helpers live in `BgpAsPathListCliUtils.h`, shared with the delete command.
- A rejected value leaves nothing staged: a list implicitly created for the failed command is rolled back.

## Sample usage

Recorded on a lab device.

```
$ fboss2-dev config protocol bgp policy as-path-list AS65000
Successfully created BGP as-path-list AS65000
Config saved to: ~/.fboss2/bgp_config.json
$ fboss2-dev config protocol bgp policy as-path-list AS65000 regex ^65000_
Successfully added regex ^65000_ for as-path-list AS65000
Config saved to: ~/.fboss2/bgp_config.json
$ fboss2-dev config protocol bgp policy as-path-list AS65000 regex ^65000_
[...]

PR #1344, opened 2026-06-29
[Nexthop][fboss2-dev] BGP-aware fboss2 config session (ConfigSession infra)

hillol-nexthop:
## Summary

Make the fboss2 `ConfigSession` and `config session` lifecycle BGP-aware. This
is the base for the `config protocol bgp` commands, which build on top (see the
stacked PR).

- `ConfigSession` owns the BGP config as a typed `bgp::thrift::BgpConfig`,
  exactly the way it owns `cfg::AgentConfig` for the agent.
  `getBgpConfig()`/`saveBgpConfig()` expose the whole typed config and stage it
  to `~/.fboss2/bgp_config.json`. `ConfigSession` is BGP-aware but agnostic
  about the config's internal structure: a command mutates whichever typed
  fields it needs, mirroring how interface commands mutate
  `getAgentConfig().sw()->ports()`. The whole config round-trips through the
  typed struct, so no surgical merge / preserve-list is needed.
- `commit()` promotes `bgp_config.json` to `/etc/coop/bgpcpp/bgpcpp.conf`,
  restarts `bgp_pp`, and versions it in the same `/etc/coop` git repo as
  `agent.conf`.
- `config session clear/diff/rebase/rollback` are BGP-aware: clear removes a
  staged BGP session, diff shows staged BGP changes, rebase 3-way-merges
  `bgpcpp.conf`, rollback restores `bgpcpp.conf` + restarts `bgp_pp` (using
  metadata history so BGP-only commits are reachable).
- `FbossServiceUtil` + `cli_metadata` gain the `bgp_pp` (`BGP`)
  service/action; the config lib gains the `bgp_config` cpp2 dependency.

## Test Plan

- `fboss/cli/fboss2/test/config:cmd_config_test` — 455 tests pass
  (the BGP-session unit tests for the fixes below land in this PR).
- `//fboss/cli/...` builds on top of `upstream/main`.
- End-to-end coverage of the BGP session lifecycle on a live `bgp_pp` runs as
  `ConfigBgpSessionTest`, which is added in the stacked PR (the integration
  harness ships with the `config protocol bgp` commands) — 6/6 pass there.

PR #1446, opened 2026-07-31
[Nexthop][fboss2-dev] Unify agent/BGP config session + decouple ConfigSession header

hillol-nexthop:
> **Base of the `fboss2-dev config protocol bgp` stack** (follow-up to the merged #1344). Everything above it in the list below builds on this PR, so merge this one first.
>
> **Stack (bottom → top):** **#1446 (this PR)** → #1345 → #1391 → #1395 → #1401 → #1485 → #1475 → #1486 → #1477 → #1487 → #1488 → #1489 → #1490 → #1491

## Summary

Follow-up to #1344 (BGP-aware fboss2 config session): unify the agent and BGP
domains inside `ConfigSession` and decouple the `ConfigSession.h` header from
the heavy generated thrift types.

- Make `ConfigActionLevel` generic across services (review feedback below):
  `HITLESS` / `SERVICE_RESTART` / `DISRUPTIVE_SERVICE_RESTART` replace
  `HITLESS` / `AGENT_WARMBOOT` / `AGENT_COLDBOOT` (+ the bgpd-only `BGP_RESTART`).
  What a level means is decided per service from the (service, level) pair in
  `FbossServiceUtil::restartTypeName()`: agent → warmboot / coldboot, bgpd (no
  warmboot or hitless reload) → plain restart. `config session commit` labels
  the restarted services accordingly (`wedge_agent (warmboot)`, `bgpd (restart)`).
  A session staged before this change (old level names in
  `~/.fboss2/cli_metadata.json`) no longer parses and loses its recorded
  restart level (WARNING logged) -- `config session clear` it first.
- Introduce a single `ConfigDomain` descriptor + `configDomains()` and shared
  per-domain helpers so `commit()`, `rollback()` and `config session diff`
  handle the agent and BGP domains uniformly (private `DiffDomain` removed).
- Make the agent skip-when-unchanged like BGP: a commit whose staged config
  equals what is already promoted is a true no-op (no git revision, no symlink
  churn, no `reloadConfig()`/bgpd restart). Change detection is semantic
  (compare the deserialized thrift structs), so formatting-only diffs don't
  count.
- Consolidate `config session clear` onto a static `stagedSessionFilePaths()`
  and reuse `ConfigSession::readStagedContent()` in diff.
- Make `ConfigSession::saveConfig(service, level)` generic over the service and
  reduce `saveBgpConfig()` to a thin wrapper.
- Keep the heavy generated thrift headers out of `ConfigSession.h`: use the
  `*_types_fwd.h` forward-declaration headers, hold `agentConfig_`/`bgpConfig_`
  by `std::unique_ptr`, and drop the `configLoaded_`/`bgpConfigLoaded_` bools
  (null == not loaded).

## Test Plan

- `fboss/cli/fboss2/test/config:cmd_config_test` passes (the config-session /
  commit / diff / BGP / clear unit tests cover the refactored paths); re-run in
  full after the action-level rename (871 tests).
- `//fboss/cli/...` builds and tests pass on top of `upstream/main`.
- Verified the agent no-op commit behaviour live on test switches.

### DIFF
--- a/fboss/cli/fboss2/cli_metadata.thrift
+++ b/fboss/cli/fboss2/cli_metadata.thrift
@@ -14,18 +14,17 @@ namespace cpp2 facebook.fboss.cli
 
 // Action level required for config changes to take effect.
 // Used to track the highest impact action needed when committing config
-// changes.
+// changes. The levels are generic across services; how a level is carried out
+// for a given service (agent warmboot vs plain bgpd restart, ...) is decided by
+// FbossServiceUtil::restartService() from the (service, level) pair.
 enum ConfigActionLevel {
-  HITLESS = 0, // Can be applied with reloadConfig() - default
-  // Requires an agent warmboot restart (forwarding state retained).
-  AGENT_WARMBOOT = 1,
-  AGENT_COLDBOOT = 2, // Requires agent coldboot restart (clears ASIC state)
-  // Plain service restart, for daemons with no warmboot or hitless-reload
-  // support (bgpd). Mechanically identical to a warmboot restart (plain
-  // systemctl restart-and-wait, no coldboot marker). Levels are only ever
-  // compared within one service, so 3 > 2 does NOT mean "more impactful
-  // than a coldboot".
-  SERVICE_RESTART = 3,
+  // Can be applied with reloadConfig() if the service supports it - default
+  HITLESS = 0,
+  // Restart the service gracefully, retaining state where the service can:
+  // an agent warmboot, or a plain restart for daemons with no warmboot (bgpd).
+  SERVICE_RESTART = 1,
+  // Restart the service disruptively: an agent coldboot (clears ASIC state).
+  DISRUPTIVE_SERVICE_RESTART = 2,
 }
 
 // Identifier for different services that can be configured
--- a/fboss/cli/fboss2/cli_metadata.thrift
+++ b/fboss/cli/fboss2/cli_metadata.thrift
@@ -17,11 +17,15 @@ namespace cpp2 facebook.fboss.cli
 // changes.
 enum ConfigActionLevel {
   HITLESS = 0, // Can be applied with reloadConfig() - default
-  // Requires a service restart that preserves state where possible. For the
-  // agent this is a warmboot (forwarding state retained); for bgpd (BGP++),
-  // which has no hitless reload, it is a plain service restart.
+  // Requires an agent warmboot restart (forwarding state retained).
   AGENT_WARMBOOT = 1,
   AGENT_COLDBOOT = 2, // Requires agent coldboot restart (clears ASIC state)
+  // Plain service restart, for daemons with no warmboot or hitless-reload
+  // support (bgpd). Mechanically identical to a warmboot restart (plain
+  // systemctl restart-and-wait, no coldboot marker). Levels are only ever
+  // compared within one service, so 3 > 2 does NOT mean "more impactful
+  // than a coldboot".
+  SERVICE_RESTART = 3,
 }
 
 // Identifier for different services that can be configured
--- a/fboss/cli/fboss2/cli_metadata.thrift
+++ b/fboss/cli/fboss2/cli_metadata.thrift
@@ -17,9 +17,11 @@ namespace cpp2 facebook.fboss.cli
 // changes.
 enum ConfigActionLevel {
   HITLESS = 0, // Can be applied with reloadConfig() - default
-  AGENT_WARMBOOT = 1, // Requires agent warmboot restart
+  // Requires a service restart that preserves state where possible. For the
+  // agent this is a warmboot (forwarding state retained); for bgpd (BGP++),
+  // which has no hitless reload, it is a plain service restart.
+  AGENT_WARMBOOT = 1,
   AGENT_COLDBOOT = 2, // Requires agent coldboot restart (clears ASIC state)
-  BGP_RESTART = 3, // Requires a restart of the bgpd (BGP++) service
 }
 
 // Identifier for different services that can be configured
--- a/fboss/cli/fboss2/commands/config/session/CmdConfigSessionCommit.cpp
+++ b/fboss/cli/fboss2/commands/config/session/CmdConfigSessionCommit.cpp
@@ -59,11 +59,6 @@ CmdConfigSessionCommitTraits::RetType CmdConfigSessionCommit::queryClient(
               fmt::format("{} (warmboot)", serviceName));
         }
         break;
-      case cli::ConfigActionLevel::BGP_RESTART:
-        for (const auto& serviceName : serviceNamesList) {
-          restartedServices.push_back(fmt::format("{} (restart)", serviceName));
-        }
-        break;
       case cli::ConfigActionLevel::HITLESS:
         for (const auto& serviceName : serviceNamesList) {
           reloadedServices.push_back(serviceName);
--- a/fboss/cli/fboss2/commands/config/session/CmdConfigSessionCommit.cpp
+++ b/fboss/cli/fboss2/commands/config/session/CmdConfigSessionCommit.cpp
@@ -16,6 +16,7 @@
 #include <folly/String.h>
 #include <vector>
 #include "fboss/cli/fboss2/session/ConfigSession.h"
+#include "fboss/cli/fboss2/session/FbossServiceUtil.h"
 
 namespace facebook::fboss {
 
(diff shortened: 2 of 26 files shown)


fboss/cli/fboss2/session/ConfigSession.cpp
```diff
@@ -951,205 +1123,106 @@ ConfigSession::CommitResult ConfigSession::commit(const HostInfo& hostInfo) {
...
+    ConfigDomain domain;
+    std::string staged;
+    std::string oldPromoted;
+    bool promotedExisted;
+  };
+  std::vector<ConfigDomain> stagedDomains;
+  std::vector<Pending> pending;
+  // actions ends up holding exactly the pending domains' required action levels
+  // (returned in CommitResult and passed to applyServiceActions).
+  auto actions = requiredActions_;
```
induvsuresh, line 1158: Here, is BGP restarted? 
IIUC A domain in the `pending` list with no entry in `requiredActions_` sees a promote and no restart.

fboss/cli/fboss2/cli_metadata.thrift
```diff
@@ -17,9 +17,11 @@ namespace cpp2 facebook.fboss.cli
 // changes.
 enum ConfigActionLevel {
   HITLESS = 0, // Can be applied with reloadConfig() - default
-  AGENT_WARMBOOT = 1, // Requires agent warmboot restart
+  // Requires a service restart that preserves state where possible. For the
+  // agent this is a warmboot (forwarding state retained); for bgpd (BGP++),
+  // which has no hitless reload, it is a plain service restart.
+  AGENT_WARMBOOT = 1,
   AGENT_COLDBOOT = 2, // Requires agent coldboot restart (clears ASIC state)
-  BGP_RESTART = 3, // Requires a restart of the bgpd (BGP++) service
```
induvsuresh, line 22: Will this have backward compatibility issues? Could we mark it deprecated stead of delete

fboss/cli/fboss2/commands/config/session/CmdConfigSessionCommit.cpp
```diff
@@ -59,11 +59,6 @@ CmdConfigSessionCommitTraits::RetType CmdConfigSessionCommit::queryClient(
               fmt::format("{} (warmboot)", serviceName));
         }
         break;
-      case cli::ConfigActionLevel::BGP_RESTART:
-        for (const auto& serviceName : serviceNamesList) {
-          restartedServices.push_back(fmt::format("{} (restart)", serviceName));
```
induvsuresh, line 64: This deletion removes the "(restart)" text for BGP. BGP now uses the AGENT_WARMBOOT case above, which prints "(warmboot)". but BGP has no warmboot mechanism. Could we rename the enum to SERVICE_RESTART

fboss/cli/fboss2/session/ConfigSession.h
```diff
@@ -9,12 +9,16 @@
 
 #pragma once
 
+// Forward-declaration-only headers for the typed configs; the full generated
+// types are heavy and are only needed in ConfigSession.cpp (agentConfig_ and
+// bgpConfig_ are held by unique_ptr, so an incomplete type suffices here).
+#include <neteng/fboss/bgp/public_tld/configerator/structs/neteng/fboss/bgp/gen-cpp2/bgp_config_types_fwd.h>
```
induvsuresh, line 15: Does cmake generate this header?

fboss/cli/fboss2/test/config/CmdConfigSessionTest.cpp
```diff
@@ -1087,8 +1143,12 @@ TEST_F(ConfigSessionTestFixture, threeWayMergeScenarios) {
...
 
   setupMockedAgentServer();
-  // 5 commits: 2 in scenario 1, 2 in scenario 2, 1 in scenario 3 (rebase fails)
-  EXPECT_CALL(getMockAgent(), reloadConfig()).Times(5);
+  // Reloads happen only when a commit actually changes the promoted config
+  // (agent skip-when-unchanged): scenario 1 = 2 commits (both change config);
+  // scenario 2's session2 rebases to the SAME value session1 committed, so its
+  // commit is a no-op and does NOT reload -> only 1 reload there; scenario 3 =
+  // 1 commit (session2's rebase throws before committing). Total = 2 + 1 + 1.
+  EXPECT_CALL(getMockAgent(), reloadConfig()).Times(4);
```
induvsuresh, line 1178: nit: could we also add EXPECT_TRUE(result.commitSha.empty())

induvsuresh (changes requested):
Overall lgtm but have some comments on whether BGP restart has been removed on config save

fboss/cli/fboss2/session/ConfigSession.cpp
hillol-nexthop: Yes. If a command issues no entry in requiredActions_ then no restart.

fboss/cli/fboss2/session/ConfigSession.h
hillol-nexthop: yes

fboss/cli/fboss2/cli_metadata.thrift
hillol-nexthop: The config CLI for agent would continue to use AGENT_WARMBOOT and for BGP, the entire stack starts from here, so no issue in backward compatibility.

fboss/cli/fboss2/test/config/CmdConfigSessionTest.cpp
hillol-nexthop: done

fboss/cli/fboss2/commands/config/session/CmdConfigSessionCommit.cpp
hillol-nexthop: added service_restart enum.

induvsuresh (approved):

fboss/cli/fboss2/cli_metadata.thrift
```diff
@@ -17,9 +17,15 @@ namespace cpp2 facebook.fboss.cli
...
+  // Requires an agent warmboot restart (forwarding state retained).
+  AGENT_WARMBOOT = 1,
   AGENT_COLDBOOT = 2, // Requires agent coldboot restart (clears ASIC state)
-  BGP_RESTART = 3, // Requires a restart of the bgpd (BGP++) service
+  // Plain service restart, for daemons with no warmboot or hitless-reload
+  // support (bgpd). Mechanically identical to a warmboot restart (plain
+  // systemctl restart-and-wait, no coldboot marker). Levels are only ever
+  // compared within one service, so 3 > 2 does NOT mean "more impactful
+  // than a coldboot".
+  SERVICE_RESTART = 3,
```
joseph5wu: @hillol-nexthop This naming is a bit confusing and not very explicit especially `AGENT_WARMBOOT` and `SERVICE_RESTART` are basically the same thing?
Now with BGP config support, or potential future Qsfp config support, I think we need to revisit the meaning of these enums.
Correct me if I'm wrong, right now, you need `ServiceType` and `ConfigActionLevel` to decide what's the highest priority you need when handing session commit. 
Can we consider to rename `ConfigActionLevel` to be:
```
enum ConfigActionLevel {
HITLESS = 0,  // Can be applied with reloadConfig() if the service supports
GRACEFUL_SERVICE_RESTART = 1, // Restart service gracefully (i.e agent warmboot or regular bgp restart)
DISRUPTIVE_SERVICE_RESTART = 2, // Restart service disruptively (i.e agent coldboot)
}
```
My main goal is to make this action level generic without worrying if someone do `session.commit(AGENT, SERVICE_RESTART);

hillol-nexthop: @joseph5wu My initial implementation didn't even consider SERVICE_RESTART and I took the AGENT_WARMBOOT path for BGP which technically works. But I agree with Indu's comment that BGP doesn't support warmboot. So the code gives a false indication. GRACEFUL_SERVICE_RESTART is nothing but warmboot only and I don't think BGP implementation qualifies for this. Coldboot or DISRUPTIVE_SERVICE_RESTART involves marker file which again BGP doesn't use. So, BGP doesn't truly fit into any of these buckets.
My mental construct is:
Agent- {Warmboot, Coldboot}
BGP - {Restart}
Also with the new implementation the warmboot would be doing systemcl stop and start while with BGP just systemctl restart is fine.
So one way is to keep ConfigSession as the base class and implement derived classes for agents, BGP, qsfp And these restart specifics can be kept in the derived class implementation.

joseph5wu: After discussed offline, we decided to go with these three enums:
- hitless
- service_restart
- disruptive_service_restart

The goal is to keep the action level enum to be generic without being too specific for certain service's behavior, while in the code, we can base on SERVICE + ACTION_LEVEL to decide what kinda service restart needed in certain config.

induvsuresh:
@hillol-nexthop Looks like number of builds are failing in the PR - Build & Test fboss2 CLI, QSFP etc
is this intended? when I import the PR also I have CI failures

fboss/cli/fboss2/cli_metadata.thrift
hillol-nexthop: Updated the code

hillol-nexthop:
> @hillol-nexthop Looks like number of builds are failing in the PR - Build & Test fboss2 CLI, QSFP etc is this intended? when I import the PR also I have CI failures

The failures are in getdeps stage where ftpmirror.gnu.org returned HTTP 502. I have rebased the PR and the CI is rerunning. Hopefully all turns green.

PR #1451, opened 2026-08-03
[Nexthop] [fboss2] Insert vlanPorts membership in switchport access vlan

hillol-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

`config interface <name> switchport access vlan <id>` set `sw.ports[].ingressVlan` and retargeted **existing** `sw.vlanPorts[]` entries, but never **inserted** one when the port had no membership — leaving a config where the port is not a member of its ingress VLAN (rejected by the agent config validator). A port with multiple entries also got *all* of them retargeted, producing duplicate `(vlan, port)` pairs.

The command now leaves each port with exactly one untagged `vlanPorts` membership in the target VLAN:
- stale untagged (access) memberships are removed,
- an existing entry in the target VLAN is deduplicated,
- the new untagged entry is inserted,
- tagged (trunk) memberships in other VLANs are preserved.

## Integration test redesign

`ConfigVlanSwitchportAccessTest` is redesigned as **NewVlanAcrossPorts**: attach a brand-new VLAN to one port (auto-create + insert path end to end), then the now-existing VLAN to a second port; after each warmboot commit verify `ingressVlan` **and** the untagged `vlanPorts` membership via the agent's running config.

## Teardown

No manual unwind: the test harness's per-suite config containment (#1557) records the `/etc/coop` baseline before the suite starts and rolls the DUT back to it after the suite ends, so the device is left as found even when the test fails mid-way.

## Port selection

`NewVlanAcrossPorts` needs two distinct ports; it calls the existing `getRandomInterfacePortName()` twice (retrying until the names differ) — no new fixture helper.

# Test Plan

- UT `//fboss/cli/fboss2/test/config:cmd_config_test` — 17/17 pass, including new UTs: insertion when membership missing, dedup of conflicting entries, idempotency, trunk-membership preservation, membership on VLAN auto-create.
- On a freshly prepared lab device (NH-4010-F), `ConfigVlanSwitchportAccessTest.NewVlanAcrossPorts` — **PASS** (binary built from our internal tree carrying the identical test body). Both VLAN commits were recorded as `AGENT_WARMBOOT` in `cli/cli_metadata.json`, agents active after each commit, and the config was rolled back to the pre-suite baseline afterwards with no test residue in `cli/agent.conf`.

### DIFF
--- a/fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
+++ b/fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
@@ -38,6 +38,7 @@
 
 #include <folly/IPAddressV6.h>
 #include <thrift/lib/cpp/util/EnumUtils.h>
+#include <thrift/lib/cpp2/protocol/Serializer.h>
 #include "fboss/agent/gen-cpp2/platform_config_types.h"
 #include "fboss/agent/gen-cpp2/switch_config_types.h"
 #include "fboss/agent/if/gen-cpp2/FbossCtrl.h"
@@ -100,6 +101,31 @@ void Fboss2IntegrationTest::discardSession() const {
   // Reset the in-memory singleton so the next CLI command starts a fresh
   // session from the current system config, not stale in-process state.
   ConfigSession::resetInstance();
+}
+
+std::string Fboss2IntegrationTest::snapshotConfig() const {
+  discardSession();
+  auto snapshot = apache::thrift::SimpleJSONSerializer::serialize<std::string>(
+      ConfigSession::getInstance().getAgentConfig());
+  discardSession();
+  return snapshot;
+}
+
+void Fboss2IntegrationTest::restoreConfig(const std::string& configJson) const {
+  if (configJson.empty()) {
+    return;
+  }
+  discardSession();
+
+  XLOG(INFO) << "Restoring config snapshot...";
+  auto& session = ConfigSession::getInstance();
+  auto& config = session.getAgentConfig();
+  apache::thrift::SimpleJSONSerializer::deserialize(configJson, config);
+  session.saveConfig(
+      cli::ServiceType::AGENT, cli::ConfigActionLevel::AGENT_WARMBOOT);
+  commitConfig();
+  waitForAgentReady();
+  XLOG(INFO) << "Config snapshot restored";
 }
 
 Fboss2IntegrationTest::Result Fboss2IntegrationTest::executeCliCommand(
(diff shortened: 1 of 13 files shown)


hillol-nexthop:
Note on `ConfigCoppCpuTrafficPolicyTest.cpp`: the file ships here (byte-identical to our internal tree) but is deliberately not added to the integration-test build targets — it drives `config copp cpu-traffic-policy match ... action ...`, which is pending in #1379. Registering it in `cmake/CliFboss2TestIntegrationTest.cmake` and `fboss/cli/fboss2/test/integration_test/BUCK` is a one-line follow-up once #1379 lands. Verified on a lab device: with the command absent the test fails at runtime (arg-parse error), and the other suites in this PR — `ConfigVlanCreateTest` (6/6, including the auto-create and switchport tests), `ConfigVlanSwitchportAccessTest.NewVlanAcrossPorts`, and `ConfigPfcTest` — all pass on this branch (8/8).

fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
```diff
@@ -102,6 +103,31 @@ void Fboss2IntegrationTest::discardSession() const {
...
+    return;
+  }
+  discardSession();
+
+  XLOG(INFO) << "Restoring config snapshot...";
+  auto& session = ConfigSession::getInstance();
+  auto& config = session.getAgentConfig();
+  apache::thrift::SimpleJSONSerializer::deserialize(configJson, config);
+  session.saveConfig(
+      cli::ServiceType::AGENT, cli::ConfigActionLevel::AGENT_WARMBOOT);
```
joseph5wu: Can we guarantee warmboot is enough for restoring config snapshot?
What if the current config has some fields that used coldboot to get them program?
Reverting them might also need a coldboot?

joseph5wu (changes requested):
I strongly recommend to move the new snapshop test config for integration test out of this PR and have a new one to take care of it.
Besides, I don't think we should make Fboss2IntegrationTest to introduce the snapshop/restore config feature. Instead, we should consider to use ConfigSession::rollback(). I would also like to see whether rollback can take care of this case as expected(whether it needs coldboot or warmboot)

cc: @benoit-nexthop

fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
hillol-nexthop: @joseph5wu now this PR is particular to VLANs. No snapshotting logic here.

PR #1458, opened 2026-08-03
[Nexthop][platform_manager] fail fast on missing BSP rpm, support bspKmodsRpmVersion wildcard

benoit-nexthop:
# Summary

During early stages of NPI when using a hand-maintained `platform_manager.json` for a new platform not yet officially supported by FBOSS,  a stale `bspKmodsRpmVersion` made `platform_manager` retry a non-existent rpm 3 times and then abort with an opaque error `Failed to install rpm ... exit code 1`.

- `installRpm` now captures `dnf`'s `stderr` and reports whether the package exists in no enabled repo. That case is permanent, so `platform_manager` no longer retries and the error names the likely cause plus the BSP versions the image actually ships.
- `bspKmodsRpmVersion` can now be `"*"`, resolved at startup to the newest BSP rpm in `/usr/local/share/local_rpm_repo` for the running kernel, which removes the need for constantly updating this hand-maintained `platform_manager.json` as the BSP version moves forward.

This small ergonomic improvement can help speed things up durnig NPI.

# Test Plan

New unit tests.

### DIFF
--- a/fboss/platform/platform_manager/ConfigValidator.cpp
+++ b/fboss/platform/platform_manager/ConfigValidator.cpp
@@ -1272,6 +1272,11 @@ bool ConfigValidator::isValidBspKmodsRpmVersion(
     XLOG(ERR) << "BspKmodsRpmVersion cannot be empty";
     return false;
   }
+  // Wildcard: PkgManager resolves it to the newest BSP rpm shipped in the
+  // image. See kBspKmodsRpmVersionWildcard in PkgManager.h.
+  if (bspKmodsRpmVersion == "*") {
+    return true;
+  }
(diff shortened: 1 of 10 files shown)


Scott8440:
I would prefer to not add extra logic into the package management system unless absolutely necessary. What if we instead just improve the error messages, and don't retry when an RPM truly doesn't exist - remove the wildcard matcher?

benoit-nexthop:
@Scott8440 that would be an improvement for sure but it would still require manual work from the diags team to periodically bump up the BSP version number from their `platform_manager.json` for new platforms. The goal was to allow them to use `*` and forget about this altogether.

benoit-nexthop:
Hi @Scott8440, any other thoughts or do you guys not want to take this change upstream because of the wildcard matcher? Just wanna know which direction to take this PR in. Thanks.

PR #1466, opened 2026-08-05
[Nexthop][fboss2-dev] Add fboss2-dev delete copp subcommands (cpu-queue, reason)

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

**What**: Delete-side counterparts for the `copp` config commands:

- `delete copp cpu-queue <id>` — removes the whole `sw.cpuQueues[]` entry. Refused while rxReason mappings (`rxReasonToQueueOrderedList`) or cpu-traffic-policy actions (send-to-queue, user-defined-trap) still reference the queue id; the error names each referencing entry.
- `delete copp reason <reason-name>` — removes the `rxReasonToQueueOrderedList` entry.

**Why**: The config-side copp commands landed without a way to unset what they set.

# Notes

### Whole-queue delete is not undoable by the CLI

`delete copp cpu-queue <id>` erases the entire `cpuQueues[]` entry, matching `delete qos default-queue-config <id>`, which erases a whole entry from `defaultPortQueues` — same `list<PortQueue>` shape.

The difference is on the config side. `config qos default-queue-config` can write `reserved-bytes`, `shared-bytes`, `weight`, `scaling-factor`, `scheduling`, `stream-type`, `buffer-pool-name` and AQM entries via `utils::applyPortQueueConfig`, so a deleted default queue can be rebuilt by hand. `config copp cpu-queue` writes only `name` and `portQueueRate`, so a deleted cpu-queue cannot.

`PortQueue` has 17 fields and the CLI can set two of them (`name`, and `portQueueRate` via `rate-limit`). `streamType` and `scheduling` are non-optional in thrift, so a recreated queue silently picks up their zero values. A platform agent.conf ships queue 9 as:

```json
{ "id": 9, "name": "cpuQueue-high", "scheduling": 1, "streamType": 1 }
```

`scheduling` 1 is `STRICT_PRIORITY`, `streamType` 1 is `MULTICAST`. Delete it and recreate it and you get `scheduling` 0 (`WEIGHTED_ROUND_ROBIN`) and `streamType` 0 (`UNICAST`), with no command to set either back. The high-priority CPU queue would start round-robining.

So: delete a platform-provisioned cpu-queue only when you want it gone for good. Queues you created yourself are safe to delete and recreate.

To *change* a name or rate limit, no delete is involved — config overwrites in place:

```
fboss2-dev config copp cpu-queue 9 name some-name
fboss2-dev config copp cpu-queue 9 rate-limit pps 300
```

### `delete copp reason` takes no queue id

Config is `config copp reason <reason> queue <id>`, delete is `delete copp reason <reason>`. `rxReasonToQueueOrderedList` holds at most one entry per reason — config overwrites an existing reason rather than appending a second entry — so the reason name alone identifies the entry, and delete looks it up that way:

```cpp
// CmdDeleteCoppReason.cpp
auto it = std::find_if(
    list.begin(), list.end(), [&args](const cfg::PacketRxReasonToQueue& e) {
      return *e.rxReason() == args.getReason();
    });
```

Passing a queue id is rejected rather than ignored (`->expected(1)`, plus the arg parser throws on a second token). The queue id shows up on the way out instead — `Deleted reason ARP -> queue 9 mapping` — read off the entry being removed.

Re-pointing a reason at a different queue is a config, not a delete: `config copp reason arp queue 5` prints `Updated reason ARP -> queue 5`. Delete is for taking the reason off the CPU path entirely.

### Clearing what blocks a queue delete

The refusal message lists each referrer: `reason <NAME>` for rxReason mappings, `matcher '<name>' send-to-queue` / `matcher '<name>' user-defined-trap` for cpu-traffic-policy actions (`findQueueReferences()`, `CmdDeleteCoppCpuQueue.cpp`). Only those two action types carry a queue id, so only they can block. 

Clear reasons with `delete copp reason <reason>`; clearing a blocking action needs the cpu-traffic-policy action delete from #1379:

```
fboss2-dev delete copp cpu-traffic-policy match <matcher> action send-to-queue
fboss2-dev delete copp cpu-queue 9
```

The deprecated `rxReasonToCPUQueue` map is ignored entirely (per review — the field is no longer used).

# Test Plan

Unit tests — 13 added by this PR (7 in `CmdDeleteCoppCpuQueueTestFixture`, 6 across the `CmdDeleteCoppReason*` fixtures), all passing:

```
$ bazel-bin/fboss/cli/fboss2/test/config/cmd_config_test --gtest_filter='CmdDeleteCoppCpuQueueTestFixture.*:CmdDeleteCoppReason*'
[==========] Running 13 tests from 4 test suites.
[----------] 7 tests from CmdDeleteCoppCpuQueueTestFixture (685 ms total)
[----------] 4 tests from CmdDeleteCoppReasonTestFixture (310 ms total)
[----------] 1 test from CmdDeleteCoppReasonNoPolicyFixture (99 ms total)
[----------] 1 test from CmdDeleteCoppReasonNoListFixture (89 ms total)
[==========] 13 tests from 4 test suites ran. (1185 ms total)
[  PASSED  ] 13 tests.
```

Integration tests on a freshly imaged T1 device — 2/2 pass)
 One end-to-end test per command, both unconditional .
 Each test creates what it deletes (`config copp reason` creates `cpuTrafficPolicy` and `rxReasonToQueueOrderedList` when absent) and removes it again, so nothing touches a live CoPP policy. 

```
$ ./fboss2_integration_test --gtest_filter='DeleteCoppTest.*'
[==========] Running 2 tests from 1 test suite.
[ RUN      ] DeleteCoppTest.DeleteWholeCpuQueue
DeleteCoppTest.cpp:119] Creating scratch cpu-queue 3
[       OK ] DeleteCoppTest.DeleteWholeCpuQueue (185 ms)
[ RUN      ] DeleteCoppTest.DeleteReasonMapping
DeleteCoppTest.cpp:150] Mapping reason eapol -> scratch cpu-queue 3
DeleteCoppTest.cpp:161] Deleting reason eapol
DeleteCoppTest.cpp:170] Removing scratch cpu-queue 3
[       OK ] DeleteCoppTest.DeleteReasonMapping (242 ms)
[==========] 2 tests from 1 test suite ran. (427 ms total)
[  PASSED  ] 2 tests.

$ fboss2-dev config session diff        # device left unchanged
No differences between current live config and session config.
```

Sample usage on the same device — note the refusal list shrinks as the referencing reason is deleted:

```
$ fboss2-dev delete copp cpu-queue 9
localhost: Thrift call failed: 'Cannot delete cpu-queue 9: still referenced by
reason NDP, reason ARP_RESPONSE, reason BGP, reason BGPV6, reason LACP, reason ARP.
Delete those references first.'

$ fboss2-dev delete copp reason arp
Deleted reason ARP -> queue 9 mapping

$ fboss2-dev config session diff
--- current live config
+++ session config
@@ -146,10 +146,6 @@
         {
           "queueId": 1,
           "rxReason": 0
-        },
-        {
-          "queueId": 9,
-          "rxReason": 1
-        }
       ]
     },

$ fboss2-dev delete copp cpu-queue 9
localhost: Thrift call failed: 'Cannot delete cpu-queue 9: still referenced by
reason NDP, reason ARP_RESPONSE, reason BGP, reason BGPV6, reason LACP.
Delete those references first.'

$ fboss2-dev config session clear
Config session cleared successfully.
```

## Review Findings

Pre-publication review (7 reviewers + verifier, confidence >= 0.7) surfaced these findings; all are addressed in this PR:

- shared parsing helpers moved to a dedicated `CoppUtils.{h,cpp}` instead of living in the config command header
- `user-defined-trap` refusal branch gained a unit test
- duplicated not-found throw in the reason delete collapsed into one helper
- integration-test reason candidates use `cfg::PacketRxReason` enum values instead of bare ints


fboss/cli/fboss2/commands/config/copp/CoppUtils.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <folly/Conv.h>
+#include <folly/String.h>
+#include <thrift/lib/cpp/util/EnumUtils.h>
+#include <algorithm>
+#include <cctype>
+#include <stdexcept>
+
+namespace facebook::fboss {
+
+namespace copp_cpu_queue {
```
joseph5wu, line 23: Can you create another PR to simplify the subcommand here to just be `copp_queue` or replace all the `cpu_queue` with just `queue`?
copp already means control plane so the queue used in here already means cpu queue, I think we can make the whole command less verbose.
I think these should work:
```
fboss2-dev config copp queue XXX
fboss2-dev delete copp queue XXX
```

fboss/cli/fboss2/commands/delete/copp/cpu_queue/CmdDeleteCoppCpuQueue.cpp
```diff
@@ -0,0 +1,135 @@
...
+  if (policy.rxReasonToCPUQueue().has_value()) {
+    for (const auto& [rxReason, mappedQueueId] : *policy.rxReasonToCPUQueue()) {
+      if (mappedQueueId == queueId) {
+        references.push_back(
+            fmt::format(
+                "reason {} (deprecated rxReasonToCPUQueue map)",
+                apache::thrift::util::enumNameSafe(rxReason)));
+      }
+    }
+  }
```
joseph5wu: You can ignore this field as we don't use it internally anymore.
Besides we don't support setting the reason to the deprecated field in 
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp#L253
I think we should just stop using `rxReasonToCPUQueue` at all

fboss/cli/fboss2/test/config/CmdDeleteCoppCpuQueueTest.cpp
```diff
@@ -0,0 +1,244 @@
...
+  HostInfo hostInfo("testhost");
+  CoppCpuQueueDeleteArgs args({"5"});
+
+  EXPECT_THROW(cmd.queryClient(hostInfo, args), std::runtime_error);
+}
+
+// The deprecated rxReasonToCPUQueue map form is still read by the agent as
+// a fallback when rxReasonToQueueOrderedList is unset, so a queue referenced
+// only there must also refuse deletion. The CLI never writes this map; it
+// only appears in configs authored outside the CLI.
```
joseph5wu: Let's stop supporting this deprecated field: rxReasonToCPUQueue

joseph5wu (commented):
The main comment for this PR to fix is to remove the logic of handling `rxReasonToCPUQueue` as we already deprecated this field. Let me know once you remove the relevant code for this field.

Besides, when reviewing this PR, it made me think we might need 3 more improvements related to copp and you can address them in the future PRs respectively:
1) Rename cpu_queue to just queue. It's already under copp, so queue implicitly mean cpu_queue, so we don't need to be verbose about that;
2) About your comment in the summary that `config copp cpu-queue` can only support changing name and rate_limit. FBOSS supports cpu queue just as port queue, can you make another PR to make sure config for cpu queue with the missing attributes?
3) Since `rxReasonToQueueOrderedList` is actually an ordered list, I think we need to adjust the current `config copp reason <reason-name> queue <id>` to allow specify order rather than always push_back to the list:
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp#L276

cc: @benoit-nexthop

fboss/cli/fboss2/test/config/CmdDeleteCoppCpuQueueTest.cpp
vybhav-nexthop: done in the latest commit

fboss/cli/fboss2/commands/delete/copp/cpu_queue/CmdDeleteCoppCpuQueue.cpp
vybhav-nexthop: done in the latest commit

vybhav-nexthop: Done — removed the rxReasonToCPUQueue handling; the delete guard now only considers rxReasonToQueueOrderedList and cpu-traffic-policy actions.

fboss/cli/fboss2/test/config/CmdDeleteCoppCpuQueueTest.cpp
vybhav-nexthop: Done — legacy-map fixture removed along with the handling.

fboss/cli/fboss2/commands/config/copp/CoppUtils.cpp
vybhav-nexthop: Done in #1548 (stacked on this PR): hard rename to `config copp queue` / `delete copp queue`, plus your other two suggestions — full PortQueue attribute support via the same utils::applyPortQueueConfig the qos queue-config commands use, and an optional `order <n>` for `config copp reason` instead of always push_back.

vybhav-nexthop:
Updated: removed the rxReasonToCPUQueue handling and its test per review. The three follow-ups (cpu-queue -> queue rename, full cpu-queue attribute support, ordered insert for reasons) are up as #1548, stacked on this PR.

PR #1471, opened 2026-08-05
[Nexthop][fboss2-dev] allow addressing interfaces by interface ID

benoit-nexthop:
# Summary

`InterfaceList` now accepts a purely-numeric argument and resolves it as an interface ID, so e.g. `fboss2-dev config interface 2001 mtu 9000` works on the interface with ID 2001.

Name lookups take precedence over ID lookups: port name -> interface name -> interface ID.

# Test Plan

Unit tests cover resolution by interface ID, by interface name, and that an unknown numeric ID still throws.

### DIFF
--- a/fboss/cli/fboss2/utils/InterfaceList.cpp
+++ b/fboss/cli/fboss2/utils/InterfaceList.cpp
@@ -9,6 +9,7 @@
  */
 
 #include "fboss/cli/fboss2/utils/InterfaceList.h"
+#include <folly/Conv.h>
 #include <folly/String.h>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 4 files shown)


fboss/cli/fboss2/utils/InterfaceList.cpp
```diff
@@ -31,21 +46,40 @@ InterfaceList::InterfaceList(std::vector<std::string> names, bool allowMissing)
   for (const auto& name : names_) {
     Intf intf(name);
 
-    // First try to look up as a port name
-    cfg::Port* port = portMap.getPort(name);
+    // First try to look up as a port name.  A purely-numeric name may also
```
joseph5wu: @benoit-nexthop : as we discussed offline, port name should be quite standard for FBOSS platform, 
so let's don't support both port id and interface id together(like port id 10 might not be interface 10).
I think this function should follow this order:
- try to look up as a port name
- try to look up as an interface name
- try to look up as an interface id

benoit-nexthop: Done.

PR #1473, opened 2026-08-06
[Cisco] Add explicit 800G ZR PID profile for Morgan

selekkala:
# Summary

Add an explicit `OPTIC_TYPE_800_ZR` PID profile to Morgan's `fan_service.json`. The new profile intentionally uses the same PID parameters as `OPTIC_TYPE_800_GENERIC`, so the current thermal behavior is unchanged while the ZR policy is made explicit in the platform configuration.

This also provides a clear configuration placeholder for ZR-specific tuning when sufficient optics samples become available. There are no fan-service C++ logic changes in this PR.

# Test Plan

- Validated the updated file with `jq empty`.
- Verified the ZR and 800G generic PID profile objects are identical.
- Passed `git diff --check`.
- Verified the net PR diff contains only `fboss/platform/configs/morgan800cc/fan_service.json`.


somasun (changes requested):
Why not just add OPTIC_TYPE_800_ZR PidProfile to fan_service.json.  fan_service.json already contains 4 identical PidProfiles for 4 optic types.   Having fallback mechanism for ZR, but rest using identical PidProfile from fan_service.json is inconsistent.  We can also avoid all the C++ logic changes in this diff.  The C++ changes will get enabled on multiple platforms, and the config change is much simpler.

Steps:
* Add json entry.
* Then delete kOpticTypeFallbacks, revert getConfigOpticData to a plain find, and don't add getEffectiveOpticPidSettings at all. All three defect sites become correct by construction:

mikechoifb:
@somasun actually, fall back mechanism is needed as a safety feature. We have seen issues on switches where we did not initially expect ZR optics, but then later we ended up using ZR optics, in which case we either left these optics unattended or cause fan_service to crash. 

Previous PR related to the fall back mechanism only covered some of the fan control types and not all, which was why Cisco saw fan_service crash issue. I believe this PR fixes the issue. 

Another use case is : while it is desirable for Cisco and other partners to always come up with ZR thermal profiles, it is often not possible during NPI phase as we do not have enough ZR samples. In those cases, we will have to rely on the fallback mechanism.

somasun:
@mikechoifb  Not arguing we dont need fallback mechanism.  Under fallback mechanism as implemented the OPTIC_TYPE_800_ZR optic uses the parameters of OPTIC_TYPE_800_GENERIC from the json config.
        "OPTIC_TYPE_800_GENERIC": {
          "kp": -1,
          "ki": 0,
          "kd": -17,
          "setPoint": 70,
          "posHysteresis": 0,
          "negHysteresis": 0.5
        },
What I am suggesting is to just add the following in the config.
        "OPTIC_TYPE_800_ZR": {
          "kp": -1,
          "ki": 0,
          "kd": -17,
          "setPoint": 70,
          "posHysteresis": 0,
          "negHysteresis": 0.5
        },
The net effect is the same.  But adding OPTIC_TYPE_800_ZR explicitly in the json is better because of the following:
* The source of truth for data affecting the fan tuning should be in configurations.  The fan_service.json already lists 4 optic types with byte-identical profiles on already.
* The NPI case is unaffected. During NPI we can't tune ZR — but the fallback doesn't tune it either, it applies the generic values. Copying those same values into the json gives the identical thermal result today, and leaves a placeholder that gets tuned when samples land, instead of an implicit dependency that has to be remembered.
* A global table can't express per-platform judgement. kOpticTypeFallbacks hardcodes ZR→800G_GENERIC for every platform. This could be fine for this Cisco platform, but a subsequent platform could end up using unintended parameters, as vendors pay attention to the configs and not the code.

selekkala:
@somasun Thanks for the detailed feedback. I added an explicit `OPTIC_TYPE_800_ZR` PID profile to Morgan's `fan_service.json` in a separate commit, [ef6579b81f](https://github.com/facebook/fboss/pull/1473/commits/ef6579b81f1037c40845a89c63a79eec3f86762b). It uses the exact same parameters as `OPTIC_TYPE_800_GENERIC`, so the current Morgan thermal behavior is unchanged while the ZR profile is now explicit in the platform configuration and can be tuned independently in the future.

I also converted the PR to the config-only approach. The earlier `ControlLogic.cpp` changes and fallback-specific tests were reverted in [faf5f0b8e2](https://github.com/facebook/fboss/pull/1473/commits/faf5f0b8e275af0d7bb3f653662feacdaf192acc). The net PR diff now contains only the Morgan `fan_service.json` update, with no fan-service C++ behavior changes.

Please take another look when you have a chance.

selekkala:
@somasun I converted the PR to the config-only approach. The `ControlLogic.cpp` changes and fallback-specific tests have been reverted in [faf5f0b8e2](https://github.com/facebook/fboss/pull/1473/commits/faf5f0b8e2), and the net PR diff now contains only the explicit `OPTIC_TYPE_800_ZR` profile in Morgan's `fan_service.json`. The PR title and description have also been updated to match the reduced scope. Please take another look when you have a chance.

srikrishnagopu:
I do not see any code changes. Please re-open as needed.

PR #1474, opened 2026-08-06
[Nexthop] [fboss2] don't stage a config session from read-only commands

hillol-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

[Slack thread](https://nexthopai.slack.com/archives/D0AMDB6A92M/p1785991859035259)

# Summary

`fboss2-dev config history` and `fboss2-dev config session diff` are purely informational, but both create a config session as a side effect.

Both call `ConfigSession::getInstance()`, which lazily default-constructs the singleton, whose constructor unconditionally runs `initializeSession()`. With nothing staged, that takes the "starting a new session" branch: `ensureDirectoryExists(~/.fboss2)`, `copySystemConfigToSession()` (seeding `~/.fboss2/agent.conf` from the live `/etc/coop/agent.conf`), `base_ = git HEAD`, and `saveMetadata()` (writing `~/.fboss2/cli_metadata.json`).

Consequences:

- Running `config history` on a clean box silently starts a config session. `config session diff`/`commit` afterwards behave as if the user has an edit in flight.
- `session diff`'s own `"No config session exists. Make a config change first."` message is checked *after* `getInstance()` has already created the session — the message is false by the time it is printed.

## Fix

Add a `ConfigSession::SessionInit { CreateIfAbsent, ReadOnly }` mode, threaded through the constructors into `initializeSession()`:

- `ReadOnly` returns before the "start a new session" branch, so **nothing under `~/.fboss2` is written**. The git handle, path getters, `configDomains()` and `readStagedContent()` all still work, which is everything these two commands actually use.
- `CreateIfAbsent` remains the default, so every editing command is byte-for-byte unchanged.
- `loadConfig()` — the entry point of every editing command — now calls `initializeSession(CreateIfAbsent)` explicitly, so even a `ReadOnly`-constructed singleton materializes a session the moment someone genuinely reads the agent config.
- Call sites: `CmdConfigHistory` and `CmdConfigSessionDiff` pass `SessionInit::ReadOnly`.

Two deliberate choices worth reviewing:

1. **`initializeGit()` still runs in `ReadOnly` mode.** It only bootstraps the *system* config repo under `/etc/coop` (never `~/.fboss2`), it is idempotent once the repo exists, and both read commands need a usable repo to report anything at all. Skipping it would make `history` fail with a raw git error on a fresh device instead of showing the baseline commit. Happy to tighten this further if reviewers prefer.
2. **`~/.fboss2` is now ensured in `saveConfig()` and `saveMetadata()`** rather than only at construction, since the directory is no longer guaranteed to have been created up front. This also removes a pre-existing latent failure in the `rollback()` path, which calls `saveMetadata()` unconditionally.

A "delete the files on exit" approach was considered and rejected: it races against a concurrent `config session start`/edit in another terminal, and does a pointless copy-then-delete on every invocation. Not creating the file at all is strictly better.

# Test Plan

New unit tests (all use a new `CmdConfigTestBase::setupReadOnlyTestableConfigSession()` helper that installs a `SessionInit::ReadOnly` session, mirroring how the commands construct theirs):

`CmdConfigHistoryTest.cpp`
- `historyDoesNotCreateSessionFiles` — asserts `~/.fboss2` does not exist beforehand, runs the command, asserts the output still lists `Initial commit`, and that `agent.conf`, `cli_metadata.json`, `bgp_config.json` and `hasActiveSession()` are all still absent/false.
- `historyLeavesExistingSessionUntouched` — with a session already staged, the command still works and the staged content is byte-identical afterwards.
- `defaultSessionStillCreatesSessionFile` — regression guard that the default `CreateIfAbsent` editing path still seeds `~/.fboss2/agent.conf`.

`CmdConfigSessionDiffTest.cpp`
- `diffNoSessionDoesNotCreateSessionFiles` — the "No config session exists" message is now true when printed; no session files created.
- `diffTwoRevisionsDoesNotCreateSessionFiles` — the revision-vs-revision mode never consults the session, and no longer stages one.
- `diffReadOnlyStillSeesStagedSession` — a real staged session is still diffed correctly and left untouched.

Existing `CmdConfigHistoryTest`/`CmdConfigSessionDiffTest`/`CmdConfigSessionTest`/`ConfigSessionSystemdTest` cases exercise the unchanged `CreateIfAbsent` path.

### DIFF
--- a/fboss/cli/fboss2/session/ConfigSession.cpp
+++ b/fboss/cli/fboss2/session/ConfigSession.cpp
@@ -281,7 +281,7 @@ std::string ConfigSession::readCommandLineFromProc() const {
   return folly::join(" ", args);
 }
 
-ConfigSession::ConfigSession() {
+ConfigSession::ConfigSession(SessionInit init) {
   username_ = getUsername();
   std::string homeDir = getHomeDirectory();
 
@@ -295,17 +295,18 @@ ConfigSession::ConfigSession() {
   sessionConfigDir_ = homeDir + "/.fboss2";
   systemConfigDir_ = coopDir;
   git_ = std::make_unique<Git>(coopDir);
-  initializeSession();
+  initializeSession(init);
 }
 
 ConfigSession::ConfigSession(
     std::string sessionConfigDir,
-    std::string systemConfigDir)
+    std::string systemConfigDir,
+    SessionInit init)
     : sessionConfigDir_(std::move(sessionConfigDir)),
       systemConfigDir_(std::move(systemConfigDir)),
       username_(getUsername()),
       git_(std::make_unique<Git>(systemConfigDir_)) {
-  initializeSession();
+  initializeSession(init);
 }
 
 ConfigSession::ConfigSession(
@@ -328,10 +329,10 @@ std::unique_ptr<ConfigSession>& getInstancePtr() {
 }
 } // namespace
 
-ConfigSession& ConfigSession::getInstance() {
+ConfigSession& ConfigSession::getInstance(SessionInit init) {
   auto& instance = getInstancePtr();
   if (!instance) {
-    instance = std::make_unique<ConfigSession>();
+    instance = std::make_unique<ConfigSession>(init);
   }
   return *instance;
 }
@@ -445,6 +446,7 @@ void ConfigSession::saveConfig(
     throw std::runtime_error("No config loaded to save");
   }
 
+<<<<<<< HEAD
   // We need to do a round-trip through serialize -> parse -> toPrettyJson
   // because SimpleJSONSerializer handles Thrift maps with integer keys
   // (like clientIdToAdminDistance) by converting them to strings.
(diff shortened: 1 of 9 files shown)


fboss/cli/fboss2/session/ConfigSession.cpp
```diff
@@ -445,6 +446,7 @@ void ConfigSession::saveConfig(
     throw std::runtime_error("No config loaded to save");
   }
 
+<<<<<<< HEAD
```
joseph5wu: Please address the rebase conflicts here

fboss/cli/fboss2/session/ConfigSession.h
```diff
@@ -85,12 +85,24 @@ namespace facebook::fboss {
  */
 class ConfigSession {
  public:
+<<<<<<< HEAD
```
joseph5wu: Please address the rebase conflicts here

joseph5wu (changes requested):
Please address the rebase conflicts here

PR #1482, opened 2026-08-07
[Nexthop][m4062nhp] Fix qsfp_service crash on graceful exit

zackary-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

qsfp_service aborted on SIGTERM/SIGINT: `exit()` was called from inside
`signalReceived()`, running global destructors that tore down EventBases
still in use.

The crash needs a thread to be inside an event loop when `exit()` runs.
A fully cabled m4062nhp has 128 transceivers, so the i2c refresh pass,
xphy stats, and thrift workers are almost always mid-flight when the
signal lands. Lightly populated boxes usually finish shutdown first.

Changes:
- `signalReceived()`: replace `exit()` with `terminateLoopSoon()`; `main()`
  unwinds naturally.
- `main()`: flush async log handlers, then `_exit(0)` to skip global
  destructors.
- Ignore a second signal while shutdown is in progress.
- `stopQsfpServerGracefully()`: use the ThriftServer `StopController`
  instead of the async `stop()`.
- Exit non-zero if the event loop ends without a completed signal-driven
  shutdown (can only happen if a new `terminateLoopSoon()` caller is added).

# Test Plan

qsfp_service used to crash consistently on a fully cabled m4062nhp when
running t0 link test. Tested on a live m4062nhp system, it now stays up.

### DIFF
--- a/fboss/qsfp_service/Main.cpp
+++ b/fboss/qsfp_service/Main.cpp
@@ -1,9 +1,12 @@
 // Copyright 2004-present Facebook. All Rights Reserved.
 
-#include <thrift/lib/cpp2/server/ThriftServer.h>
+#include <thrift/lib/cpp2/server/ThriftServer.h> // NOLINT(misc-include-cleaner)
+
+#include <unistd.h>
 
 #include <folly/executors/FunctionScheduler.h>
 #include <folly/logging/Init.h>
+#include <folly/logging/LoggerDB.h>
 
 #include "fboss/qsfp_service/PortManager.h"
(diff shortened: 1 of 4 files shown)


xanabani (changes requested):
@zackary-nexthop  can you pull to this change
https://github.com/facebook/fboss/blob/main/fboss/lib/oss/ThriftServiceUtils.cpp#L47

and see if it fixed the original issue?

PR #1496, opened 2026-08-10
[Nexthop][fboss2-dev] Delete L3/SVI interfaces and manage SVI IP addresses

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

**What**: two related gaps in how `fboss2-dev` addresses L3/SVI interfaces, fixed on one shared resolver:

1. `delete interface <name|id>` can now remove an L3/SVI interface, not just a physical port.
2. IP addresses can be added to and removed from a VLAN SVI, addressed as `vlan<id>`:

```
fboss2-dev config interface vlan2001 ip-address 10.0.0.1/24
fboss2-dev config interface vlan2001 ipv6-address 2001:db8:2001::1/64
fboss2-dev delete interface vlan2001 ip-address 10.0.0.1/24
fboss2-dev delete interface vlan2001 ipv6-address 2001:db8:2001::1/64
```

**Why**: both commands previously required a handle that SVIs rarely have. `delete interface` only collected ports, so naming an L3 interface returned "No port found for the specified interface(s)". Address configuration resolved interfaces by exact `Interface.name` only — but generated configs frequently leave the name unset, and CLI-created SVIs are auto-named `fboss<id>`, so there was no reliable way to reach an SVI at all.

## Background: ports, VLANs, and L3 interfaces

`SwitchConfig` keeps them as separate lists — `sw.ports`, `sw.vlans`, and `sw.interfaces` (interfaces are where IP addresses live) — tied together like this:

```
Port ──< sw.vlanPorts >── VLAN ──── Interface(type=VLAN)    many VLANs per port; ≤1 SVI per VLAN
Port ─────────────────────────────  Interface(type=PORT)    routed port; exactly one interface
                                    Interface(loopback/virtual) — attached to nothing
```

- A port can be a member of many VLANs (802.1Q trunking). Membership lives in the `sw.vlanPorts` join list — one `{vlanID, logicalPort, emitTags}` row per port-VLAN pair — and `Port.ingressVlan` classifies untagged frames.
- A VLAN may have one `VLAN`-type interface, its SVI: the shared routed gateway for all of the VLAN's member ports. The SVI belongs to the VLAN (`Interface.vlanID`), not to any port — the VLAN a frame classifies into decides which SVI routes it.
- A routed port skips the VLAN hop: exactly one `PORT`-type interface is bound to it directly via `Interface.portID`.
- Loopback/virtual interfaces attach to no port or VLAN at all.

## How: name resolution

Every argument goes through the shared `utils::InterfaceList` resolver (from #1471), which now tries, in order:

1. port name
2. interface name
3. interface ID (numeric arguments)
4. `vlan<digits>` → the SVI: the interface whose `vlanID` matches (new `PortMap::getInterfaceForVlan`)

Earlier levels always win, so existing resolution is unchanged. A number that is both a port logical ID and an interface ID resolves to the interface — bare numbers are never looked up as port logical IDs, matching the amendment made to #1471 during its review; `numericArgPrefersInterfaceOverPortLogicalId` pins that. An unresolvable `vlan<id>` fails with `VLAN N not found in configuration` or `VLAN N has no L3 interface (SVI) configured` — nothing is created as a side effect.

A VLAN carrying two interfaces makes `vlan<id>` ambiguous, so `PortMap` records the collision and resolution refuses rather than mutating whichever interface happened to come last in the config.

`vlan<id>` is a real handle into the shared resolver, so every `config interface` / `delete interface` form now reaches SVIs. Two consequences, both guarded:

- Port-only attributes (`description`, `shutdown`, `type`, `queue-config`, `lldp-expected-*`, flow-control, loopback-mode, lookup-class) and `switchport access vlan` have nothing to write on a port-less SVI. They fail with "resolve to an L3 interface with no port; no changes made" instead of reporting success for a write that went nowhere; `switchport access vlan` refuses before it would create the target VLAN.
- `delete interface vlan<id>` with no attribute deletes the SVI, and when the VLAN is not the default and keeps no other interface it is cascaded away with its `vlanPorts` rows and static MAC entries — the L3-interface delete behavior described below, now reachable by VLAN name. The output names the cascaded VLANs ("also removed VLAN(s) N left without an interface").

## How: deleting an L3/SVI interface

`CmdDeleteInterface` partitions the resolved arguments into ports and portless L3 interfaces, then folds into the interface set every interface that cannot outlive a port being deleted: the port's router interface (bound by `portID`) and the SVI of a VLAN the port was the last member of. Both spellings of the same delete therefore go through one checked path, `deleteInterfaces`, which validates the whole set before mutating anything, so a refusal leaves the session untouched. `removePortsFromConfig` still drops the ports afterwards, but by then it only sweeps L2-only VLANs — it has no referrer checks of its own, so nothing reaches it unchecked.

It refuses any delete that would leave a config the agent rejects — or crashes on — at apply time:

- a `PORT`-type interface named on its own: its port would be left with no router interface (delete the port instead, which now takes the interface with it);
- a `SYSTEM_PORT` interface: the agent derives one per port on VOQ/DSF switches, so erasing the config row dangles an interface ID;
- a tunnel's underlay interface: `Tunnel.underlayIntfID` is a required field, so there is nothing to clear (delete the tunnel first);
- the VLAN-type interface of a VLAN that still has an enabled member port, or whose VLAN is still some port's untagged ingress VLAN: the agent rejects such a config outright.

Interfaces are removed before ports so the refusal checks see the pre-delete state, and `portsToDelete` is passed into the member-port check so `delete interface <svi> <its-only-port>` is accepted rather than wrongly refused. `VlanManager::eraseVlan` is the single owner of what a VLAN cascades (its interfaces, `vlanPorts` rows, static MAC entries); `delete vlan` and `delete interface` both call it.

## How: SVI IP addresses

Once a `vlan<id>` name resolves, the SVI flows through the existing `config interface` / `delete interface` ip-address branches unchanged — the agent programs addresses exclusively from `cfg::Interface::ipAddresses` (`ThriftConfigApplier::getInterfaceAddresses`), which is where those branches already write. `cfg::Vlan::ipAddresses` is deliberately not written: the `Vlan` constructor never reads it, so it is dead config, and a test pins that VLAN objects stay unmutated.

Validation: `ip-address` accepts IPv4 only, `ipv6-address` IPv6 only (shared `validateInterfaceIpAttr`); duplicate add is an idempotent success; deleting an absent address reports `not configured` without touching state.

# Test Plan

Unit — full `cmd_config_test` suite passes. Interface-delete coverage: resolution precedence, every refusal reason, combined port+SVI delete, no-mutation-on-refusal. SVI IP coverage (15 tests): add/delete v4+v6, family mismatch, unknown VLAN, VLAN without SVI, duplicate add, absent delete, port path unchanged, no mutation of VLAN objects or unrelated interfaces.

Integration on a T1 DUT — full `fboss2_integration_test` suite: **185 tests, 175 passed, 10 skipped, 0 failed** (20m20s). The 10 skips are platform/config-gated (`ConfigIpRouteTest`, `ConfigDhcpSourceOverrideTest`, one profile-subsume case, one default-VLAN precondition) and are untouched by this PR. Both agents ended the run with `NRestarts=0`, so every commit applied without a crash.

Covering this change specifically:

```
[       OK ] ConfigInterfaceVlanIpTest.AddThenDeleteSviAddresses (326 ms)
             # config vlan N -> commit -> config interface vlanN ip-address + ipv6-address
             # -> commit -> running config shows both -> delete both -> gone
[       OK ] DeleteInterfaceTest.DeleteWholePortRemovesCreatedSubport (679 ms)
             # whole-port delete takes the port's router interface with it
```

The positive whole-interface delete and every refusal reason are covered by the unit tests rather than by integration tests: a PORT-type interface cannot be created through the CLI, and a session's port map is not rebuilt for interfaces staged later in the same session, so an integration test could only exercise those paths conditionally on the DUT's running config.


vybhav-nexthop:
Ported the internal review follow-ups from nexthop-ai/private-fboss#1615:

- Inlined `deleteInterfaces` into the delete-interface command (dropped the one-function `InterfaceManager` class).
- Cascading a VLAN that would otherwise be left without an interface (except the default VLAN, which keeps the VLAN and only clears `intfID`).
- Rebuilding the session `PortMap` after VLAN/interface staging mutations.
- `SCOPE_EXIT` VLAN cleanup is now noexcept so clang-tidy `bugprone-exception-escape` does not fire.

PR #1497, opened 2026-08-10
[Nexthop][fboss2-dev] Add delete qos policy subcommands

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

## Background: what a QoS policy is and how it is used

A QoS policy is a named bundle of mappings that decides how the switch treats each packet. Its `qosMap` translates packet markings into an internal traffic class (`dscpMaps` for IP DSCP, `expMaps` for MPLS EXP, `pcpMaps` for 802.1p), then translates that traffic class into hardware treatment: which egress queue (`trafficClassToQueueId`), which priority group buffer (`trafficClassToPgId`), and how PFC priorities map to queues and priority groups for lossless traffic (`pfcPriorityToQueueId`, `pfcPriorityToPgId`).

The policy does nothing until it is attached: `dataPlaneTrafficPolicy.defaultQosPolicy` applies it switch-wide, `portIdToQosPolicy` applies it per port, and `cpuTrafficPolicy` uses one for traffic headed to the CPU. The attachment is the policy's name as a plain string, which is why deleting a still-referenced policy has to be refused rather than allowed to leave a dangling name.

**What.** Adds the delete counterpart to the existing `config qos policy` family:

```
fboss2-dev delete qos policy <name>
fboss2-dev delete qos policy <name> map dscp <dscp-value>
fboss2-dev delete qos policy <name> map mpls-exp <exp>
fboss2-dev delete qos policy <name> map dot1p <pcp>
fboss2-dev delete qos policy <name> map tc-to-queue <tc>
fboss2-dev delete qos policy <name> map pfc-pri-to-queue <pri>
fboss2-dev delete qos policy <name> map tc-to-pg <tc>
fboss2-dev delete qos policy <name> map pfc-pri-to-pg <pri>
```

**Why.** The QoS policy object and its mappings could be added but never taken
away. The map delete covers every QosMap type `CmdConfigQosPolicyMap` writes.
Anything less leaves one-way doors: settable via CLI with no CLI way back.

**How.**

Whole-policy delete refuses when the policy is still named by
`dataPlaneTrafficPolicy` or `cpuTrafficPolicy.trafficPolicy`, through either
`defaultQosPolicy` or `portIdToQosPolicy`, and reports the exact referring
field. These are plain strings in thrift, so nothing else stops a delete from
leaving a dangling reference. Refusing rather than cascading is deliberate:
clearing the referrer would change forwarding on ports the user did not name.

The three list maps (`dscpMaps`, `expMaps`, `pcpMaps`) are structurally
identical (groups keyed by `internalTrafficClass`, not flat maps), so they
share one erase helper: the codepoint is removed from the group that lists it.
A group left with no ingress codepoints *and* no egress rewrite carries no
information and is dropped. A group that still has the rewrite is kept, since
the two halves are independent.

The four simple maps (`trafficClassToQueueId`, `pfcPriorityToQueueId`,
`trafficClassToPgId`, `pfcPriorityToPgId`) are plain key erases. The optional
ones are reset once their last key goes, returning the field to its unset
default rather than leaving an empty map behind.

The policy-by-name scan moves into a shared `utils::findQosPolicy`
(`commands/config/qos/QosPolicyUtils.h`, alongside the existing
`PortQueueConfigUtils`), replacing three separate copies: one in each new
delete command and the hand-rolled loop in `CmdConfigQosPolicyMap`. The name
argument reuses `QosPolicyName` from the config side for the same reason, the
way `delete qos queuing-policy` reuses `QueuingPolicyName`.

## Sample usage

```
$ fboss2-dev delete qos policy olympic map dscp 46
Successfully deleted QoS policy 'olympic' dscp mapping for value 46
$ fboss2-dev delete qos policy olympic map tc-to-queue 2
Successfully deleted QoS policy 'olympic' tc-to-queue mapping for traffic class 2
$ fboss2-dev delete qos policy olympic map pfc-pri-to-pg 3
Successfully deleted QoS policy 'olympic' pfc-pri-to-pg mapping for priority 3
$ fboss2-dev delete qos policy olympic
Cannot delete QoS policy 'olympic': still referenced by dataPlaneTrafficPolicy.defaultQosPolicy. Unset those fields first, then retry the delete.
$ fboss2-dev delete qos policy staging
Successfully deleted QoS policy 'staging'
```

# Test Plan

24 unit tests in `CmdDeleteQosPolicyTest.cpp`, seeded with one deletable
policy carrying every map type and one policy referenced by
`dataPlaneTrafficPolicy.defaultQosPolicy`. Per-kind coverage: delete an
existing entry, delete a missing entry errors, the emptied-optional reset,
and value-range validation (dscp 0-63, everything else 3-bit).

```
[==========] 24 tests from 1 test suite ran. (1378 ms total)
[  PASSED  ] 24 tests.
```

Full `cmd_config_test` suite, unfiltered, no regressions:

```
//fboss/cli/fboss2/test/config:cmd_config_test    PASSED in 46.6s
Executed 1 out of 1 test: 1 test passes.
```

Integration test: `test/integration_test/DeleteQosPolicyTest.cpp`, one test,
self-cleaning and positive-flow only (missing-entry/missing-policy errors are
unit-tested): create a scratch policy carrying a dscp mapping and a
tc-to-queue mapping, delete the dscp entry and assert `trafficClassToQueueId`
is untouched, delete the tc-to-queue entry, delete the policy.

# Review Findings

Pre-publication review (fboss-review) is clean after fixes: the three refusal
branches (portIdToQosPolicy, cpuTrafficPolicy, and the no-qosMap guard) gained
unit coverage, and the DSCP bounds and find-policy-or-throw were de-duplicated
into QosPolicyUtils.


vybhav-nexthop:
Extended the map delete to every QosMap type the config command writes (mpls-exp, dot1p, pfc-pri-to-queue, tc-to-pg, pfc-pri-to-pg) — the remaining types were settable via CLI with no CLI way back. Same tokens as the config command; unit tests cover each kind plus the emptied-optional reset behavior.

PR #1522, opened 2026-08-17
[Nexthop][fboss2-dev] Add config and delete interface sflow sample-dest

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

## Background: what sFlow sampling is and what these three attributes control

A switch forwards far more traffic than it could ever copy for analysis, so sFlow samples instead: the ASIC picks one in N packets on a port and exports just those. Statistically that is enough to see top talkers, flow mix, and anomalies without touching line-rate forwarding.

Three `Port` fields control this:

- `sFlowIngressRate` / `sFlowEgressRate`: the sampling rate in each direction. Every 1/N packets is sampled; 0 disables sampling in that direction.
- `sampleDest`: where each sampled copy goes once sampling is enabled, trading CPU cost against flexibility:
  - `cpu`: the ASIC punts the sample to the switch CPU, and the agent wraps it into sFlow datagrams for the configured collectors. Flexible, but each sample costs CPU and punt-path bandwidth.
  - `mirror`: the ASIC forwards the sample out the port's ingress mirror session straight to a remote collector box, encapsulated, with no CPU involvement. Line rate, but the collector does all the work, and it only exists for ingress sampling — `mirror` is invalid while `sFlowEgressRate` is nonzero.

**What**: Adds `fboss2-dev config interface <ports> sflow <attr> <value> [<attr> <value> ...]` and its delete counterpart, for three attributes. Any combination can be set (or deleted) together in one command:

| Attribute | Sets | Delete resets to |
|---|---|---|
| `sample-dest <cpu\|mirror>` | `Port.sampleDest` | unset (agent default) |
| `ingress-rate <N>` | `Port.sFlowIngressRate` | `0` |
| `egress-rate <N>` | `Port.sFlowEgressRate` | `0` |

**Why**: none of these three fields had CLI coverage in either direction.

**How**: `CmdConfigInterfaceSflow` / `CmdDeleteInterfaceSflow` each parse one or more `<attr> <value>` pairs (or `<attr>` names for delete) and apply every attribute present in the call together, the same way `CmdConfigInterface` handles its own attributes (`mtu`, `description`, ...) — rather than a separate leaf class and CLI11 subcommand per attribute, and rather than an if/else chain that would only let one attribute take effect per call. Adding a future sflow attribute is one more branch in the same file.

Validation: rate values must be a non-negative integer. `sample-dest mirror` is refused when the port's `sFlowEgressRate` — from this same command if given, else the port's current value — is nonzero, and symmetrically, a nonzero `egress-rate` is refused when the port's effective `sample-dest` is `mirror`. This is evaluated against the combined result of the whole command, so `sample-dest mirror egress-rate 0` succeeds and `sample-dest mirror egress-rate 50` fails regardless of which attribute is given first. It's the same constraint the agent enforces (`ApplyThriftConfig`: "Egress sampling to mirror destination is unsupported"), caught by the CLI before the config is touched. Hitless: sampling is (re)programmed from the port delta at runtime (`SaiPortManager::addSamplePacket`/`removeSamplePacket`).

## Sample usage

```
$ fboss2-dev config interface eth1/1/1 sflow sample-dest cpu ingress-rate 100 egress-rate 50
Successfully set sFlow sample-dest=cpu, ingress-rate=100, egress-rate=50 for interface(s) eth1/1/1
$ fboss2-dev config interface eth1/2/1 sflow sample-dest mirror egress-rate 50
Port eth1/2/1: sample-dest mirror requires sFlowEgressRate 0 — egress sampling to a mirror destination is unsupported
$ fboss2-dev delete interface eth1/1/1 sflow sample-dest ingress-rate egress-rate
Reset sFlow sample-dest, ingress-rate, egress-rate for interface(s) eth1/1/1
```

## Changed files

| File | Change |
|---|---|
| `commands/config/interface/sflow/CmdConfigInterfaceSflow.{h,cpp}` | Parses one or more `<attr> <value>` pairs, applies all attributes present together |
| `commands/delete/interface/sflow/CmdDeleteInterfaceSflow.{h,cpp}` | Parses one or more `<attr>` names, resets all present together |
| `CmdListConfig.cpp` | `sflow` is now a leaf command (was a stub + one nested subcommand per attribute) |
| `test/config/CmdConfigInterfaceSflowSampleDestTest.cpp` | Unit tests for all three attributes, both commands, combined-attribute and cross-validation cases |
| `test/integration_test/ConfigSflowSampleDestTest.cpp` | End-to-end: `sample-dest` set/delete round trip |
| `test/integration_test/ConfigSflowRateTest.cpp` | End-to-end: combined `ingress-rate`+`egress-rate` set/delete round trip |

# Test Plan

Unit tests (29, config + delete, covering all three attributes individually, combined in one call, and the order-independent cross-attribute mirror/egress-rate validation): all pass.

Full config suite: `//fboss/cli/fboss2/test/config:cmd_config_test PASSED`.

Integration tests on a hardware DUT — `sample-dest` (set cpu, verify running config, delete, verify unset) and combined `ingress-rate`+`egress-rate` (set both in one command, verify running config, delete both in one command, verify reset to 0):

```
[ RUN      ] ConfigSflowRateTest.SetThenDeleteIngressAndEgressRateCombined
[       OK ] ConfigSflowRateTest.SetThenDeleteIngressAndEgressRateCombined
[ RUN      ] ConfigSflowSampleDestTest.SetThenDeleteSampleDest
[       OK ] ConfigSflowSampleDestTest.SetThenDeleteSampleDest
[  PASSED  ] 2 tests.
```

Agents stayed active throughout; DUT config verified restored to its pre-test state.


fboss/cli/fboss2/commands/config/interface/sflow/sample_dest/CmdConfigInterfaceSflowSampleDest.h
```diff
@@ -0,0 +1,72 @@
...
+        "sample_dest",
+        args,
+        "<cpu|mirror> - where sFlow samples are sent: processed on-box (cpu) "
+        "or sent to the port's ingress mirror (mirror)");
+  }
+  using ObjectArgType = SampleDestValue;
+  using RetType = std::string;
+};
+
+class CmdConfigInterfaceSflowSampleDest
```
joseph5wu: Instead of creating a standalone `CmdConfigInterfaceSflowSampleDest`, can we have `CmdConfigInterfaceSflow` to handle all sflow related attributes in the same class:
- sample_dest
- ingress_rate
- egress_rate?

joseph5wu (changes requested):
Let's follow how we handle multiple interface attributes in CmdConfigInterface and do the same thing for `CmdConfigInterfaceSflow` rather than create three different classes.

fboss/cli/fboss2/commands/config/interface/sflow/sample_dest/CmdConfigInterfaceSflowSampleDest.h
vybhav-nexthop: will add these 3 here in this PR itself, since they seemed to be simple enough to implement

vybhav-nexthop:
Added ingress-rate and egress-rate too, in 0b89a7e: `config/delete interface <ports> sflow ingress-rate|egress-rate <N>`, mapping to Port.sFlowIngressRate/sFlowEgressRate, both handled by the same CmdConfigInterfaceSflow/CmdDeleteInterfaceSflow class as sample-dest. egress-rate carries the same MIRROR/egress-rate constraint sample-dest already enforces, checked from the other direction (refuses a non-zero egress-rate on a port whose sample-dest is already MIRROR).

All three attributes now go through one class each on the config/delete side, per your original comment.

fboss/cli/fboss2/commands/config/interface/sflow/CmdConfigInterfaceSflow.cpp
```diff
@@ -0,0 +1,198 @@
...
+          *port->sampleDest() == cfg::SampleDestination::MIRROR) {
+        throw std::invalid_argument(
+            fmt::format(
+                "Port {}: egress-rate must be 0 while sample-dest is {} — "
+                "egress sampling to a mirror destination is unsupported",
+                *port->name(),
+                kSampleDestMirror));
+      }
+      port->sFlowEgressRate() = rate;
+    }
```
joseph5wu: Instead of using if-else, can you support just if independently for the three attributes?
This way we can support a combined command like `fboss2-dev config interface XXX sflow sample-dest CPU ingress-rate XXX egress-rate YYY`?

fboss/cli/fboss2/commands/config/interface/sflow/CmdConfigInterfaceSflow.cpp
```diff
@@ -0,0 +1,198 @@
...
+      : (attr == kAttrIngressRate ? "ingress-rate" : "egress-rate");
+  std::string message = fmt::format(
+      "Successfully set sFlow {} for interface(s) {} to {}",
+      attrLabel,
+      folly::join(", ", updatedNames),
+      displayValue);
+  if (!skippedNames.empty()) {
+    message +=
+        fmt::format("; skipped (no port): {}", folly::join(", ", skippedNames));
+  }
```
joseph5wu: The benefit to have the same CmdConfigInterfaceSflow.cpp to control three attributes is to program the three attributes at the same time. Let's make sure we don't make it exclusive

joseph5wu (commented):
Overall it looks pretty good to me. Just make sure you can support setting multiple attributes at the same time like CmdConfigInterface.cpp

fboss/cli/fboss2/commands/config/interface/sflow/CmdConfigInterfaceSflow.cpp
vybhav-nexthop: got it , will do it the next commit . Thanks

PR #1540, opened 2026-08-20
[Nexthop][m4062nhp] Use stable paths for DIMM temp sensors

anna-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

The DDR5 SPD hubs are auto-instantiated by the kernel and the i2c paths may change depending on the module load order. Resolve the DIMM temp sensor to a stable path provided by the BSP.

# Test Plan

Verified on hardware.

### DIFF
--- a/fboss/platform/configs/m4062nhp/platform_manager.json
+++ b/fboss/platform/configs/m4062nhp/platform_manager.json
@@ -69,11 +69,11 @@
       "embeddedSensorConfigs": [
         {
           "pmUnitScopedName": "DIMM1_TEMP",
-          "sysfsPath": "/sys/bus/i2c/devices/1-0050"
+          "sysfsPath": "/run/nexthop_bsp/sensors/DIMM1_TEMP"
         },
         {
           "pmUnitScopedName": "DIMM2_TEMP",
-          "sysfsPath": "/sys/bus/i2c/devices/1-0051"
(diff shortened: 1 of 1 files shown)


fboss/platform/configs/m4062nhp/platform_manager.json
```diff
@@ -69,11 +69,11 @@
...
         {
           "pmUnitScopedName": "DIMM1_TEMP",
-          "sysfsPath": "/sys/bus/i2c/devices/1-0050"
+          "sysfsPath": "/run/nexthop_bsp/sensors/DIMM1_TEMP"
         },
         {
           "pmUnitScopedName": "DIMM2_TEMP",
-          "sysfsPath": "/sys/bus/i2c/devices/1-0051"
+          "sysfsPath": "/run/nexthop_bsp/sensors/DIMM2_TEMP"
         }
```
somasun, line 77: This shouldn't have slipped in. Thanks for the fix.  Wondering if we can update platform_manager/ConfigValidator to allow only stable paths?

anna-nexthop: @somasun I think that's an improvement if that is indeed the intention. However, I'm hesitant about adding it as part of this change since that's a narrowing requirement and I don't know how many other platforms might be flagged, so proposing to have that improvement be in its own PR.

PR #1542, opened 2026-08-20
[Nexthop][fboss2-dev] Add hitless/warmboot/coldboot options for config reload

benoit-nexthop:
# Summary

Add an ending token for `hitless|warmboot|coldboot` to the `fboss2-dev config reload` command, with hitless being the default (existing) behavior if unspecified. This allows loading an arbitrary agent config and reliably doing a cold agent restart using the new config.

# Test Plan

Added new unit tests.

Manually verified:

```
# fboss2-dev config reload
Config reloaded successfully

# fboss2-dev config reload hitless
Config reloaded successfully

# fboss2-dev config reload warmboot
I0521 21:58:49.517317  4001 SystemdInterface.cpp:85] fboss_hw_agent@0 is now active
I0521 21:58:49.528407  4001 SystemdInterface.cpp:85] fboss_sw_agent is now active
Config reloaded successfully via fboss_hw_agent@0 (warmboot), fboss_sw_agent (warmboot) restart.

# fboss2-dev config reload coldboot
I0521 21:56:00.869602  3598 SystemdInterface.cpp:85] fboss_hw_agent@0 is now active
I0521 21:56:00.880660  3598 SystemdInterface.cpp:85] fboss_sw_agent is now active
Config reloaded successfully via fboss_hw_agent@0 (coldboot), fboss_sw_agent (coldboot) restart.
```


benoit-nexthop:
Original change by @rtl-nexthop

cmake/AgentHwSaiStoreTest.cmake
```diff
@@ -35,7 +35,6 @@ add_executable(store_test
     fboss/agent/hw/sai/store/tests/UdfStoreTest.cpp
     fboss/agent/hw/sai/store/tests/VlanStoreTest.cpp
     fboss/agent/hw/sai/store/tests/WredStoreTest.cpp
-    fboss/agent/hw/sai/store/tests/UdfStoreTest.cpp
```
joseph5wu, line 38: Why this PR is trying to remove a bunch of cpp files?

fboss/cli/fboss2/commands/config/CmdConfigReload.h
```diff
@@ -10,16 +10,46 @@
...
+
+ private:
+  cli::ConfigActionLevel level_ = cli::ConfigActionLevel::HITLESS;
+};
+
 struct CmdConfigReloadTraits : public WriteCommandTraits {
-  using ObjectArgType = std::monostate;
+  static void addCliArg(CLI::App& cmd, std::vector<std::string>& args) {
+    cmd.add_option(
+        "boot_type",
```
joseph5wu, line 41: Since now we can support config bgp, do we also need to consider specifying which service to reload their configs? Based on the implementation here, we can only support reload agent config?

joseph5wu (changes requested):
Please don't remove cpp files from the targets without any reason.
Also should we also specify reload which service's config?

cmake/AgentHwSaiStoreTest.cmake
rtl-nexthop: Hi Joseph, sorry for the noise -- there were some issues creating this upstream PR and some unrelated changes filtered in. I'll be pushing updates to get rid of these.

fboss/cli/fboss2/commands/config/CmdConfigReload.h
rtl-nexthop: Currently, yes only sw/hw agents are supported. It's probably worth a separate discussion on how you want to approach other agents like bgp (which as I understand has separate CLI already). Also I wonder if warmboot/coldboot semantics work for other agents as well?

rtl-nexthop:
Pushed two commits fixing a bug in the original change: **`config reload warmboot` cold booted both agents in split mode** while still reporting success.

`getServicesToRestart()` returns `[fboss_hw_agent@N..., fboss_sw_agent]` and `performWarmboot()` ran `systemctl restart` on each in turn, so a hw agent went down while the sw agent was still up. The sw agent treats a vanishing hw agent as a crash: `HwSwitchConnectionStatusTable::disconnected()` writes `sw_cold_boot_once` and `hw_cold_boot_once_<idx>`, forcing both sides cold. Its own comment states the contract the CLI was violating — *"In normal shutdown sequence, we exit SwSwitch first before shutting down HwSwitch."*

Restart order is now stop-all-then-start-all:

    stop sw -> await exit -> stop hw -> await exit
    -> start hw -> await ready -> start sw

- `SystemdInterface::waitForServiceInactive()`, mirroring `waitForServiceActive`.
- `FbossServiceUtil::performStopAllThenStart()`, shared by warmboot and coldboot. Coldboot passes marker-file creation as a between-phases hook, so markers are written once every service is down and cannot be clobbered by a service's own shutdown. Single-service lists (monolithic `wedge_agent`, `bgpd`) keep the existing `restartService` path.
- systemd does not restart a unit that was explicitly stopped, so a throw between the phases would leave the box with no agents at all; `performStopAllThenStart` best-effort starts everything back up before rethrowing.
- After the stop phase, warmboot logs an error naming any service that left no `can_warm_boot` flag behind — i.e. was SIGKILLed past `TimeoutStopSec` and would silently cold boot. Diagnostic only.

The second commit raises the integration-test `waitForAgentReady()` default from 120s to 180s: the new sequence takes longer to reach thrift-ready, and `ConfigReloadTest`'s `Warmboot`/`Coldboot` cases call it bare right after triggering the restart.

Validated on hardware: `config reload warmboot` now warm boots both agents in split mode, where previously both came up cold. Unit tests for the new ordering added in `ConfigSessionSystemdTest.cpp`.

PR #1548, opened 2026-08-24
[Nexthop][fboss2-dev] Rename copp cpu-queue to queue, full queue attributes, reason ordering, per-queue shaping

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

`config copp queue <id>` accepts every `cfg::PortQueue` attribute, through the same
parser and applier `config qos queue-config` uses, and the copp subcommand is
`queue` rather than `cpu-queue`.

```
fboss2-dev config copp queue 2 scheduling wrr weight 4
fboss2-dev config copp queue 2 rate-limit pps 1000 50000
fboss2-dev config copp queue 5 stream-type multicast weight 3
fboss2-dev config copp reason arp queue 9 order 0
fboss2-dev config qos queue-config qph queue-id 0 name queue_per_host
```

## Background: what a queue is and what these attributes control

Every port drains through a set of egress queues, and a queue is identified by
(`streamType`, `id`) — not by `id` alone. Front-panel queues come from
`sw.portQueueConfigs` (named lists) or `sw.defaultPortQueues` (the fallback for
ports with no `portQueueConfigName`); CPU queues come from `sw.cpuQueues`. All
three hold the same `cfg::PortQueue`, which is why one parser and one applier can
serve both command families.

The attributes group by what they reach in hardware:

- **Scheduling** — `scheduling` selects strict priority or DWRR
  (`SAI_SCHEDULER_ATTR_SCHEDULING_TYPE`). `weight` is the DWRR share
  (`SAI_SCHEDULER_ATTR_SCHEDULING_WEIGHT`), ignored under strict priority.
- **Buffering** — all of these land in a buffer profile. `reserved-bytes` is the
  queue's guaranteed allocation; `shared-bytes` is a *static* maximum threshold,
  and its mere presence switches the profile from dynamic to static;
  `scaling-factor` is the MMU alpha that drives the dynamic threshold instead;
  `max-dynamic-shared-bytes` caps that dynamic threshold; `buffer-pool-name` forces
  the queue onto a non-default egress pool.
- **Shaping** — `rate-limit <kbps|pps> [<min>] <max>` becomes the scheduler's meter
  type plus `MIN_BANDWIDTH_RATE` / `MAX_BANDWIDTH_RATE`, with kbits converted to
  bytes per second.
- **AQM** — `active-queue-management congestion-behavior <ECN|EARLY_DROP>` plus a
  linear detection profile becomes a WRED profile: green min/max thresholds and
  drop probability for `EARLY_DROP`, ECN mark mode and thresholds for `ECN`. A
  queue may carry one entry of each, so an edit is keyed by behavior rather than
  overwriting `aqms[0]`.
- **`name`** — never programmed to hardware; it names the queue's fb303 stats. On
  the CPU port an unnamed queue is left out of `configuredQueues` entirely, so it
  collects no stats at all.

## What changes

1. **Rename `cpu-queue` to `queue`** across config and delete — hard rename, no
   alias; classes and directories follow. copp already means control plane, so the
   `cpu-` prefix said nothing.
2. **`config copp queue <id>` takes every `cfg::PortQueue` attribute**, through the
   same parser and applier `config qos queue-config` uses. `cpuQueues` holds the
   same struct, so the qos helper applies unchanged. The [0, 255] weight bound
   lives in that shared applier, so both families enforce it.
3. **Creating a CPU queue requires `stream-type`.** A queue is identified by
   (`streamType`, `id`) and the correct value is per-platform, so the CLI refuses to
   guess; editing an existing queue is unaffected.
4. **`config copp reason <r> queue <id> order <n>`** places the entry at 0-based
   index n of the position-sensitive `rxReasonToQueueOrderedList`. Without `order`,
   existing reasons keep their position and new ones append.
5. **`rate-limit`, `name` and `max-dynamic-shared-bytes` are shared attributes.**
   `rate-limit <kbps|pps> [<min>] <max>` writes the `PortQueueRate` union; `<min>`
   is optional because that is the form `config copp queue` already shipped.
6. **One `commands/config/QueueConfigUtils.{h,cpp}`** holds the shared queue-name
   resolution, the attribute grammar and the applier, next to its only users
   (`commands/config/copp/`, `commands/config/qos/`), so the two families cannot
   drift apart.
7. **Integration-test surface stays flat.** New behavior is covered by unit tests
   against the same `cfg::PortQueue` the commands write; the hardware cases assert
   only what a unit test cannot — that the agent accepts the config on commit and
   the running config round-trips it.

## Note on `bandwidth-burst`

Deliberately not offered. `PortQueue` carries the two burst fields and
`ApplyThriftConfig` stores them, but the SAI scheduler is created from a tuple with
no burst member, so they would reach the running config and change nothing on the
wire. `rate-limit` is unaffected — both bandwidth rates are in that tuple.

# Test Plan

**Unit**, full `cmd_config_test` suite:

```
[==========] 905 tests from 76 test suites ran. (67898 ms total)
[  PASSED  ] 905 tests.
```

Coverage: both `rate-limit` forms, the optional-`<min>` lookahead against a
trailing attribute, inverted ranges, bad units, per-attribute arity, `name` and
`max-dynamic-shared-bytes` on both paths, the empty-name rejection, the
duplicate-attribute guard, `bandwidth-burst` being refused, and — for CPU queues —
creation without `stream-type` refused and creation with it landing the typed entry.

**Integration**, on an NH-4010-F DUT. The hardware cases assert what a unit test
cannot: that the agent accepts each commit and the running config round-trips the
value. `QueueConfigCreateShapeAndDelete` builds a throwaway named queue-config with
a shaped queue, verifies it in the running config, then deletes the whole entry and
verifies the agent accepts its absence. The copp cases edit an existing CPU queue
and restore it; `DeleteCoppTest` creates its own scratch queue — passing
`stream-type`, read off the box's existing queues — and removes it afterwards.

```
$ fboss2_integration_test --gtest_filter='ConfigCoppTest.*:DeleteCoppTest.*:ConfigPortQueueConfigTest.*'
[==========] Running 10 tests from 3 test suites.
[----------] 6 tests from ConfigCoppTest
[       OK ] ConfigCoppTest.CpuQueueEnsureExistsIsNoOp (52 ms)
[       OK ] ConfigCoppTest.CpuQueueSetName (161 ms)
[       OK ] ConfigCoppTest.CpuQueueSetRateLimitPps (152 ms)
[       OK ] ConfigCoppTest.ReasonToQueueUpdate (154 ms)
[       OK ] ConfigCoppTest.CpuQueueSetSchedulingAndWeight (189 ms)
[       OK ] ConfigCoppTest.ReasonOrderMove (145 ms)
[----------] 2 tests from ConfigPortQueueConfigTest
[       OK ] ConfigPortQueueConfigTest.BuildAndAssignPolicy (22050 ms)
[       OK ] ConfigPortQueueConfigTest.QueueConfigCreateShapeAndDelete (43867 ms)
[----------] 2 tests from DeleteCoppTest
[       OK ] DeleteCoppTest.DeleteWholeCpuQueue (154 ms)
[       OK ] DeleteCoppTest.DeleteReasonMapping (229 ms)
[==========] 10 tests from 3 test suites ran. (89395 ms total)
[  PASSED  ] 10 tests.
```

The DUT was left as found: `fboss_sw_agent` reports `NRestarts: 0`,
`config session diff` reports no session, and the running config carries no
leftover `cli_e2e_test_*` entries.

Validation ran in our internal tree, which carries this same change. The upstream
tree itself was not built locally; this PR's CI is the build verification.


vybhav-nexthop:
Pushed a review-response commit (`Address review findings on the queue command`) fixing what a pre-publication review pass found:

- A queue created by the CLI now seeds `streamType` from the box's existing `cpuQueues` (MULTICAST when there are none). The agent's `updatePortQueues` filters entries by streamType, so a thrift-default UNICAST entry committed fine but was silently never installed.
- Create-plus-edit is one transaction: a rejected attribute no longer leaves a half-built `{id}` entry in the session.
- Bare `aqm` and repeated `name`/`rate-limit` are parse errors instead of silent drops.
- The WRR weight bound moved into `utils::applyPortQueueConfig` as [0, 255], so `config qos queue-config` gets it too; weight 0 stays legal (SAI coerces it to 1).
- The generic `<attr> <value>` walk is factored into `utils::walkQueueAttributes`, shared by the copp and qos parsers; the copp help text is built from `utils::validQueueAttrs()` so it cannot drift.
- The queue option registers with `required()+expected()` like the reason command. Values past the first token that spell sibling command names can still be reclassified by CLI11; that residual is documented in the code.
- Removed the stale pre-rename files (`CmdDeleteCoppCpuQueue.{cpp,h}` + test) the merge from main resurrected.

Unit suite green on the identical internal source: 70 tests across the copp and qos fixtures, including 6 new cases for the above.

vybhav-nexthop:
DUT validation of the review-response commit, Tomahawk4 device:

All 9 copp integration tests pass (`ConfigCoppTest.*:DeleteCoppTest.*`, 1.7s total).

The streamType fix verified end-to-end: `config copp queue 3 rate-limit pps 500` on a previously nonexistent id, committed, and the running config shows the new queue with `streamType: 1` (MULTICAST) matching its siblings plus the pps cap — under the old code this entry would have defaulted to UNICAST and never been installed.

Bonus confirmation of the division of labor: creating queue id 11 passes the CLI and the agent rejects the commit with "Port queue config listed for invalid queues. Maximum number of queues on this platform is 10", rolling back automatically. Test queue deleted and config restored afterward.

fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
```diff
@@ -167,36 +166,83 @@ CoppReasonArgs::CoppReasonArgs(std::vector<std::string> v) {
...
+    work = *it;
+  } else {
+    work.id() = args.getQueueId();
+    // The agent only programs cpuQueues entries whose streamType matches the
+    // CPU port's stream types (ApplyThriftConfig's updatePortQueues filters
+    // on it); the thrift default is UNICAST while CPU queues ship MULTICAST,
+    // so a defaulted entry would commit fine and silently never be installed.
+    // Follow the box's existing queues; MULTICAST when there are none.
+    work.streamType() = queues.empty() ? cfg::StreamType::MULTICAST
+                                       : *queues.front().streamType();
```
joseph5wu: Unfortunately this is not the correct assumption about CPU queue only uses MULTICAST.
Based on https://github.com/facebook/fboss/blob/main/fboss/agent/ApplyThriftConfig.cpp#L6243, current FBOSS can support different streamtypes for cpu queues based on the asic type.

Conceptually, a queue is identified by (streamType, queueId), not merely queueId. Therefore, I would recommend to enforce users to specify both queue id and stream type when trying to create a new cpu queue.

Currently CPU ports generally expose one:
  - Broadcom XGS/Tomahawk: MULTICAST
  - Jericho/Chenab: UNICAST
  - Some older platforms: ALL

fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
```diff
@@ -93,68 +82,78 @@ void applyRateLimit(cfg::PortQueue& q, CoppCpuQueueArgs::Op op, int32_t max) {
...
+            throw std::invalid_argument(
+                fmt::format(
+                    "rate-limit unit must be '{}' or '{}', got '{}'",
+                    kRateUnitKbps,
+                    kRateUnitPps,
+                    unit));
+          }
+          rateLimit_ = {rateUnit, parseRateMax(v[i + 2], unit)};
+          return 3;
+        }
```
joseph5wu: Since this PR is trying to make `applyPortQueueConfig` to be more generic to handle port and cpu queue config, why not move all these rateLimit attributes to `applyPortQueueConfig`. And you can keep different `validQueueAttrs` for port and cpu queue?

fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
```diff
@@ -167,36 +166,83 @@ CoppReasonArgs::CoppReasonArgs(std::vector<std::string> v) {
...
+    work.name() = *args.getName();
+  }
+  if (args.getRateLimit().has_value()) {
+    applyRateLimit(
+        work, args.getRateLimit()->first, args.getRateLimit()->second);
+  }
+  if (!args.getAttributes().empty() || !args.getAqmAttributes().empty()) {
+    utils::applyPortQueueConfig(
+        work, args.getAttributes(), args.getAqmAttributes());
+  }
```
joseph5wu: As comments above, let's consider to move all port queue related config in `applyPortQueueConfig`

fboss/cli/fboss2/commands/config/QueueConfigUtils.h
```diff
@@ -0,0 +1,106 @@
...
+
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "fboss/agent/gen-cpp2/switch_config_types.h"
+#include "fboss/cli/fboss2/utils/CmdUtilsCommon.h"
+
+namespace facebook::fboss::utils {
```
joseph5wu, line 22: Can we keep this utils files in fboss2/commands/config instead? Since this is only used by copp/ and qos/ folder in the same config/

fboss/cli/fboss2/commands/config/qos/QueueConfigUtils.cpp
```diff
@@ -0,0 +1,84 @@
...
+      // "port-<logicalID>", so use that rather than reporting an empty name.
+      if (port.name().has_value()) {
+        portNames.push_back(*port.name());
+      } else {
+        portNames.push_back(fmt::format("port-{}", *port.logicalID()));
+      }
+    }
+  }
+  return portNames;
+}
```
joseph5wu: Can we keep just one QueueConfigUtils? Right now we have PortQueueConfigUtils and this qos/QueueConfigUtils?

joseph5wu (changes requested):
Please address my comments

fboss/cli/fboss2/commands/config/QueueConfigUtils.h
vybhav-nexthop: done kept it under config

fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
vybhav-nexthop: so i decided to share all attributes for cpu and port, since they dont make any difference for the asic

fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
vybhav-nexthop: done, we now have one queue config utils file for all

fboss/cli/fboss2/commands/config/qos/QueueConfigUtils.cpp
vybhav-nexthop: yeh we will have only one , as you mentioned

fboss/cli/fboss2/commands/config/copp/CmdConfigCopp.cpp
vybhav-nexthop: changed it, mandated the user to give a stream type too

PR #1549, opened 2026-08-24
[Nexthop][fboss2-dev] Add acl rule lookup-class match fields

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

Three new match fields on `config acl rule`:

```
config acl rule <table> <rule> lookup-class-l2       <class-id>
config acl rule <table> <rule> lookup-class-neighbor <class-id>
config acl rule <table> <rule> lookup-class-route    <class-id>
```

They set `AclEntry.lookupClassL2` / `lookupClassNeighbor` / `lookupClassRoute`, the qualifiers a queue-per-host config matches on (e.g. a `queue-per-host-queue-0-l2` rule paired with a send-to-queue action).

Class ids are spelled the same way `config interface <intf> lookup-class` takes them: a numeric id (`10`) or a class name (`CLASS_QUEUE_PER_HOST_QUEUE_0`, case-insensitive). That command's parser moves to `utils/LookupClassUtils.{h,cpp}` and both commands share it, so one place decides which classes an operator may name and rejects the agent-internal ones (`CLASS_DROP`, `DST_CLASS_L3_LOCAL_*`).

The new attrs are rows in the existing grammar table; `--help` and arity errors pick them up without CLI11 changes.

# Test Plan

Unit (`//fboss/cli/fboss2/test/config:cmd_config_test`): full suite passes. New cases: each attr by id and by name, agent-reserved class rejected, unknown id/name rejected. The interface lookup-class tests still pass on the shared parser.

Integration (`ConfigAclRuleTest`, Tomahawk4): 27/27. The three new Set cases stage the attr, commit, read back the running config, then delete the rule and commit again.

Validated on the identical source in our internal tree


cmake/AgentHwSaiStoreTest.cmake
```diff
@@ -35,7 +35,6 @@ add_executable(store_test
     fboss/agent/hw/sai/store/tests/UdfStoreTest.cpp
     fboss/agent/hw/sai/store/tests/VlanStoreTest.cpp
     fboss/agent/hw/sai/store/tests/WredStoreTest.cpp
-    fboss/agent/hw/sai/store/tests/UdfStoreTest.cpp
```
joseph5wu: Please don't include any outside of fboss2/ changes in this PR

cmake/AgentState.cmake
```diff
@@ -126,7 +126,6 @@ add_library(state
   fboss/agent/state/UdfGroupMap.cpp
   fboss/agent/state/UdfPacketMatcher.cpp
   fboss/agent/state/UdfPacketMatcherMap.cpp
-  fboss/agent/state/IpTunnel.cpp
```
joseph5wu: Please don't include any outside of fboss2/ changes in this PR

cmake/AgentTestAgentHwTests.cmake
```diff
@@ -218,7 +218,6 @@ add_library(agent_hw_test_src
   fboss/agent/test/agent_hw_tests/AgentRollbackTests.cpp
   fboss/agent/test/agent_hw_tests/AgentRouteRollbackTests.cpp
   fboss/agent/test/agent_hw_tests/AgentPacketStreamHandlerTests.cpp
-  fboss/agent/test/agent_hw_tests/AgentAclTableGroupTests.cpp
```
joseph5wu: Please don't include any outside of fboss2/ changes in this PR

cmake/ThriftCowVisitors.cmake
```diff
@@ -33,7 +33,6 @@ add_library(
   thrift_cow_visitors
   fboss/thrift_cow/visitors/DeltaVisitor.h
   fboss/thrift_cow/visitors/ExtendedPathVisitor.h
-  fboss/thrift_cow/visitors/ExtendedPathVisitor.h
```
joseph5wu: Please don't include any outside of fboss2/ changes in this PR

joseph5wu (changes requested):
The PR related to lookup-class looks good to me.
But please remove all the changes unrelated to this PR

PR #1550, opened 2026-08-24
[Nexthop][fboss2-dev] One traffic-policy implementation behind a copp and a data-plane verb

vybhav-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

## What changed

FBOSS has two traffic policies: one for CPU-bound traffic (`cpuTrafficPolicy`) and one for the data plane (`dataPlaneTrafficPolicy`). Today the CLI can only reach the second, and only sideways — through `config acl rule <table> <rule> action <...>` (#1183), which writes the data-plane policy without ever saying so. There is no way to configure CPU traffic-policy actions at all.

This PR gives each policy its own explicit verb, config and delete, all four running on one shared implementation:

```
config copp       traffic-policy match <rule> action <type> [<value>]
config data-plane traffic-policy match <rule> action <type> [<value>]

delete copp       traffic-policy match <rule> action <type>
delete data-plane traffic-policy match <rule> action <type>
```

It supersedes #1379, which proposed the copp side alone as `config copp cpu-traffic-policy`. That command is folded in here with the `cpu-` prefix dropped: once a data-plane sibling exists, the parent word (`copp` vs `data-plane`) already says which policy you mean.  This is also the direction #1379's review settled on: rich ACL actions belong on policy-named verbs, not on `config acl rule` — see the removal section below.

## Usage: a rule and its action

A `cfg::AclEntry` is the rule — match conditions, plus `actionType`, which is only permit, deny or deny-data-and-control-plane. Everything richer (queue, DSCP, mirror, counter, to-CPU, redirect) is a `cfg::MatchAction`, stored separately and tied to the rule by name:

```thrift
struct MatchToAction {
  1: string matcher;        // the AclEntry's name
  2: MatchAction action;
}
```

Those entries live in one of two lists, `sw.cpuTrafficPolicy.trafficPolicy` or `sw.dataPlaneTrafficPolicy`, and a rule can be in only one of them.

End to end:

```
# a table group, and a table inside it
fboss2-dev config acl table-group acl-table-group-ingress stage ingress
fboss2-dev config acl table AclTable1 group acl-table-group-ingress priority 3

# the rule: match conditions, and permit/deny if you want one
fboss2-dev config acl rule AclTable1 my-rule destination-ip 10.0.0.0/24

# the action, on the policy it belongs to
fboss2-dev config data-plane traffic-policy match my-rule action send-to-queue 7
```

## Why the policy choice is behavior, not naming

`ThriftConfigApplier::updateAclsImpl` looks a rule up in the CPU policy first, then falls back to the data-plane policy, and passes the answer down as `isCoppAcl`. Downstream the same queue id lands on a different queue depending on that flag. On BCM it is literal: `BcmAclEntry` passes one `queueId` to either `bcmFieldActionCosQNew` or, when the flag is set, `bcmFieldActionCosQCpuNew` plus `bcmFieldActionCopyToCpu`. On SAI it is indirect: `SaiAclTableManager` turns the queue id into a traffic class and sets the same `ActionSetTC` either way, adding a copy/trap action when the flag is set, so the number resolves through the CPU port's TC-to-queue map instead of a front-panel port's. `isCoppAcl` also picks the priority space: `aclPriority = isCoppAcl ? cpuPriority++ : dataPriority++`, so the two policies number their ACLs independently.

Two consequences shaped the design:

- These actions cannot live on the `AclEntry` itself. Without knowing which policy the rule belongs to, a queue id is ambiguous. The `AclEntry` keeps what is unambiguous: the match conditions and plain permit/deny.
- A rule must not appear in both policies. The lookup order means the data-plane entry would be silently dead. The config commands now refuse this up front instead of letting it happen.

## How the shared implementation works

`commands/config/TrafficPolicyUtils.{h,cpp}` contains everything both verbs need: resolving the right policy for a `PolicyKind`, finding or creating a rule's `MatchToAction` entry, the both-policies guard, one argument parser, and one action table.

The action table is the core of it. Each row pairs an action keyword with its setter (for config) and its reset (for delete), so the two commands can never disagree about which actions exist or what values they take

It also widens coverage from #1379's 4 action keywords to 13, one for each `MatchAction` field that `ApplyThriftConfig` actually reads: `send-to-queue`, `set-dscp`, `set-tc`, `mirror-ingress`, `mirror-egress`, `counter`, `trap-to-cpu`, `copy-to-cpu`, `redirect nexthop`, `user-defined-trap`, `flowlet`, `ecmp-hash`, `alternate-ars-members`. The two remaining fields, `macsecFlow` and `packetCounter_DEPRECATED`, are skipped because the agent never reads them.

Two of the 13 are gated: `ApplyThriftConfig` reads `setTc` and `userDefinedTrap` only under `FLAGS_sai_user_defined_trap`, which defaults false. The CLI still accepts them and writes them to config -- that is the right split, since the config is what a later flag flip would activate -- but on a stock switch neither reaches hardware. Worth knowing before testing `set-tc` on a box that has not enabled the flag.

One validation choice worth calling out: `set-tc` accepts the full thrift byte range rather than 0–7. Real CPU policies pair traffic class 9 with CPU queue 9, and a 0–7 limit would reject them.

## Why `config acl rule ... action <...>` no longer takes rich actions

#1183 made rich actions settable directly on an ACL rule (`config acl rule <table> <rule> action set-dscp 34`). That form has a built-in flaw: it writes to the data-plane policy unconditionally, because nothing in the syntax says which policy is meant. A CPU action simply cannot be expressed through it, and a user configuring CoPP gets a rule that looks right but lands in the wrong policy.

`action trap-to-cpu` is the sharpest case. It sets `toCpuAction` on a data-plane matcher, and `SaiAclTableManager` only reaches `setCopyOrTrap` from the CPU-matched path (and from the user-defined-trap branch) — so that command saved cleanly, reported success, and programmed nothing.

The first round of this PR kept the form working behind a deprecation warning and left removal as a follow-up. Per review, it is removed here instead. `config acl rule ... action` now accepts only the three keywords that live on the `AclEntry` itself — `permit`, `deny`, `deny-data-and-control-plane` — and anything else is rejected at parse time with the valid list. Every richer action goes through the two policy verbs above, which name the policy the action belongs to.

That removal simplifies the rule command rather than complicating it: `AclRuleMutation` loses its `MatchAction` half, `CmdConfigAclRule` no longer reaches into a traffic policy at all, and the CLI11 arity drops from `(4, 6)` to `(4, 5)` now that `redirect nexthop <ip>` is gone.

## Guardrails carried over from #1379's review

Deleting an action that was never set reports that rather than failing, and unknown action keywords are rejected at parse time. Delete also prunes a matcher entry once its last action is removed, including pre-existing empty entries. One CLI11 quirk is handled explicitly: a matcher literally named `action` would be swallowed as the subcommand token, making it undeletable, so the config side rejects that name up front.

## The commands in action

Captured on a test switch running this branch. Setup — two match-only rules in the existing ingress table:

```
$ fboss2-dev config acl rule AclTable1 cpu-bgp-rule source-ip 10.0.0.0/24
Created and set acl rule 'cpu-bgp-rule' (table 'AclTable1', group 'acl-table-group-ingress') source-ip to 10.0.0.0/24

$ fboss2-dev config acl rule AclTable1 web-traffic-rule dscp 10
Created and set acl rule 'web-traffic-rule' (table 'AclTable1', group 'acl-table-group-ingress') dscp to 10
```

An action on each policy:

```
$ fboss2-dev config copp traffic-policy match cpu-bgp-rule action send-to-queue 4
Set copp traffic-policy action 'send-to-queue' 4 on matcher 'cpu-bgp-rule'

$ fboss2-dev config data-plane traffic-policy match web-traffic-rule action set-dscp 34
Set data-plane traffic-policy action 'set-dscp' 34 on matcher 'web-traffic-rule'
```

The both-policies refusal — `cpu-bgp-rule` already has copp actions:

```
$ fboss2-dev config data-plane traffic-policy match cpu-bgp-rule action set-tc 3
Thrift call failed: 'Rule 'cpu-bgp-rule' already has actions under the copp traffic policy. A rule resolves to one policy only (the CPU one wins), so the data-plane entry would never take effect; delete the other one first'
```

Delete work

```
$ fboss2-dev delete copp traffic-policy match cpu-bgp-rule action send-to-queue
Successfully deleted action 'send-to-queue' from copp traffic-policy matcher 'cpu-bgp-rule'

$ fboss2-dev delete data-plane traffic-policy match web-traffic-rule action set-tc
Action 'set-tc' is already absent for matcher 'web-traffic-rule'
```

After deleting everything added above, the running config was byte-identical to the pre-test snapshot.

The transcripts above were captured before the review fixes, on the revision that still carried the
deprecated rule-side form; the commands shown are unchanged by those fixes. The acl-rule test file
gained a case asserting each of the nine removed action keywords is now rejected, and the two
integration tests that used the removed form to seed a matcher now seed it through
`config data-plane traffic-policy match ... action ...`.

# Test Plan

Unit tests, `//fboss/cli/fboss2/test/config:cmd_config_test` (`CmdConfigCoppTest`, `CmdConfigDataPlaneTrafficPolicyTest`, `CmdDeleteCoppTrafficPolicyMatchActionTest`, `CmdDeleteDataPlaneTrafficPolicyMatchActionTest` cover every action shape including the valueless and two-token forms, multiple actions stacking on one matcher, unknown keywords, wrong arity, `redirect` without its `nexthop` keyword, the both-policies refusal, delete removing one action and then the whole entry, delete of an absent action being reported, and pruning of pre-existing empty matchToAction entries):

```
[==========] 953 tests from 89 test suites ran. (82655 ms total)
[  PASSED  ] 953 tests.
```

Integration, on a Tomahawk test switch running this branch — the suite drives the real CLI against the live agent through an add → verify → delete → verify cycle and restores the pre-test config afterwards:

```
[ RUN      ] ConfigCoppTrafficPolicyTest.addAndDeleteAction
[       OK ] ConfigCoppTrafficPolicyTest.addAndDeleteAction (22275 ms)
[==========] 1 test from 1 test suite ran. (44375 ms total)
[  PASSED  ] 1 test.
```

## How to read this

41 files overstates it. Two of them carry the whole design (~750 lines), 18 are CLI11 subcommand
scaffolding of 19-58 lines each, 6 are tests, and the rest is registration and file lists. Net
+3274/-475, where the deletions are mostly the rule-side action rows this PR removes. Two commits:
the feature, then the review fixes.

1. **`commands/config/TrafficPolicyUtils.h`, then `.cpp`.** The design is entirely here. The header is the
   contract; the action table in the `.cpp` is the one place that decides which actions exist, what
   values each takes, and how each is reset. Get these two right and the rest follows.
2. **"Why the policy choice is behavior, not naming", below, before judging any of it.** The
   `isCoppAcl` asymmetry in `updateAclsImpl` is why these actions cannot sit on `AclEntry`, and why
   a rule appearing in both policies is refused up front.
3. **`commands/config/acl/rule/CmdConfigAclRule.cpp` and `AclRuleAttrs.{h,cpp}`.** The removal of
   the rule-side rich actions, and why the rule command no longer touches a traffic policy.
4. **One command leaf, then skim the other 17.**
   `commands/delete/copp/traffic_policy/match/action/` is the deepest chain; its data-plane twin is
   the same file with a different `PolicyKind`. Diff one pair and you have seen all four chains.
5. **Tests: check coverage, don't read line by line.**
   `CmdDeleteCoppTrafficPolicyMatchActionTest.cpp` is the thorough one;
   `ConfigCoppTrafficPolicyTest.cpp` is the on-switch add/verify/delete cycle.
6. **Skip unless CI goes red.** `CmdListConfig.cpp` registration plus the cmake and BUCK file lists.


fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
```diff
@@ -107,21 +108,29 @@ CmdConfigAclRuleTraits::RetType CmdConfigAclRule::queryClient(
...
+              << " action ...` (or `config copp "
+                 "traffic-policy ...` for CPU actions) instead"
+              << std::endl;
+    traffic_policy::assertNotInOtherPolicy(
+        swConfig, args.getRuleName(), traffic_policy::PolicyKind::DataPlane);
+    auto& action = traffic_policy::upsertMatcher(
+        traffic_policy::policyFor(
+            swConfig, traffic_policy::PolicyKind::DataPlane),
+        args.getRuleName());
+    args.applyActionTo(action);
```
joseph5wu: Since now we have a clear data-plane and copp traffic policy, I don't think we should keep this.
Let's remove matchAction from this acl rule cli

fboss/cli/fboss2/commands/config/TrafficPolicyUtils.cpp
```diff
@@ -0,0 +1,622 @@
...
+  return false;
+}
+
+// One action keyword: its value arity, a form string for --help and errors, a
+// setter, and a reset used by the delete path. Keeping set and reset in the
+// same row is what stops the two verbs drifting apart.
+struct ActionRow {
+  std::string_view key;
+  std::size_t minVals;
+  std::size_t maxVals;
```
joseph5wu: The minVals and maxVals naming is a bit confusing.
I thought it means the range of the action parameters. 
Based on the `actionRows` always has minVals == maxVals, and it looks like we should just use `std::size_t requiredExtraArgsNum` to make it more clear?
I'm afraid even if you want to support some action can use optional args, your current implementation won't work any way. So let's make it clear with one single `requiredExtraArgsNum`

joseph5wu (changes requested):
Overall this PR looks very solid to me. And this reusable utility makes the whole traffic policy looks so much cleaner.

Please address the two comments

fboss/cli/fboss2/commands/config/acl/rule/CmdConfigAclRule.cpp
vybhav-nexthop: yes removed it

fboss/cli/fboss2/commands/config/TrafficPolicyUtils.cpp
vybhav-nexthop: just created requiredExtraArgsNum and handle the optional part later if needed

PR #1561, opened 2026-08-26
[Nexthop][m4062nhp] Remove compiled-in platform mapping

zackary-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

The compiled-in `M4062nhpPlatformMapping` was only a stub that loaded an empty mapping, so any service that fell back to it started with no ports. Remove the stub and its CMake/BUCK wiring, and make `initPlatformMapping` throw for `PLATFORM_M4062NHP` when no external mapping is given, with the platform name in the error. The mapping now comes from `generated_platform_mappings/nexthop/m4062nhp` via `--platform_descriptor_config_path`.

`M4062nhpLedManager` uses the `init<Bsp>(PlatformType)` overload. The OSS `default_configs/m4062nhp/agent.conf` picks up the new flag and is refreshed to the current Nexthop config (per-port VLANs, management port); a `qsfp.conf` with the same flag is added.

# Test Plan

- OSS CMake build of wedge_agent, qsfp_service and led_service with the `m4062nhp_platform_mapping` target gone.
- On an m4062nhp unit: agent, qsfp_service and led_service start with `--platform_descriptor_config_path` and load the mapping from the descriptor; without the flag they fail with `Platform PLATFORM_M4062NHP requires an external platform mapping`.
- Platform service and led_service HW test suites pass on the same unit (41 tests, 0 failed).


srikrishnagopu:
I have re-imported this but there are build failures @zackary-nexthop. Could you please check ?

zackary-nexthop:
> I have re-imported this but there are build failures @zackary-nexthop. Could you please check ?

@srikrishnagopu the failure was `https://ftpmirror.gnu.org/gnu/binutils/binutils-2.43.tar.xz ...: HTTP Error 502: Bad Gateway`. I've seen that a few times, there might be an issue with the CI.

That said, I've rebased my change and the error doesn't seem to show anymore, should be ok now.

PR #1563, opened 2026-08-27
[Nexthop][fboss2-dev] Add recordConfigBaseline()/rollbackConfig() test fixture helpers

hillol-nexthop:
# Summary

Integration tests that mutate the running config need to leave the DUT as they found it even when the test body fails part-way. Rather than giving `Fboss2IntegrationTest` its own snapshot/restore mechanism (serialize the session config, re-commit it with a hardcoded action level), this adds two thin helpers on top of the CLI's own `config rollback`:

- **`recordConfigBaseline()`** — discards any pending session and returns the `/etc/coop` git HEAD. Constructing the session initializes the repo on a pristine box, so a baseline always exists. Call in `SetUp()`.
- **`rollbackConfig(sha)`** — runs `config rollback <sha>` in-process and waits for the agent; a no-op when HEAD is still at the baseline; a failure to create the rollback commit is an error. Call in `TearDown()`.

Because `ConfigSession::rollback()` re-applies at the highest action level recorded by the commits being undone (#1447), the restore takes exactly the restart the test's own commits did — warmboot for a VLAN membership change, hitless for a CoPP action, coldboot if a commit recorded `AGENT_COLDBOOT`. The test never has to guess.

The dropped-connection tolerance already in `commitConfig()` (an agent restart can cut the thrift call after the commit landed) is factored into `isAgentRestartSignature()` so the rollback path shares it.

No test is switched over in this PR; the first user is the `switchport access vlan` test in the next PR of the stack.

# Test Plan

- Builds as part of `fboss2_integration_test`.
- Exercised on a lab device (NH-4010-F) via the stacked VLAN test: `SetUp` recorded HEAD, two `AGENT_WARMBOOT` commits, `TearDown` rolled back as a new commit whose metadata recorded `AGENT_WARMBOOT`; agents active afterwards; `git diff <baseline> HEAD -- cli/agent.conf` empty. A hitless-only suite using the same helpers rolled back with `HITLESS` in ~80 ms.

Based-on: "main"

### DIFF
--- a/fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
+++ b/fboss/cli/fboss2/test/integration_test/Fboss2IntegrationTest.cpp
@@ -50,6 +50,7 @@
 #include "fboss/cli/fboss2/CmdLocalOptions.h"
 #include "fboss/cli/fboss2/commands/config/vlan/VlanManager.h"
 #include "fboss/cli/fboss2/session/ConfigSession.h"
+#include "fboss/cli/fboss2/session/Git.h"
 #include "fboss/cli/fboss2/utils/CmdClientUtilsCommon.h"
 #include "fboss/cli/fboss2/utils/CmdInitUtils.h"
 #include "fboss/cli/fboss2/utils/HostInfo.h"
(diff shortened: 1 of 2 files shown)


hillol-nexthop:
https://github.com/facebook/fboss/pull/1557 already does the same thing in the test harness.

PR #1576, opened 2026-08-31
[Nexthop] Resolve RX packet's L3 interface from its VLAN, not its port

benoit-nexthop:
## Summary

A trunk port belongs to several VLANs and thus several L3 interfaces, so resolving the ingress interface from the port alone is ambiguous: `Port::getInterfaceID()` CHECK-fails, aborting the sw agent, on the first CPU-punted frame received on such a port. VLAN <-> interface stays 1:1 even under trunking, and the hardware already resolves the ingress VLAN, so resolve from the packet instead:

- **New `getInterfaceIDForPkt()`** (`fboss/agent/Utils.{h,cpp}`): prefer the packet's source VLAN -> `Vlan::getInterfaceID()`; fall back to port-based resolution when the packet carries no VLAN (VOQ/fabric switches) or the VLAN isn't in the switch state.
- **Converted the RX-path resolution sites** to it: `SwSwitch::handlePacket`, IPv4/IPv6 handlers, DHCPv4/v6 handlers.
- **`SwitchState::getInterfaceIDForPortIf()`** now returns `nullopt` (with an error log) instead of aborting when a port has more than one interface; unresolvable packets are dropped and counted, matching the existing no-interface path.
- **HW agent** (`SaiSwitch::packetRxCallbackPort`) stamps the punt's VLAN from its 802.1Q tag when present (the ASIC inserts the resolved ingress VLAN on the punt copy) instead of the static port->PVID map; untagged, priority-tagged, QinQ and unparseable punts fall back to the PVID. Also fixes the RX debug log printing the VLAN captured before resolution.
- **Probe paths** that only know a port degrade gracefully: NDP reachability probes fall back to the cache's own interface; unicast neighbor solicitation is skipped with a debug log when the port is ambiguous.

## Test plan

Tested on a Minipack3 (TH5) with 800G optical ports, connected to a peer switch plus self-loopback cabled port pairs.

Baseline (unpatched): adding 2 VLANs to an up port via `fboss2 config interface <port> switchport trunk allowed vlan add <v1>,<v2>` + commit aborts the sw agent within 37-89s from ambient control-plane traffic alone (`Check failed: intfs.size() == 1 (3 vs. 1)`), re-crashing on the next punted frame after every restart. Confirmed triggers (via hw agent RX hex dumps): ordinary IPv6 NS / ARP / LLDP.

With this change, on a trunk port with native + 2 tagged VLANs (3 interfaces):

- Untagged broadcast ARP injected from the peer switch (previously an instant abort): punted, resolved to the native VLAN's interface, handled, FDB learned.
- 802.1Q-tagged ARP from the peer switch: hw agent derives the VLAN from the frame's tag (`Rx packet on port: 52 vlan: 2010`) and the sw agent resolves the tagged VLAN's interface — correct trunk attribution end-to-end.
- RX injections via thrift `sendPkt`: trunk VLAN -> that VLAN's interface; native VLAN -> native interface; VLAN not in switch state -> gracefully dropped through the new nullopt path (previously the abort site).
- Stress: a broadcast storm on a loopback-cabled trunk pair pushed ~1.4M punts through the new resolution path (~800k/~575k attributed to the two tagged VLANs, ~11k untagged to native) with zero CHECK failures and zero crashes.
- Full `ConfigInterfaceSwitchportTrunkAllowedVlanTest` integration suite on the device: 8/8 pass (previously `RemoveVlansOneByOne` failed with the agent abort), with the peer link up so ambient punts arrived on the trunk port throughout.
- 10+ config commits/warmboots across add/remove sequences with trunk state present; agents stayed healthy.

## Review notes

Pre-publication review findings, for reviewer attention:

- Fixed before posting: `EthHdr` parsing of the punt buffer is wrapped in try/catch (falls back to the PVID) so a runt/truncated punt cannot throw out of the SAI RX callback.
- Known behavior, unchanged trust model: the 802.1Q VID from the punt is taken verbatim by the hw agent (same as the existing CPU-port punt path). The sw agent consumer (`getInterfaceIDForPkt`) validates it against the switch state and falls back / drops for unknown VLANs.

## Known issue observed during testing (why this PR is a draft)

This change makes the control plane correct and crash-free, but while validating it we found that the **forwarding plane does not classify tagged ingress into the tagged VLAN** on the switch we tested (Minipack3 / TH5). With a trunk port carrying native VLAN A + tagged VLANs B and C, ARP requests injected from the peer switch behave as follows:

| Frame | Punt attribution / interface resolution (this PR) | Hardware MAC learning |
|---|---|---|
| tagged VID B | VLAN B (correct) | learned in **VLAN A** (wrong) |
| tagged VID C | VLAN C (correct) | learned in **VLAN A** (wrong) |
| untagged | VLAN A (correct) | VLAN A (correct) |

The agent programs a SAI VLAN member per (VLAN, port) with `SAI_VLAN_TAGGING_MODE_TAGGED` (`SaiVlanManager::createVlanMember`), and the punt copy retains the original 802.1Q tag (which this PR's attribution relies on), yet the pipeline assigns every ingress frame to the port's PVID — so MACs learn in the wrong VLAN and tagged traffic is presumably bridged in the native VLAN as well. In other words, the crash fixed here was only the first blocker for trunk ports; per-VLAN L2 behavior needs additional (likely port ingress-classification) programming that FBOSS does not do today.

We're keeping this PR as a draft until there's agreement on whether trunk-port support is a direction worth pursuing in FBOSS. Feedback welcome — including pointers if the hardware behavior above is expected with the current SAI programming and there's a known missing attribute.

### DIFF
--- a/fboss/agent/DHCPv4Handler.cpp
+++ b/fboss/agent/DHCPv4Handler.cpp
@@ -22,6 +22,7 @@
 #include "fboss/agent/SwitchIdScopeResolver.h"
 #include "fboss/agent/SwitchStats.h"
 #include "fboss/agent/TxPacket.h"
+#include "fboss/agent/Utils.h"
 #include "fboss/agent/packet/DHCPv4Packet.h"
 #include "fboss/agent/packet/EthHdr.h"
 #include "fboss/agent/packet/Ethertype.h"
@@ -304,8 +305,7 @@ void DHCPv4Handler::processRequest(
 
   auto switchIp = state->getDhcpV4RelaySrc();
   if (switchIp.isZero()) {
-    auto interfaceIDOpt = sw->getState()->getInterfaceIDForPortIf(
-        PortDescriptor(pkt->getSrcPort()));
+    auto interfaceIDOpt = getInterfaceIDForPkt(*pkt, sw->getState());
     if (!interfaceIDOpt) {
       sw->stats()->dhcpV4DropPkt();
       XLOG(ERR) << "No interface for port " << pkt->getSrcPort()
--- a/fboss/agent/DHCPv6Handler.cpp
+++ b/fboss/agent/DHCPv6Handler.cpp
@@ -246,8 +246,7 @@ void DHCPv6Handler::processDHCPv6Packet(
 
   auto switchIp = state->getDhcpV6RelaySrc();
   if (switchIp.isZero()) {
-    auto intfIDOpt = sw->getState()->getInterfaceIDForPortIf(
-        PortDescriptor(pkt->getSrcPort()));
+    auto intfIDOpt = getInterfaceIDForPkt(*pkt, sw->getState());
     if (!intfIDOpt) {
       sw->stats()->dhcpV6DropPkt();
       XLOG(ERR) << "No interface for port " << pkt->getSrcPort()
--- a/fboss/agent/IPv4Handler.cpp
+++ b/fboss/agent/IPv4Handler.cpp
@@ -311,7 +311,7 @@ void IPv4Handler::handlePacket(
       return;
     }
     // Forward multicast packet directly to corresponding host interface
-    auto intfIDOpt = state->getInterfaceIDForPortIf(PortDescriptor(port));
+    auto intfIDOpt = getInterfaceIDForPkt(*pkt, state);
     if (intfIDOpt) {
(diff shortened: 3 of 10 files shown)


benoit-nexthop:
After discussing with the FBOSS team at Meta, it was decided to not fix this. Trunk ports are not supported on FBOSS. Closing.

PR #1580, opened 2026-09-01
[Nexthop][m4062nhp][minipack3] Change l2LearningMode to SOFTWARE in agent.conf default config

vvasavada-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

As per https://github.com/facebook/fboss/blob/b6c29b90b7ab2b2e6690935bb40861eac6e42a38/fboss/agent/hw/sai/switch/SaiSwitch.cpp#L1002-L1006, we should only be using SW learning mode in production. Change l2LearningMode from default HARDWARE to default SOFTWARE in agent.conf.

# Test Plan

None.

### DIFF
--- a/fboss/oss/scripts/run_configs/default_configs/minipack3/agent.conf
+++ b/fboss/oss/scripts/run_configs/default_configs/minipack3/agent.conf
@@ -8775,7 +8775,7 @@
 
     },
     "switchSettings": {
-      "l2LearningMode": 0,
+      "l2LearningMode": 1,
       "qcmEnable": false,
       "ptpTcEnable": false,
       "l2AgeTimerSeconds": 300,


srikrishnagopu:
can u rebase and resubmit ?

joseph5wu:
@vvasavada-nexthop Is this hardware l2 learning mode blocking any test?
In Meta, we don't always use SOFTWARE l2 learning mode unless it's RSW or multi_host_nic RACK switch.
While our majority FBOSS Switches are using HARDWARE mode by default.

Besides, if you wanna switch the l2 learning mode, you can also use fboss2-dev command:
https://github.com/facebook/fboss/blob/main/fboss/cli/fboss2/commands/config/l2/learning_mode/CmdConfigL2LearningMode.cpp
```
fboss2-dev config l2 learning-mode [hardware/software/disabled]
```

vvasavada-nexthop:
@joseph5wu it is not blocking any test. However, we noticed that `fboss2 show mac details` won't show entries if we use `HARDWARE` mode because it reads FBOSS's software FDB map rather than hardware MAC table. Given that production uses `SOFTWARE` anyway, we thought it bit to keep it as default.

If Meta is mostly running `HARDWARE` mode, we can close this PR. However, the mode will need to be changed to `SOFTWARE` for `fboss2 show mac details` to show anything in that case.

joseph5wu:
> vvasavada-nexthop

Even with hardware L2 learning mode, we can still show mac details. I've verified it on our production Montblanc switch which is also using Tomahawk5.

But I just confirmed with FBOSS Agent team(cc: @tanmayaudupa ), we will try to roll out software l2 learning mode as the new default for FBOSS. 
In that case, I think you can keep this PR

vvasavada-nexthop:
> Even with hardware L2 learning mode, we can still show mac details. I've verified it on our production Montblanc switch which is also using Tomahawk5.

@joseph5wu 
Could you confirm if these entries are ARP/NDP resolved or h/w L2 learned? 

`StaticL2ForNeighborUpdater` creates a static MAC entry per resolved neighbor regardless of learning mode. Our case is an L2 bridging topology with a traffic generator and no resolved neighbors, so the only MACs are hardware-learned dynamic ones — and those never reach SwitchState in HARDWARE mode.

```
void SaiSwitch::fdbEventCallbackLockedBottomHalf(
    std::vector<FdbEventNotificationData> fdbNotifications) {
  if (getSwitchRunState() == SwitchRunState::EXITING) {
    return;
  }
  std::vector<std::pair<L2Entry, L2EntryUpdateType>> l2Entries;
  {
    auto lock = std::lock_guard<std::mutex>(saiSwitchMutex_);
    if (managerTable_->bridgeManager().getL2LearningMode() !=
        cfg::L2LearningMode::SOFTWARE) {
      // Some platforms call fdb callback even when mode is set to HW. In
      // keeping with our native SDK approach, don't send these events up.
      XLOG(DBG4) << "Ignoring FDB learn notification as mode is not software";
      return;
    }
```

joseph5wu:
> > Even with hardware L2 learning mode, we can still show mac details. I've verified it on our production Montblanc switch which is also using Tomahawk5.
> 
> @joseph5wu Could you confirm if these entries are ARP/NDP resolved or h/w L2 learned?
> 
> `StaticL2ForNeighborUpdater` creates a static MAC entry per resolved neighbor regardless of learning mode. Our case is an L2 bridging topology with a traffic generator and no resolved neighbors, so the only MACs are hardware-learned dynamic ones — and those never reach SwitchState in HARDWARE mode.
> 
> ```
> void SaiSwitch::fdbEventCallbackLockedBottomHalf(
>     std::vector<FdbEventNotificationData> fdbNotifications) {
>   if (getSwitchRunState() == SwitchRunState::EXITING) {
>     return;
>   }
>   std::vector<std::pair<L2Entry, L2EntryUpdateType>> l2Entries;
>   {
>     auto lock = std::lock_guard<std::mutex>(saiSwitchMutex_);
>     if (managerTable_->bridgeManager().getL2LearningMode() !=
>         cfg::L2LearningMode::SOFTWARE) {
>       // Some platforms call fdb callback even when mode is set to HW. In
>       // keeping with our native SDK approach, don't send these events up.
>       XLOG(DBG4) << "Ignoring FDB learn notification as mode is not software";
>       return;
>     }
> ```

Yeah, they're from NDP.

joseph5wu:
@vvasavada-nexthop can you rebase this PR to the latest master? We're having trouble to import such PR and it's potentially due to your PR is a bit outdated

vvasavada-nexthop:
@joseph5wu Done!

joseph5wu:
> @joseph5wu Done!

Now this PR is changing minipack3? I thought you want to enable software l2 learning mode on NH platforms

vvasavada-nexthop:
@joseph5wu my intention was to make it default for all platforms agent.conf. Before rebase, it was changing both minipack3 and NH. But it seems NH is already changed by https://github.com/facebook/fboss/pull/1561.

If we don't want to change for minipack3, we don't need this PR anymore. NH is already changed.

vvasavada-nexthop:
>we will try to roll out software l2 learning mode as the new default for FBOSS.

However, I believe this PR is still valid if we are going to do above (and applies to all paltforms).

vvasavada-nexthop:
>However, I believe this PR is still valid if we are going to do above (and applies to all paltforms).

Any update on whether we want to still merge minipack change or discard this? @joseph5wu

joseph5wu:
@vvasavada-nexthop let's close this PR without changing Minipack

PR #1621, opened 2026-09-10
[Nexthop][fboss2-dev] Document the difference between `set` and `config` operations

benoit-nexthop:
# Summary

This adds some additional text to `fboss_cli.md` detailing the differences between the `fboss2`/`fboss2-dev` `set` and `config` commands, and what one should expect from each:

- `set` is a runtime-only write via a thrift client to the running agent; nothing is persisted.
- `config` edits the configuration file through a staged session and is persisted on commit.
- A `config` commit or reload (or a warm/cold boot) re-applies config and reverts any `set`.
- `set` changes are invisible to `config session diff` / `config history`.

# Test Plan

Docs-only change; no test additions or changes. Confirmed with visual inspection of the generated output using docusaurus.

Co-Authored-By: Steve Ulrich <sulrich@nexthop.ai>


benoit-nexthop:
cc @sulrich-nexthop

PR #1631, opened 2026-09-13
[Cisco] Prevent FSDB build OOM by sharding explicit instantiations

selekkala:
## Summary

- Split the `FsdbOperStateRoot` `NaivePeriodicSubscribableStorage` explicit instantiation into access, lifecycle, mutation, read, and subscription translation units.
- Keep a declaration-only whole-class instantiation to retain the existing vtable and RTTI ownership.
- Preserve normal template behavior by moving method definitions to an inline implementation header.
- Update both CMake and Buck source lists.

## Root cause

Growth in the switch-state schema increased the recursive Thrift COW storage and visitor template graph. The schema changes are valid, but they exposed a scalability problem in the existing monolithic explicit instantiation.

`FsdbNaivePeriodicSubscribableStateStorage.cpp` explicitly instantiated the complete class in one translation unit. Consequently, one compiler process had to materialize the graph for every non-template member, independent of overall build parallelism. The measured maximum RSS for that compiler process was 15,511,040 KiB, so reducing the global job count alone cannot reliably fix the single-process memory requirement.

## Fix

The method definitions remain available to generic users through `NaivePeriodicSubscribableStorage-inl.h`. The concrete `FsdbOperStateRoot` instantiation is divided by method group so no compiler process owns the complete implementation graph. The original translation unit still performs a declaration-only whole-class instantiation, preserving class metadata symbols while the new translation units provide the concrete method definitions.

No storage behavior or public API is changed.

## Validation

- Rebased cleanly onto upstream `main` at `866e812417`.
- Changed C++ files pass `clang-format --dry-run --Werror`.
- `git diff --check` passes.
- All shard translation units compiled in the forwarding-stack CMake build.
- The `fsdb` executable and storage archive linked successfully.
- The state-storage exported symbol set matched the monolithic build.
- Largest measured shard maximum RSS was below 9.4 GiB, versus approximately 14.8 GiB for the monolithic translation unit.
- A complete forwarding-stack build finished without the LLVM out-of-memory failure.

## Related follow-up

Draft PR [#1632](https://github.com/facebook/fboss/pull/1632) adds cgroup-aware build parallelism. It intentionally remains a draft and should land after this PR because its 10 GiB forwarding-job estimate is based on the post-sharding memory measurements.


srikrishnagopu:
Few enhancements went in this area to optimize the build time. I am closing this for now but please re-open if needed.

PR #1636, opened 2026-09-14
[Nexthop] [distro] Patch kiwi oem-dump blocksize check to use logical block size

saty-nexthop:
**Pre-submission checklist**
- [x] `pre-commit run`

# Summary

PXE installs fail on every platform with a 512e boot drive (512 byte logical / 4096 byte physical NVMe). Console:

```
Target blocksize(4096b) doesn't match image blocksize(512b)
[   16.718787] dracut-pre-mount[1111]: Rebooting.
```

The installer aborts before fetching the rootfs, so the box PXE-loops.

**Cause:** kiwi [#3022](https://github.com/OSInside/kiwi/pull/3022) (in kiwi 10.3.11) added `check_image_matches_target_blocksize` to the oem-dump dracut hook. It compares the image's `kiwi_target_blocksize` (`512`, kiwi's default since our `config.xml` sets none) against the target's **physical** block size. kiwi's own documentation defines `target_blocksize` as the **logical** size, so on 512e disks the check is a false positive. It arrived through an unpinned package update, not a source change here.

I reported this upstream as [OSInside/kiwi#3050](https://github.com/OSInside/kiwi/issues/3050) with a one-line fix in [OSInside/kiwi#3051](https://github.com/OSInside/kiwi/pull/3051).

**Fix here, until that ships:** `config.sh` rewrites that one assignment to read `logical_block_size`.

- Guard and `sed` match the same full assignment string, so they cannot diverge. Other `block_size` uses in the file (`dd bs=`, sector-table logic) are untouched.
- No-op once upstream fixes it or drops the check.
- Does not weaken the upstream intent: on a true 4Kn disk logical is also 4096, so a 512 image is still refused. Only the 512e false positive goes away.

# Test Plan

- `bash -n`, `shellcheck` and `shfmt` pass.
- Block run against real kiwi 10.3.11: stock file patches exactly one line; an already-fixed file is left alone.
- Image built with this change was PXE-installed on an NH-BW800-F (`physical_block_size=4096`, `logical_block_size=512`), the platform that hits the bug. Install reached a login prompt in about 4.5 minutes where it previously timed out at 10 minutes in a reboot loop. No `Target blocksize` line in the install log.
- Confirmed the patch is baked into the shipped artifact: the initrd inside the built image has `logical_block_size` in the `target_blocksize=` assignment, and `check_image_matches_target_blocksize` is still present and still called.


AarjunC:
Do we still need this change? https://github.com/OSInside/kiwi/issues/3050 landed yesterday.

